PR #3, opened 2016-03-09
Various fixes

williamgibb:
- Add a .gitignore file so that files created by packaging with pyinstaller are not at risk of being included in the git repo.
  - Redefine `exit()` by using `sys.exit()`
  - Replace print statements with `print()` function calls.


williballenthin:
LGTM, thanks!

PR #25, opened 2016-04-08
more refactoring

### LINKED ISSUE #23: make output function names consistent
`print` vs `output` etc.

williballenthin:
two things here:

  1. renamed `output_strings` to `print_string` to make consistent. closes #23.
  2. removed the `StringDecoder` class. made all its methods plain old functions.

i did this because the class was not used to maintain any state, but rather group related functions together. a better way of this grouping is namespaces, which in python, is done via modules. it think it makes sense to move many of the functions in `main.py` to `string_decoder.py` or something like that, and keep the script file small.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -402,12 +402,12 @@ def print_decoding_results(decoded_strings, min_length, group_functions, quiet=F
             if len_ds > 0:
                 if not quiet:
                     print("Decoding function at 0x%X (decoded %d strings)" % (fva, len_ds))
-                output_strings(ds_filtered, min_length, quiet=quiet)
+                print_strings(ds_filtered, min_length, quiet=quiet)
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -112,40 +112,26 @@ def get_candidate_functions(self):
...
-        return decoded_strings
+def identify_decoding_functions(vw, identification_plugins, functions):
+    identification_manager = IdentificationManager(vw, identification_plugins)
+    identification_manager.run_plugins(functions)
+    identification_manager.apply_plugin_weights()
+    return identification_manager
+
+
+def extract_decoding_contexts(vw, function):
+    return get_function_contexts(vw, function)
```
williballenthin: should just rename and call this function directly

PR #36, opened 2016-04-08
add stackstrings support

williballenthin:
currently implemented as a extra pass that bruteforces all function and searches the contents of the active stack frame at every call instruction

### DIFF
--- a/floss/stackstrings.py
+++ b/floss/stackstrings.py
@@ -1,0 +1,141 @@
+from collections import namedtuple
+
+import viv_utils
+import envi.archs.i386
+import envi.archs.amd64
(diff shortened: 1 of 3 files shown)


floss/stackstrings.py
```diff
@@ -0,0 +1,141 @@
...
+            # the offset into the stack frame at which the stackstring existed.
+            "offset",
+
+            # type: int
+            # the offset from the function frame at which the stackstring existed.
+            "frame_offset",
+        ])
+
+
+def getPointerSize(vw):
```
mr-tz: is it in viv_utils?

PR #69, opened 2016-04-13
flesh out test infrastructure

williballenthin:
purpose: to decouple tests and make them self contained. they're now simply directories with C source code an a YAML metadata file. the FLOSS test harness automatically picks them up and runs them.

how?
for the test cases in `tests/src`, move `README.md` to `test.yml`. The YAML file reads almost exactly the same to a human, but is easily parsable by a machine.

add a py.test `conftest.py` that recognizes `test.yml` files as metadata about a test case. it uses the metadata to extract the output binaries generated from a single test directores (that is, from `decode-base64.c` to `decode-base64` for ELF, `decode-base64.exe` for PE32, *and* `decode-base6464.exe` for PE64) (these files are declared in the YAML file). py.test will then  run FLOSS against each file as a separate test case. *well, it should, but the invocation of FLOSS is currently stubbed out, pending @mr-tz work.*

invoke the tests like this:

    $  py.test <path to test.yml or test directory>
    $  py.test tests/src
    $  py.test tests/src/decode-base64/test.yml

output:
```
$  py.test tests/src/decode-base64/test.yml -v
==================== test session starts ====================
platform linux2 -- Python 2.7.6, pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /home/user/env/bin/python2
cachedir: .cache
rootdir: /home/user/src/flare-floss, inifile: 
collected 3 items 

tests/src/decode-base64/test.yml::test-decode-base64::Windows::64bit PASSED
tests/src/decode-base64/test.yml::test-decode-base64::Linux::32bit PASSED
```


williballenthin:
outstanding work:

  - [ ] provide implementation of `conftest.py:extract_strings()`. this should come from @mr-tz work on extracting strings from test sample appended data.
  - [ ] add build step to append the expected decoded string to each generated test binary. can pull the expected strings from the `test.yml` file. add this step to the Makefiles and `test.yml` build instructions.

mr-tz:
Need to add new dependencies to setup.py?
```
q
pyyaml
```

williballenthin:
good catch.

definitely need `pyyaml` for parsing the .yml files.

`q` is a neat module for debugging. quick tutorial: you can import like `import q`, and then use it as a function or decoration. use it like a print statement, except it returns the response, so you can embed it in other function calls. like: `f.write(q(file_header))`. output goes to `/tmp/q`, so you can use `tail -f /tmp/q` to watch the log entries (they're nicely formatted and colorized). if you use it as a decorator, it traces the arguments, return value, and execution time of the function, eg:

```
@q
def do_something(a, b,c):
    ...
```

its called `q` cause its easy to type.

williballenthin:
so... there shouldn't be any imports of `q`, since those are used for debugging.

PR #70, opened 2016-04-13
main: use tabulate package to format tables.

### LINKED ISSUE #65: output table alignment is bad on 64-bit samples
```
λ floss.exe 5a92...8ab1b

[...snip...]

1 strings decoded:
Offset       Called At    String
----------   ----------   -------------------------------------
0xFFFFFFFFBFB3F6A0   0x140001DCA   Error in GetTokenInformation\n
Finished execution after 15.212000 seconds
```

recommend using the package `tabulate` to format tables and automatically apply padding.

williballenthin:
output looks basically the same, but the column sizes dynamically change based on what data is put in them.

closes #65.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -9,6 +9,7 @@ import pkg_resources
 from time import time
 from optparse import OptionParser
 
+import tabulate
 import plugnplay
 import viv_utils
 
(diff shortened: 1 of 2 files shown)


floss/main.py
```diff
@@ -313,17 +315,15 @@ def print_all_strings(path, n=4, quiet=False):
             print("%s" % (s.s))
     else:
         print("Static ASCII strings")
-        print("Offset       String")
-        print("----------   -------------------------------------")
-        for s in strings.extract_ascii_strings(b, n=n):
-            print("0x%08X   %s" % (s.offset, s.s))
+        print(tabulate.tabulate(
+            [("0x%X" % (s.offset,), s.s) for s in strings.extract_ascii_strings(b, n=n)],
```
williballenthin: this is ugly, sorry. but we need a list of 2-tuples containing formatted strings. its going to be messy no matter how you do it...

floss/main.py
```diff
@@ -333,15 +333,14 @@ def print_stack_strings(extracted_strings, n=4, quiet=False):
...
-        print("----------  ------------  -------")
+        extracted_strings = list(filter(lambda s: len(s.s) >= n, extracted_strings))
+        count = len(extracted_strings)
 
-        for ss in extracted_strings:
-            if len(ss.s) >= n:
-                print("0x%08x  0x%04x    %s" % (ss.fva, ss.frame_offset, ss.s))
+        print("FLOSS extracted %d stack strings" % (count))
+        print(tabulate.tabulate(
+            [("0x%X" % (s.fva,), "0x%X" % (s.frame_offset,), s.s) for s in extracted_strings],
```
williballenthin: this is even worse...

PR #81, opened 2016-04-13
add test/decode-string-with-header.

### LINKED ISSUE #43: test: encoded blob has header with key
```
typedef struct ENCODED_STRING {
    unsigned char rc4_key[0x10];
    uint32_t length;
    unsigned char encoded_data[length;
} ENCODED_STRING;
```

williballenthin:
closes #43.


mr-tz:
I'll update the Makefile (see #83)

PR #82, opened 2016-04-13
add test/decode-reencode-string

### LINKED ISSUE #51: test: string re-encoded after use
```
decode(s);
gethostbyname(s);
encode(s);
```

williballenthin:
closes #51.


mr-tz:
I'll update the Makefile (see #83)

PR #83, opened 2016-04-13
Build executables to bin directory

mr-tz:
Build the test executables to tests/src/<TESTNAME>/bin/
.gitignore /bin directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,9 @@ build/
 dist/
 lib/
 .cache
+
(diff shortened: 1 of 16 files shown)


williballenthin:
kinda sucks there's so much duplication across every test directory. makes updates like this kinda tedious. but, it allows each test to be completely independent (outside of requiring `make all` and `make clean`).

williballenthin:
looks good to me

Issue #76, opened 2016-04-13
mark tests as xfail is FLOSS doesn't currently support the behavior


williballenthin:
these declarations should be stored in `test.yml`, and raised by the `conftest.py` handler where appropriate.

use py.test `xfail`: https://pytest.org/latest/skipping.html#imperative-xfail-from-within-a-test-or-setup-function

mr-tz:
add Xfail: to test.yml

PR #86, opened 2016-04-14
Testing architecture gap filling

mr-tz:
Addresses the outstanding work from #69.
1. fix for decoded strings in decode-split-stackstrings test.yml
2. added script that appends the test data (decoded strings) from a test.yml file to a binary
3. run FLOSS and assert that deobfuscated strings have been found via conftest.py

Current limitations:
- Need to invoke tests from FLOSS main directory
- Only analyze Windows PE files (as FLOSS can't analyze ELF files currently, anyway)
- Might want to add decoding function offsets in test.yml file, although this might be infeasible due to chaning offsets, currently don't extract it from appended test data either
- Currently, gets expected strings from test.yml files instead of appended data

Example run output:
```
tests/src/decode-base64/test.yml::test-decode-base64::Windows::64bit FAILED
tests/src/decode-from-global/test.yml::test-decode-from-global::Windows::32bit PASSED
tests/src/decode-from-global/test.yml::test-decode-from-global::Windows::64bit PASSED
tests/src/decode-from-heap/test.yml::test-decode-from-heap::Windows::32bit FAILED
tests/src/decode-from-heap/test.yml::test-decode-from-heap::Windows::64bit FAILED
tests/src/decode-from-stack/test.yml::test-decode-from-stack::Windows::32bit PASSED
tests/src/decode-from-stack/test.yml::test-decode-from-stack::Windows::64bit PASSED
tests/src/decode-global-stackstrings/test.yml::test-decode-global-stackstrings::Windows::32bit FAILED
tests/src/decode-global-stackstrings/test.yml::test-decode-global-stackstrings::Windows::64bit FAILED
tests/src/decode-in-place/test.yml::test-decode-in-place::Windows::32bit PASSED
tests/src/decode-in-place/test.yml::test-decode-in-place::Windows::64bit PASSED
tests/src/decode-local-stackstrings/test.yml::test-decode-local-stackstrings::Windows::32bit PASSED
tests/src/decode-local-stackstrings/test.yml::test-decode-local-stackstrings::Windows::64bit PASSED
tests/src/decode-rc4/test.yml::test-decode-rc4::Windows::32bit FAILED
tests/src/decode-rc4/test.yml::test-decode-rc4::Windows::64bit FAILED
tests/src/decode-single-byte-xor/test.yml::test-decode-single-byte-xor::Windows::32bit PASSED
tests/src/decode-single-byte-xor/test.yml::test-decode-single-byte-xor::Windows::64bit PASSED
tests/src/decode-split-stackstrings/test.yml::test-decode-split-stackstrings::Windows::32bit PASSED
tests/src/decode-split-stackstrings/test.yml::test-decode-split-stackstrings::Windows::64bit PASSED
tests/src/decode-string-with-header/test.yml::test-decode-string-with-header::Windows::32bit FAILED
tests/src/decode-string-with-header/test.yml::test-decode-string-with-header::Windows::64bit FAILED
tests/src/decode-to-global/test.yml::test-decode-to-global::Windows::32bit PASSED
tests/src/decode-to-global/test.yml::test-decode-to-global::Windows::64bit PASSED
tests/src/decode-to-heap/test.yml::test-decode-to-heap::Windows::32bit PASSED
tests/src/decode-to-heap/test.yml::test-decode-to-heap::Windows::64bit PASSED
tests/src/decode-to-output-buf/test.yml::test-decode-to-output-buf::Windows::32bit PASSED
tests/src/decode-to-output-buf/test.yml::test-decode-to-output-buf::Windows::64bit PASSED
tests/src/decode-to-stack/test.yml::test-decode-to-stack::Windows::32bit PASSED
tests/src/decode-to-stack/test.yml::test-decode-to-stack::Windows::64bit PASSED
```

### DIFF
--- a/tests/append_test_data.py
+++ b/tests/append_test_data.py
@@ -1,0 +1,131 @@
+"""
+Use this script to append data from the test.yml file to a sample.
+
+Dictionary format:
+{
+ "fva_string (0x%x)":["string1","string2",...]
+ <...>
+ optional: "all":["stringN",...]
+}
+
+Example:
+{
+ "0x401000":["hostid","SYSTEM"]
+ "all":["8.8.8.8","explorer.exe"]
+}
+"""""
+
+import os
+import sys
+import yaml
+import json
+import struct
+from pprint import pprint
+
+
+FILE_START = 0
+FILE_END = 2
+
+MAGIC = "FLSS"
+SIZE_OFFSET = 4
+SIZE_LEN = 4
+SIZE_MAGIC = len(MAGIC)
+
+
+def contains_magic(sample_path):
+    try:
+        f = open(sample_path, "rb")
+        f.seek((-SIZE_MAGIC), FILE_END)
+        m = f.read(SIZE_MAGIC)
+    except Exception as e:
+        print str(e)
+    finally:
+        f.close()
+    return m == MAGIC
+
+
+def read_test_dict_from_file(sample_path):
+    if not contains_magic(sample_path):
+        return None
+
+    test_dict = None
+    try:
+        f = open(sample_path, "rb")
+        f.seek((-(SIZE_MAGIC + SIZE_OFFSET + SIZE_LEN)), FILE_END)
+        test_dict_meta = f.read(SIZE_OFFSET + SIZE_LEN)
+        (data_offset, data_len) = struct.unpack("<II", test_dict_meta)
+        f.seek(data_offset, FILE_START)
+        test_dict_meta = f.read(data_len)
+        test_dict = json.loads(test_dict_meta)
+    except Exception as e:
+        print str(e)
+    finally:
+        f.close()
+
+    return test_dict
+
+
+def get_test_dict_from_yaml(yaml_file):
+    test_dict = {}
+    try:
+        f = open(yaml_file, "r")
+        spec = yaml.safe_load(f)
+        test_dict["all"] = spec["Decoded strings"]
+        # TODO add decoding function offsets
+    except Exception as e:
+        print str(e)
+        return test_dict
+    finally:
+        f.close()
+    return test_dict
+
+
+def append_test_dict_to_file(sample_path, test_dict):
+    json_data = json.dumps(test_dict)
+    data_len = len(json_data)
+    file_size = os.path.getsize(sample_path)
+    test_data = struct.pack("<II", file_size, data_len)
+    return append_data(sample_path, json_data + test_data + MAGIC)
+
+
+def append_data(sample_path, data):
+    try:
+        f = open(sample_path, "ab")
+        f.write(data)
+    except Exception as e:
+        print str(e)
+        return False
+    finally:
+        f.close()
+    return True
+
+
+def main():
+    if len(sys.argv) != 2:
+        print("Usage: %s <SAMPLE_PATH>" % sys.argv[0])
+        return -1
+
+    sample_path = sys.argv[1]
+    if not os.path.isfile(sample_path):
+        print("%s is not a file" % sample_path)
+        return
+
+    if contains_magic(sample_path):
+        print("%s already contains test data:" % sample_path)
+        pprint(read_test_dict_from_file(sample_path))
+        return
+
+    testdir = os.path.dirname(os.path.dirname(sample_path))
+    yaml_file = os.path.join(testdir, "test.yml")
+
+    test_dict = get_test_dict_from_yaml(yaml_file)
+
+    if test_dict:
+        print("Created test dictionary from %s:" % yaml_file)
+        pprint(test_dict)
+
+    if append_test_dict_to_file(sample_path, test_dict):
+        print("Successfully appended test data to %s" % sample_path)
+
+if __name__ == "__main__":
+    main()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,48 @@ import os
 import yaml
 import pytest
 
+import viv_utils
+
+import floss.main as floss_main
+import floss.identification_manager as im
+import floss.stackstrings as stackstrings
+from append_test_data import read_test_dict_from_file
+
+
+def read_test_dict(sample_path):
+    """
+    Reads appended test dictionary from sample and returns "all" items for now.
+    Use tests/append_test_data.py to append test data from test.yml file to a sample.
+    """
+    test_dict = read_test_dict_from_file(sample_path)
+    if test_dict is None:
+        # sample does not contain test data
+        return []
+    else:
+        # TODO can extract decoding functions offsets from test_dict
+        return test_dict["all"]
+
+
+def extract_strings(sample_path):
+    """
+    Deobfuscate strings from sample_path
+    """
+    vw = viv_utils.getWorkspace(sample_path)
+    function_index = viv_utils.InstructionFunctionIndex(vw)
+    decoding_functions_candidates = identify_decoding_functions(vw)
+    decoded_strings = floss_main.decode_strings(vw, function_index, decoding_functions_candidates)
+    decoded_stackstrings = stackstrings.extract_stackstrings(vw)
+    decoded_strings.extend(decoded_stackstrings)
+    return [ds.s for ds in decoded_strings]
+
+
+def identify_decoding_functions(vw):
+    selected_functions = floss_main.select_functions(vw, None)
+    selected_plugin_names = floss_main.select_plugins(None)
+    selected_plugins = filter(lambda p: str(p) in selected_plugin_names, floss_main.get_all_plugins())
+    decoding_functions_candidates = im.identify_decoding_functions(vw, selected_plugins, selected_functions)
+    return decoding_functions_candidates
+
 
 def pytest_collect_file(parent, path):
     if path.basename == "test.yml":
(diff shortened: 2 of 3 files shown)


tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+import os
+import sys
+import yaml
+import json
+import struct
+from pprint import pprint
+
+
+FILE_START = 0
+FILE_END = 2
```
williballenthin: nice, thank you

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+
+MAGIC = "FLSS"
+SIZE_OFFSET = 4
+SIZE_LEN = 4
+SIZE_MAGIC = len(MAGIC)
```
williballenthin: good

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def contains_magic(sample_path):
+    try:
+        f = open(sample_path, "rb")
+        f.seek((-SIZE_MAGIC), FILE_END)
+        m = f.read(SIZE_MAGIC)
+    except Exception as e:
+        print str(e)
```
williballenthin: should use `logging` for this

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+    finally:
+        f.close()
```
williballenthin: alternatively, can use `open` as a context manager:

```
with open(sample_path, "rb") as f:
     ...
```

automatically handles cleaning up

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+SIZE_OFFSET = 4
+SIZE_LEN = 4
+SIZE_MAGIC = len(MAGIC)
+
+
+def contains_magic(sample_path):
+    try:
+        f = open(sample_path, "rb")
+        f.seek((-SIZE_MAGIC), FILE_END)
+        m = f.read(SIZE_MAGIC)
```
williballenthin: can simplify by removing the temp variable:

    return f.read(size_MAGIC) == MAGIC

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+FILE_START = 0
+FILE_END = 2
+
+MAGIC = "FLSS"
+SIZE_OFFSET = 4
+SIZE_LEN = 4
+SIZE_MAGIC = len(MAGIC)
+
+
+def contains_magic(sample_path):
```
williballenthin: minor style nit: if a function returns a boolean, recommend naming it starting with `is_`, `has_`. or `does_`.

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+        return None
+
+    test_dict = None
+    try:
+        f = open(sample_path, "rb")
+        f.seek((-(SIZE_MAGIC + SIZE_OFFSET + SIZE_LEN)), FILE_END)
+        test_dict_meta = f.read(SIZE_OFFSET + SIZE_LEN)
+        (data_offset, data_len) = struct.unpack("<II", test_dict_meta)
+        f.seek(data_offset, FILE_START)
+        test_dict_meta = f.read(data_len)
```
williballenthin: should check expected size vs actual size

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+        test_dict = json.loads(test_dict_meta)
```
williballenthin: return this directly. don't need the temp variable `test_dict`.

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+    finally:
+        f.close()
+
+    return test_dict
+
+
+def get_test_dict_from_yaml(yaml_file):
+    test_dict = {}
+    try:
+        f = open(yaml_file, "r")
```
williballenthin: use mode 'rb' for consistent handling on Windows systems. on Linux, doesn't matter.

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+        # TODO add decoding function offsets
+    except Exception as e:
+        print str(e)
+        return test_dict
+    finally:
+        f.close()
+    return test_dict
+
+
+def append_test_dict_to_file(sample_path, test_dict):
```
williballenthin: should ensure there's no existing data appended to the file. not required, but its good to be defensive.

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+    if len(sys.argv) != 2:
+        print("Usage: %s <SAMPLE_PATH>" % sys.argv[0])
+        return -1
+
+    sample_path = sys.argv[1]
+    if not os.path.isfile(sample_path):
+        print("%s is not a file" % sample_path)
+        return
+
+    if contains_magic(sample_path):
```
williballenthin: ah, i see you already do this check here. nvm.

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+        print("%s already contains test data:" % sample_path)
+        pprint(read_test_dict_from_file(sample_path))
+        return
+
+    testdir = os.path.dirname(os.path.dirname(sample_path))
```
williballenthin: this seems fragile. it assumes the `test.yml` file is two directories up, but this could be configured differently on a per-test basis. perhaps we should provide the `test.yml` file as a parameter? alternatively, the `test.yml` could be the primary argument, and the script would locate the binary files to update.

tests/conftest.py
```diff
@@ -3,6 +3,48 @@
...
+    """
+    Deobfuscate strings from sample_path
+    """
+    vw = viv_utils.getWorkspace(sample_path)
+    function_index = viv_utils.InstructionFunctionIndex(vw)
+    decoding_functions_candidates = identify_decoding_functions(vw)
+    decoded_strings = floss_main.decode_strings(vw, function_index, decoding_functions_candidates)
+    decoded_stackstrings = stackstrings.extract_stackstrings(vw)
+    decoded_strings.extend(decoded_stackstrings)
+    return [ds.s for ds in decoded_strings]
```
williballenthin: i love how simple this is

tests/conftest.py
```diff
@@ -15,14 +57,12 @@ def collect(self):
...
-                if os.path.exists(filepath):
-                    yield FLOSSTest(self, platform, arch, filename, spec)
-
-
-def extract_strings(sample_path):
-    # TODO: implement me
-    return []
+                # TODO only test Windows PE files for now
+                _, fileext = os.path.splitext(filename)
+                if fileext == ".exe":
```
williballenthin: check `platform == 'WINDOWS'` rather than file extension

williballenthin: also, i think these are items that should be marked `xfail` see #76, rather than ignored. this indicates we'd like to support the functionality, but that it not expected to work yet.

tests/conftest.py
```diff
@@ -15,14 +57,12 @@ def collect(self):
...
+                    filepath = os.path.join(test_dir, "bin", filename)
+                    if os.path.exists(filepath):
+                        yield FLOSSTest(self, platform, arch, os.path.join("bin", filename), spec)
```
williballenthin: i think `test.yml` should indicate the output directory and path. assuming there's a `bin` directory is probably usually correct, but should be configurable.

tests/conftest.py
```diff
@@ -42,11 +82,13 @@ def runtest(self):
...
 
+        # TODO: alternatively, get test data from appended data:
+        # expected_strings = set(read_test_dict(test_path))
         expected_strings = set(self.spec["Decoded strings"])
         found_strings = set(extract_strings(test_path))
 
-        # TODO: enable this
-        # assert expected_strings < found_strings
+        # assert that expected_strings is not empty
+        assert expected_strings and (expected_strings <= found_strings)
```
williballenthin: good catch

tests/src/decode-split-stackstrings/test.yml
```diff
@@ -5,8 +5,8 @@ Input buffer location: n/a
 Output buffer location: stack
 
 Decoded strings:
-    - hello world
-    - hello moon
+    - goodbye world
+    - goodbye moon
```
williballenthin: thanks

williballenthin:
overall, looks really good! i'm excited to merge this and see how the tests do.

i provided a code review as comments on this PR. please check those out. the main functional thing was assumptions about the `bin` directory. if we can avoid it, i think we should avoid this assumption (but definitely keep the `bin` directory around). we can do this by including the `bin` directory in the paths listed in `test.yml`.

williballenthin:
i'm going to merge this and address some of the things i commented on.

tests/append_test_data.py
```diff
@@ -0,0 +1,131 @@
...
+    except Exception as e:
+        print str(e)
+    finally:
+        f.close()
+    return m == MAGIC
+
+
+def read_test_dict_from_file(sample_path):
+    if not contains_magic(sample_path):
+        return None
```
williballenthin: don't return mixed types from a function. raise an exception instead.

tests/append_test_data.py
mr-tz: I like the second idea.

PR #87, opened 2016-04-14
address comments on #86

williballenthin:
major change is that `append_test_data.py` now accepts the `test.yml` file instead of the sample.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,8 @@ except ImportError:
 
 
 requirements = [
+    "q",
+    "pyyaml",
     "tabulate",
     "vivisect",
     "plugnplay",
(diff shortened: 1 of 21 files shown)


setup.py
```diff
@@ -9,6 +9,8 @@
 
 
 requirements = [
+    "q",
```
mr-tz: So, you want to include `q` here?

tests/conftest.py
```diff
@@ -82,13 +79,17 @@ def runtest(self):
...
+
         # TODO: alternatively, get test data from appended data:
         # expected_strings = set(read_test_dict(test_path))
         expected_strings = set(self.spec["Decoded strings"])
         found_strings = set(extract_strings(test_path))
 
-        # assert that expected_strings is not empty
-        assert expected_strings and (expected_strings <= found_strings)
+        if expected_strings:
+            assert expected_strings <= found_strings
```
mr-tz: Need `assert False` for else case?

mr-tz:
Very nice

setup.py
williballenthin: nope :-)

tests/conftest.py
williballenthin: if there are no expected strings, then the test should not fail

PR #88, opened 2016-04-14
move footer code to its own module

williballenthin:
note: review this after merging #87

moves get/set methods for sample test data ("footers") to their own module in `tests/`. makes it easy for different test scripts to access footers (like `conftest.py`).

unrelated: update `conftest.py` to pretty-format test failures. for example:

```
$ py.test tests/src/decode-base64/test.yml -v      
========== test session starts ==========
platform linux2 -- Python 2.7.6, pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /home/user/env/bin/python2
cachedir: .cache
rootdir: /home/user/src/flare-floss, inifile: 
collected 3 items 

tests/src/decode-base64/test.yml::test-decode-base64::Windows::64bit FAILED
tests/src/decode-base64/test.yml::test-decode-base64::Linux::32bit xfail

========== FAILURES ==========
__________ usecase: test-decode-base64::Windows::64bit __________
FLOSS extraction failed:
   expected: set([u'hello world'])
   got: set([])
========== 1 failed, 1 passed, 1 xfailed in 7.39 seconds ==========
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,12 +57,9 @@ class YamlFile(pytest.File):
         test_dir = os.path.dirname(str(self.fspath))
         for platform, archs in spec["Output Files"].items():
             for arch, filename in archs.items():
-                # TODO only test Windows PE files for now
-                _, fileext = os.path.splitext(filename)
-                if fileext == ".exe":
-                    filepath = os.path.join(test_dir, "bin", filename)
-                    if os.path.exists(filepath):
-                        yield FLOSSTest(self, platform, arch, os.path.join("bin", filename), spec)
+                filepath = os.path.join(test_dir, filename)
+                if os.path.exists(filepath):
+                    yield FLOSSTest(self, platform, arch, filename, spec)
 
 
 class FLOSSTest(pytest.Item):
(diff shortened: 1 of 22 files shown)


mr-tz:
Neat changes! Thanks.

tests/conftest.py
```diff
@@ -57,12 +42,15 @@ def collect(self):
...
         for platform, archs in spec["Output Files"].items():
             for arch, filename in archs.items():
-                # TODO only test Windows PE files for now
-                _, fileext = os.path.splitext(filename)
-                if fileext == ".exe":
-                    filepath = os.path.join(test_dir, "bin", filename)
-                    if os.path.exists(filepath):
-                        yield FLOSSTest(self, platform, arch, os.path.join("bin", filename), spec)
+                filepath = os.path.join(test_dir, filename)
+                if os.path.exists(filepath) and footer.has_footer(filepath):
```
mr-tz: Do we want to fallback to read strings from test.yml file if the sample has no appended test data?

tests/footer.py
```diff
@@ -0,0 +1,63 @@
+import json
```
mr-tz: misses `import os`

tests/conftest.py
williballenthin: this would be easy to do and deprioritize #89 for a while

PR #91, opened 2016-04-14
Feature/wb tests3

williballenthin:
two things:

  - `conftest.py` pulls decoded strings from `test.yml` if there's no footer on the sample binary.
  - add test for "wrapped decoder functions"


tests/conftest.py
```diff
@@ -43,7 +43,8 @@ def collect(self):
         for platform, archs in spec["Output Files"].items():
             for arch, filename in archs.items():
                 filepath = os.path.join(test_dir, filename)
-                if os.path.exists(filepath) and footer.has_footer(filepath):
+                if os.path.exists(filepath):
+                    #and footer.has_footer(filepath):
```
mr-tz: delete

PR #143, opened 2016-05-06
Refactor hooks, add a few new ones

williballenthin:
Refactor hooks into their own file, so they can be used during brute force emulation and decoder routine emulation.

Add hook for `memcpy` and `calloc`.


mr-tz:
looks great

PR #145, opened 2016-05-06
hooks: use callconv over raw stack manip

### LINKED ISSUE #124: Make API hooks independent from calling convention
Currently `RtlAllocateHeapHook`, `AllocateHeap`, and `MallocHeap` only focus on the `__stdcall` calling convention.

williballenthin:
rely on existing vivisect analysis to extract args rather than touching the stack directly. will support 64bit samples better.

closes #124

### DIFF
--- a/floss/api_hooks.py
+++ b/floss/api_hooks.py
@@ -196,7 +196,7 @@ class RtlAllocateHeapHook(viv_utils.emulator_drivers.Hook):
         # works for kernel32.HeapAlloc
         if callname == "ntdll.RtlAllocateHeap":
             emu = driver
-            size = driver.getStackValue(0xC)
+            hheap, flags, size = argv
             va = self._allocate_mem(emu, size)
             callconv.execCallReturn(emu, va, len(argv))
             return True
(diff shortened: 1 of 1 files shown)


floss/api_hooks.py
```diff
@@ -234,7 +233,7 @@ def hook(self, callname, driver, callconv, api, argv):
         if callname == "msvcrt.malloc" or \
            callname == "msvcrt.calloc":
             emu = driver
-            size = driver.getStackValue(0x4)
+            size = argv[0]
             va = self._allocate_mem(emu, 0x100)  # TODO hard-coded!
```
mr-tz: The allocation is hard-coded here anyway. What do we want to do?

williballenthin: oh, that doesn't look correct. let me fix this.

floss/api_hooks.py
```diff
@@ -234,7 +234,7 @@ def hook(self, callname, driver, callconv, api, argv):
            callname == "msvcrt.calloc":
             emu = driver
             size = argv[0]
-            va = self._allocate_mem(emu, 0x100)  # TODO hard-coded!
+            va = self._allocate_mem(emu, size)
```
mr-tz: Do we want to add a sanity check and limit `size` to a max value?

williballenthin: this is already done in the `_allocate_mem` method

PR #147, opened 2016-05-08
better handle large files

williballenthin:
- suggest providing flag -a for non-PE files
  - suggest providing flag -a for large files
  - extract static strings from large files
  - warn users of truncated static strings for 4GB files on 32bit systems

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -476,12 +476,11 @@ def main():
     if options.all_strings:
         floss_logger.info("Extracting static strings...")
         print_static_strings(sample_file_path, min_length=min_length, quiet=options.quiet)
-    elif magic not in SUPPORTED_FILE_MAGIC:
-        floss_logger.warning("Unsupported input file type, automatically extracting only static strings...")
-        print_static_strings(sample_file_path, min_length=min_length, quiet=options.quiet)
 
     if magic not in SUPPORTED_FILE_MAGIC:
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -390,42 +391,77 @@ def create_script(sample_file_path, ida_python_file, decoded_strings):
             raise e
     # TODO return, catch exception in main()
 
+KILOBYTE = 1024
+MEGABYTE = 1024 * KILOBYTE
+MAX_FILE_SIZE = 16 * MEGABYTE
```
williballenthin: these should go at the top of the file

floss/main.py
```diff
@@ -451,6 +487,9 @@ def print_stack_strings(extracted_strings, min_length, quiet=False):
         print("")
 
 
+SUPPORTED_FILE_MAGIC = set(["MZ"])
```
williballenthin: should go at the top of the file

floss/main.py
williballenthin: i picked `MAX_FILE_SIZE` out of thin air. we can tweak as appropriate. but i think generating a viv workspace for a 16MB file will take a little bit.

mr-tz:
From the most recent Travis tests:
```30 passed, 21 xfailed in 482.10 seconds```
vs.
```32 passed, 22 xfailed in 497.10 seconds```

PR #148, opened 2016-05-08
Better handle rep instruction

mr-tz:
Do not shortcut `reps` (vivisect emulator option `reponce`), use viv-utils' `maxrep` to limit number of instructions executed.


floss/stackstrings.py
```diff
@@ -50,7 +50,7 @@ def extract_call_contexts(vw, fva):
     monitor = CallContextMonitor(vw, emu.getStackCounter())
     driver = viv_utils.emulator_drivers.FunctionRunnerEmulatorDriver(emu)
     driver.add_monitor(monitor)
-    driver.runFunction(fva, maxhit=1, func_only=True)
+    driver.runFunction(fva, maxhit=1, maxrep=0x100, func_only=True)
```
mr-tz: Any opinions whether we are able to turn `maxrep` down for stackstrings?

williballenthin: we could, by editing here: https://github.com/fireeye/flare-floss/blob/master/floss/stackstrings.py#L53

i'm fine with doing it, or even limiting `maxrep=1` unless we see a sample otherwise.

floss/utils.py
```diff
@@ -10,11 +10,13 @@ def makeEmulator(vw):
     removeStackMemory(emu)
     emu.initStackMemory(stacksize=int(0.5 * ONE_MB))
     emu.setStackCounter(emu.getStackCounter() - int(0.25 * ONE_MB))
+    emu.setEmuOpt('i386:reponce', False)  # do not short circuit rep prefix
     return emu
 
 
 def removeStackMemory(emu):
     # TODO this is a hack while vivisect's initStackMemory() has a bug (see issue #27)
+    # TODO does this bug still exist?
```
williballenthin: vivisect#27 doesn't seem related. i'm not sure what the bug is.

tests/src/decode-to-stack-rep-mov/test-decode-to-stack-rep-mov.c
```diff
@@ -0,0 +1,58 @@
...
+
+    // memcpy(in, raw, 12);
+    #if UINTPTR_MAX == 0xffffffff
+    __asm__ (
+        "leal %0, %%edi\n" // ; lea edi, [in]
+        "movl %1, %%esi\n" // ; mov esi, [raw]
+        "movl $3, %%ecx\n" // ; mov ecx, 3
+        "rep movsd\n"
+        : "=m" ( in )
+        : "r" ( raw )
```
williballenthin: h4x0r

williballenthin:
looks good, thanks

mr-tz:
From the most recent Travis tests:
```30 passed, 21 xfailed in 482.10 seconds```
vs.
```32 passed, 22 xfailed in 497.10 seconds```
(added this to the wrong PR initially)

PR #153, opened 2016-05-10
reformat code to adhere to pep8

williballenthin:
pep8 

excludes E501 (line width limited to 80 characters), since this blindly adds a bunch of line breaks that increases indentation and weird structure. some lines could be shorter, but i dont think we should require exactly 80 chars or less.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 
 install:
     - pip install https://github.com/williballenthin/vivisect/zipball/master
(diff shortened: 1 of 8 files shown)


williballenthin:
these changes were generated automatically (and reviewed manually) with `autopep8`: 

    ~/env/bin/autopep8 --in-place --ignore E501 ./path/to/file.py

mr-tz:
Looks good. I agree on excluding E501.

Issue #167, opened 2016-05-18
outputted strings should be unique


williballenthin:
```
>> how come the output lines aren't unique? like certain binaries it will show the same decoder/offset/string a dozen times

[3:30:23 PM] Willi Ballenthin: the default view displays the address of the data that was decoded, and the decoding routine
[3:30:34 PM] Willi Ballenthin: so these will be the same
[3:30:49 PM] Willi Ballenthin: but the context in which the data was decoded is different
[3:31:19 PM] Willi Ballenthin: i'll open an issue to make the output unique, since I think this is definitely more intuitive
```

by default, we should output only unique decoded strings, and display all strings & contexts if requested.

williballenthin:
closed in #168

PR #168, opened 2016-05-18
add parameter for printing only unique results

pilate:


pilate:
Satisfies #167

williballenthin:
i wonder if perhaps this should be the default mode, and a "more advanced mode" (hardcore mode?) that prints duplicates with their different contexts? cc: @mr-tz thoughts?

mr-tz:
Yes, I think enabling it by default makes sense for most users.

mr-tz:
Awesome, thanks. We can expand the advanced behavior based on this now.

PR #172, opened 2016-05-19
Memdiff rewrite

pilate:
Currently, ~25% of FLOSS processing time is spent in envi.memory.memdiff.

Doing a little research into the arguments/output, I noticed a few patterns:

- The average input size is ~150kb
- The average amount of diffs per call is ~1
- The average diff starts ~50% of the way through the input
- The average diff is ~10 bytes long

This made a few optimizations pretty obvious, which are implemented with a new memdiff function:

- Try to locate the first diff without manually comparing each individual byte (binary search)
- After each diff, check if the remaining bytes are equivalent 

I'm running some tests to verify no regression in functionality, but so far everything looks good. 

The only concern I have with this implementation is that memdiff_search **may** hit the recursion limit on extremely long inputs with a diff located in specific locations. However, after a few hundred runs on random samples, I haven't seen it come anywhere close to that.

cProfile comparison of before/after runs show an almost 25% improvement in run time: https://gist.github.com/pilate/b825f84b0494afce0f134a909bfbabd2

### DIFF
--- a/floss/string_decoder.py
+++ b/floss/string_decoder.py
@@ -17,10 +17,8 @@ def memdiff_search(a, b):
 
     :param a: The original sequence of bytes
     :param b: The sequence of bytes to compare with
-    :param offset: Offset to be added to the return value
     :type a: str
     :type b: str
-    :type offset: int
     :rtype: int offset of the first location a and b differ, None if strings match
     '''
 
--- a/floss/string_decoder.py
+++ b/floss/string_decoder.py
@@ -48,7 +48,7 @@ def memdiff(bytes1, bytes2):
     :type bytes1: str
     :type bytes2: str
     :rtype: list of (offset, length) tuples indicating locations bytes1 and
-      bytes 2 differ
+      bytes2 differ
     '''
     # Shortcut matching inputs
     if bytes1 == bytes2:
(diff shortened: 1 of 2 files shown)


williballenthin:
i'm very excited to include this patch. improving FLOSS's performance is great news, since it can sometimes run a bit slowly.

A few things:

  - please add function level documentation describing parameters and return value, and generally what the function does. for example, what does `memdiff_search` return? from a quick glance, im not quite sure.

  - i think we'd better add some test cases to exercise this diffing function different ways. these should go in the `tests/` directory in a new `.py` file. id like to see that it works with 0 diffs, 1 diff, 10 diffs, and diffs happening at different levels of the binary search.

pilate:
Added docstrings and a variety of tests. Let me know if that covers your concerns.

williballenthin:
looks good to me. @mr-tz its up to you to merge.

will need to update travis/appveyor configs to also run these tests.

williballenthin:
pep8 nits:

```
./tests/test_memdiff.py:10:1: E303 too many blank lines (3)
./tests/test_memdiff.py:27:1: E302 expected 2 blank lines, found 1
```

mr-tz:
Looks good!

PR #174, opened 2016-05-20
skip extraction on bufs of one repeated character

pilate:
I was running into some samples* that were spending over 10 minutes in extract_unicode_strings. Debugging the calls, there were thousands of occurrences of inputs consisting of all "A" and NULL characters, occasionally over 2gb in size.

This patch shortcuts the extract functions if a buf consists of nothing but a single repeated 'banned' character.

I went with .count instead of multiplication because of the sizes of some inputs.

Before/after profile: https://gist.github.com/pilate/45907ae43a85292fcd7393aed2c25d2e

* aa8a1c3afdf1d9b16fae80e6444d5889


williballenthin:
LGTM

i love that these patches are being driven by real-world use of FLOSS :-)

williballenthin:
@mr-tz up to you to merge

mr-tz:
Very nice!

PR #176, opened 2016-05-20
Save workspace with option and for expert mode

mr-tz:


williballenthin:
build failure is just a flaky network connection by appveyor to github. not a problem.

PR #180, opened 2016-05-23
show cleaner error message on getWorkspace failure

pilate:
Basic handling of getWorkspace errors. Using this locally for cleaner output, not sure if you guys want it in or not.

Patched outputs:

$ python floss/main.py f12e2a67-20b4-5469-ac65-379b89196b3a
Vivisect failed to load the input file: bad char in struct format

$ python floss/main.py 82afe1bd-878e-5b27-b4f9-636393a3fbcb
Vivisect failed to load the input file: Machine 0401 is not supported for PE!

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -563,7 +563,11 @@ def main(argv=None):
     if options.expert:
         options.save_workspace = True
 
-    vw = viv_utils.getWorkspace(sample_file_path, should_save=options.save_workspace)
+    try:
+        vw = viv_utils.getWorkspace(sample_file_path, should_save=options.save_workspace)
+    except Exception, e:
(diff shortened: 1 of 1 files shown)


mr-tz:
Looks good to me, up to @williballenthin to merge.

williballenthin:
The formatting looks great! I like how clean the code is that results in
much better output.

Some thoughts:
  - if in verbose mode, we should print the stack track to enable
easier debugging. If we use logging (see below) then it's as easy as
providing 'exc_info=True' or 'False' to the output method call.
  - i wonder when we should use stderr/logging versus print. Tentatively
leaning towards using logging since it works better with quiet mode ( don't
want people thinking "vivisect error" is a valid string from the binary).
But I really haven't thought this through and would appreciate feedback.

PR #181, opened 2016-05-23
Improvements to IDA script output, escaping

mr-tz:
- Includes stackstrings: adds repeatable comment to stack offset, if it can be located
- Improved character escaping for immediate and script output
- General improvements to script output

Unrelated:
- Apply filter regex to stackstrings

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -67,6 +67,7 @@ def sanitize_string_for_printing(s):
     :return: sanitized string
     """
     sanitized_string = s.encode('unicode_escape')
+    sanitized_string = sanitized_string.replace('\\\\', '\\')  # print single backslashes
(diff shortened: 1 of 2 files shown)


floss/main.py
```diff
@@ -66,9 +66,8 @@ def sanitize_string_for_printing(s):
     :param s: input string
     :return: sanitized string
     """
-    sanitized_string = s.replace('\n', '\\n')
-    sanitized_string = sanitized_string.replace('\r', '\\r')
-    sanitized_string = sanitized_string.replace('\t', '\\t')
+    sanitized_string = s.encode('unicode_escape')
```
williballenthin: very cool

williballenthin:
nice work!

PR #182, opened 2016-05-24
Basic and advanced mode

mr-tz:
- Basic output is new default (basic header output, but no offsets, duplicate strings etc.)
- For expert mode show headers, offsets, duplicates (different contexts), etc.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -303,42 +303,46 @@ def print_identification_results(sample_file_path, decoder_results):
         print(tabulate.tabulate(
             [(hex(fva), "%.5f" % (score,)) for fva, score in candidates],
             headers=["address", "score"]))
-    print("")
 
 
-def print_decoding_results(decoded_strings, min_length, group_functions, quiet=False):
+def print_decoding_results(decoded_strings, min_length, options):
     """
     Print results of string decoding phase.
     :param decoded_strings: list of decoded strings ([DecodedString])
     :param min_length: minimum string length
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -303,42 +303,46 @@ def print_identification_results(sample_file_path, decoder_results):
         print(tabulate.tabulate(
             [(hex(fva), "%.5f" % (score,)) for fva, score in candidates],
             headers=["address", "score"]))
-    print("")
 
 
-def print_decoding_results(decoded_strings, min_length, group_functions, quiet=False):
+def print_decoding_results(decoded_strings, min_length, options):
```
williballenthin: i recommend keeping all arguments explicit, rather than grouping them into this `options` variable

  - if we use kwargs, python can warn us if extra args are provided that aren't expected. if we have the the options catchall, python can't help us. 
  - kwargs are a nice form of documentation that don't require a user to read the function comments.

if the number of args passed to functions begins to feel uncomfortable, then this is a code smell that indicates a refactor is needed. some solutions might be to split off modes of execution into different classes, or maybe have some for of formatter object.

mr-tz:
I think it's a good idea to divide the different execution modes to keep the code easier to maintain. It's already pretty tedious to keep track of the existing options. And they are likely to get expanded soon.

williballenthin:
acknowledged. i'll take a look into this. in the mean time, changes look good!

Issue #185, opened 2016-05-24
use exit status to indicate failure


williballenthin:
for example, for max file size, should indicate error, versus simply returning from `main()`.

williballenthin:
closed in #187

PR #187, opened 2016-05-24
non zero return code for errors in main

pilate:
return status 1 on non-pe files, or files over max_file_size ( #185 )


williballenthin:
ref #185

PR #199, opened 2016-05-30
Add initial r2script support

Maijin:
Still need https://github.com/radare/radare2/issues/4373 (https://github.com/radare/radare2/issues/1885) to be fixed to be able to add decoded stack strings 

floss-r2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ __pycache__/
 # Distribution / packaging
 build/
 dist/
+floss.egg-info/
 lib/
 .cache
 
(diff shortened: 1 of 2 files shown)


mr-tz:
Nice, looks good to me. Up to @williballenthin to merge. We actually discussed using Base64 encoded strings initially. We might pick it up again?! Thanks!

williballenthin:
yeah, i think we should consider using base64 for the IDA script, too

williballenthin:
the various output mechanisms need to be refactored so `main` isn't so gigantic. but this isnt @Maijin's problem. looks good to me!

PR #200, opened 2016-06-05
substitution cipher test case

mr-tz:
- decoding routine currently not identified
- emulation relies on memchr hook (addressed in #201)

### DIFF
--- a/tests/src/Makefile
+++ b/tests/src/Makefile
@@ -17,7 +17,8 @@ TESTS= \
 	decode-string-by-index \
 	decode-reencode-string \
 	decode-wrapped-decoder \
-	decode-to-stack-rep-mov
+	decode-to-stack-rep-mov \
+    decode-substitution-cipher
 
 
 clean:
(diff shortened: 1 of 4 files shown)


tests/src/Makefile
```diff
@@ -17,7 +17,8 @@ TESTS= \
 	decode-string-by-index \
 	decode-reencode-string \
 	decode-wrapped-decoder \
-	decode-to-stack-rep-mov
+	decode-to-stack-rep-mov \
+    decode-substitution-cipher
```
williballenthin: indentation is off here

tests/src/decode-substitution-cipher/test-decode-substitution-cipher.c
```diff
@@ -0,0 +1,58 @@
...
+
+		*pEncodedChar = c;
+	}
+}
+
+int main(int argc, char **argv)
+{
+    char in_out[] = "xzhhKDmKjhT";
+
+	substitution_cipher(in_out, strlen(in_out));
```
williballenthin: indentation is off here

williballenthin:
thanks for developing a test case to demonstrate the issue. awesome!

PR #201, opened 2016-06-05
memchr API hook

mr-tz:
Hook for memchr

### DIFF
--- a/floss/api_hooks.py
+++ b/floss/api_hooks.py
@@ -270,7 +270,6 @@ class StrlenHook(viv_utils.emulator_drivers.Hook):
             return True
         raise viv_utils.emulator_drivers.UnsupportedFunction()
 
-
     def readStringAtRva(self, emu, rva, maxsize=None):
         """
         Borrowed from vivisect/PE/__init__.py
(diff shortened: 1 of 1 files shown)


floss/api_hooks.py
```diff
@@ -284,13 +283,36 @@ def readStringAtRva(self, emu, rva, maxsize=None):
...
+    Hook and handle calls to memchr
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(MemchrHook, self).__init__(*args, **kwargs)
+
+    def hook(self, callname, driver, callconv, api, argv):
+        if callname == "msvcrt.memchr":
+            emu = driver
+            ptr, value, num = argv
```
williballenthin: might want to check and limit `num` so we don't try to read 16GB accidentally.

floss/api_hooks.py
```diff
@@ -284,13 +283,36 @@ def readStringAtRva(self, emu, rva, maxsize=None):
             if maxsize and maxsize <= len(ret):
                 break
             x = emu.readMemory(rva, 1)
-            if x == '\x00' or x == None:
+            if x == '\x00' or x is None:
```
williballenthin: good fix

williballenthin:
LGTM, thanks!

PR #204, opened 2016-06-06
Fix extract_XXX_strings error when buf is mmap object

shizhx:
`mmap object` don't have `count` method

### DIFF
--- a/floss/strings.py
+++ b/floss/strings.py
@@ -10,6 +10,12 @@ REPEATS = ["A", "\x00", "\xfe", "\xff"]
 
 String = namedtuple("String", ["s", "offset"])
 
+def buf_filled_with(buf, c):
+    for i in range(len(buf)):
+        if buf[i] != c:
+            return False
+
+    return True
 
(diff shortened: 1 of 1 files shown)


williballenthin:
any sense for the performance implications of this fix? my intuition is that its probably slower than `string.count` (which is probably implemented in C), but I'm not too worried unless its prohibitively slow.

shizhx:
there is a bug, try this:
1. use `echo -n "AAAAAAAA" > test` to generate a file 
2. run `floss -a test` to extract static strings
then an exception is threw:
```
Traceback (most recent call last):
  File "floss/main.py", line 644, in <module>
  File "floss/main.py", line 578, in main
  File "floss/main.py", line 506, in print_static_strings
  File "floss/strings.py", line 28, in extract_ascii_strings
AttributeError: 'mmap.mmap' object has no attribute 'count'
Failed to execute script main
```
I am working on an android boot.img, then trigger this bug

williballenthin:
i agree this was a bug

Issue #222, opened 2016-06-12
floss Crashes and gives MemoryError


oddcoder:
This is file I used
[re001.zip](https://github.com/fireeye/flare-floss/files/310868/re001.zip)

```
➜  ~ floss --version 
floss 1.2.0
https://github.com/fireeye/flare-floss/
➜  ~ floss diskm2.dll           

FLOSS decoded 202 strings
VA<AAA
<AAA
AAAAAAAAAAAAAAAAr
BCC`.//2
RTT0
wzz<
p044^
`ff`
Cff@Oww
8%``*
kl(xDEn
@<*,jJ{}
OR,C
Ve0,
^b"Ut:
Lugr,
$?&D
(FQP
*'?|b
2,XxKp
3T ^,
C 6rX
|OW&)
qShJ
By2l
0I$2
P%V"@
Xy~]
^+za
)@cYQ
."/0\
$zXj370
9O Vvz
~%wo^
>~N[D|~
~w_N
}jfpF
$dHiPT
tXpt
M@[5m
 b85F.Py6zj
lpe<
=AUx
2"y^
=%^p
,#FL
Z/bS
[UlJ
(30b
P7Vg
(XWR8
0HSC
h_Z7
_8bD
:S-V
VJN^+
`Oq@+
p@\a
VHPn
 `fX
5a&@
S2vrA`*
:! 5
Q)zLy|
RsW4
8`zT
0'Ld
, ~q
_6u(
xxit
y c h<
^EFt_
k vE
?x6(
6>[OX
uXnoB
qAU0
4|Vv
8%MW`1:U
w\F[Td8
 BDrC^
Gp.0?2
(#MS(
N v=hH
@/_>`
`%n&
;NIW
fu1K
812B
`8n%
r-I(
qH_7BxH1
'0c@
&Jmt2
y8j_
_<OL
VlRjT#
X':$
_HBz
2rP-
DPR8
bpkW
wkJL
$"De
TI V
fspc
`=2D
ip-`
p`bSs`K_/
@<r:
a6Onu"f
@RLj@
l~(p
urq)V
6Z|B
LT#+h
L669
o<Q[
76`X
XPJCxEW
8%yb
P #W
h,R^
D2EI
H9lc
4A9^
`,)<
vQf$
 X)N
LTz6
<k9vX+
'`xB
#I@X
Xtap
fX#?
KO@Sf
(WNE@
!8exh
]X4dS(
L t{
0Y"\P
iF`t
x` PV
[...]

Issue #205, opened 2016-06-06
do performance regression testing for #204


williballenthin:
#204 may be much slower on some samples because `count` is implemented in pure Python. need to demonstrate there is not a serious performance regression. appropriate sample might be the one in #174 aa8a1c3afdf1d9b16fae80e6444d5889.

shizhx:
1. should replace `range()` function with `xrange()` function
2. I did performance regression testing just now, it is **terrible**!

So, how to resolve this problem? `count` method is faster than pure python implement code, but `main.py` pass a `mmap object` as `buf` parameter, `mmap object` has no method `count`.

williballenthin:
using `xrange` is a fine idea

williballenthin:
what if we did something like:

```
def buf_filled_with(buf, c):
    if not hasattr(buf, 'count'):
        buf = bytes(buf)
    return buf.count(buf[0]) == len(buf)
```

shizhx:
that sounds great

shizhx:
Unfortunately, `bytes(mmap object)` doesn't seems to work

PR #224, opened 2016-06-16
Fix performance regression of buf_filled_with function

shizhx:
because of previous modification #204  of `strings.py`, performance of `buf_filled_with` is bad.
but anyway, `mmap object` has no `count` method, I have to implement it with pure python.
now, `mmap object` + `buf_filled_with` is even **faster** than `read` + `count`.

here is test script and full performance testing report: https://github.com/shizhx/others/tree/master/flare-floss_test

part of performance testing report:
use `read` with `count`
-------------------------------
```
2048MB sample:
         1918 function calls (1898 primitive calls) in 11.757 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.051    0.051   11.757   11.757 strings_count.py:1(<module>)
        1    0.000    0.000   11.702   11.702 strings_count.py:89(main)
        1    7.188    7.188    7.188    7.188 {method 'read' of 'file' objects}
        2    4.514    2.257    4.514    2.257 {method 'count' of 'str' objects}
        1    0.000    0.000    2.297    2.297 strings_count.py:33(extract_ascii_strings)
        1    0.000    0.000    2.217    2.217 strings_count.py:60(extract_unicode_strings)
```

use `mmap` with `buf_filled_with`
---------------------------------------------
```
2048MB sample:
         3147648 function calls (3147628 primitive calls) in 8.093 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.043    0.043    8.093    8.093 strings_buf_filled_with.py:1(<module>)
        1    0.000    0.000    8.047    8.047 strings_buf_filled_with.py:90(main)
        2    2.636    1.318    8.046    4.023 strings_buf_filled_with.py:15(buf_filled_with)
  1048576    5.271    0.000    5.271    0.000 {method 'count' of 'str' objects}
        1    0.001    0.001    5.191    5.191 strings_buf_filled_with.py:34(extract_ascii_strings)
        1    0.000    0.000    2.856    2.856 strings_buf_filled_with.py:61(extract_unicode_strings)
2097660/2097656    0.140    0.000    0.140    0.000 {len}
```

**please perform code review carefully and help me to improve these**

relevant issues: #205 #222

### DIFF
--- a/floss/strings.py
+++ b/floss/strings.py
@@ -7,15 +7,28 @@ ASCII_RE_4 = re.compile("([%s]{%d,})" % (ASCII_BYTE, 4))
 UNICODE_RE_4 = re.compile(b"((?:[%s]\x00){%d,})" % (ASCII_BYTE, 4))
 REPEATS = ["A", "\x00", "\xfe", "\xff"]
 
+SLICE_SIZE = 4096
 
 String = namedtuple("String", ["s", "offset"])
 
 def buf_filled_with(buf, c):
-    for i in range(len(buf)):
-        if buf[i] != c:
+    if hasattr(buf, 'count'):
+        return buf.count(c) == len(buf)
+
+    begin = 0
+    end = 0
+    while begin < len(buf):
+        end = begin + SLICE_SIZE
+        if end > len(buf):
+            end = len(buf)
+
+        if not buf[begin:end].count(c) == (end - begin):
             return False
 
+        begin = end
+
     return True
+
 
 def extract_ascii_strings(buf, n=4):
     '''


pilate:
We can pick up a little more performance by caching the last chunk and doing a straight comparison. 

This looks like it's actually several times faster than count for both strings and mmap objects:

```python
def buf_filled_with(buf, c):
    prev_chunk = ""
    for offset in xrange(0, len(buf), SLICE_SIZE):
        new_chunk = buf[offset: offset + SLICE_SIZE]
        new_len = len(new_chunk)

        if prev_chunk and (prev_chunk[:new_len] == new_chunk):
            continue

        if not new_chunk.count(c) == new_len:
            return False

        prev_chunk = new_chunk

    return True
```

Need to validate I'm not overlooking something simple, though.

floss/strings.py
```diff
@@ -7,16 +7,29 @@
...
 
 def buf_filled_with(buf, c):
-    for i in range(len(buf)):
-        if buf[i] != c:
+    if hasattr(buf, 'count'):
+        return buf.count(c) == len(buf)
+
+    begin = 0
+    end = 0
+    while begin < len(buf):
```
williballenthin: `for end in xrange(0, len(buf), SLICE_SIZE):`

williballenthin:
@shizhx thank you very much for taking ownership of the testing and enhancement of this routine. i think your patch is moving in the right direction. the attr-checking for `count` is just what i had in mind when i made a similar suggestion in the issue. processing the data in chunks is a neat idea.

i'm interested in the results that @pilate reports. i like that his implementation is short, and following the same lines of thinking in using chunked data.

aside from further testing, my only thought is: can we avoid the creation of temporary str/arrays? seems each time we take a chunk, it has to be copied. avoiding the copy *could* improve performance, but it might also be too small to matter.

my first thought was using a `memoryview`, but apparently these aren't supported for `mmap` objects (http://bugs.python.org/issue9229).

second thought is using `buf.index(dupe_chunk, start, end)` and catching the `ValueError` if there's no match (immediately return false, no loopy exceptions).

williballenthin:
travis build failure is not related to this patch.

shizhx:
@pilate  great code! It is really faster than  "count implementation". I think it is the best implementation at the moment.

@williballenthin completely reasonable thinking.

I am not very similar with Python, I will create an issue for this kind of problem next time.

pilate:
@williballenthin I tried to avoid temporary strings using mmap.find, but it was far slower than any of the existing solutions.

williballenthin:
closed in #226

PR #227, opened 2016-06-20
New default output mode

mr-tz:
- show all string types by default
- new options to individually disable string type: --no-XXX-strings, .e.g --no-static-strings
- grouped options in usage info
- configured expert mode profile


mr-tz:
I've also removed the offset output for static strings as it adds little value for an analyst

williballenthin:
looks awesome, nice work!

PR #235, opened 2016-07-26
change to hex value that falls out of ASCII range

mr-tz:
This should help to get rid of a bunch of false positives due to heap allocations.


williballenthin:
good catch!

Issue #203, opened 2016-06-05
failing test: test-decode-substitution-cipher::Windows::64bit


mr-tz:
Decoding routine identification fails.
With #201 emulation works.

williballenthin:
closed in 4a5017aad472875d4a2d821031cff7e61066ebbc

williballenthin:
this is a detection issue

mr-tz:
closed via #238

PR #238, opened 2016-07-30
plugin to identify suspicious mov instructions

### LINKED ISSUE #202: failing test: test-decode-substitution-cipher::Windows::32bit
Decoding routine identification fails.
With #201 emulation works.

mr-tz:
Identifies functions with register dereferenced writes to memory, e.g. `mov [eax], cl`

Testing this has shown good results. I like that it's so simple.

I think this approach has more potential (see all the TODO comments in the plugin).

Among other things this closes #202 and #203.

### DIFF
--- a/floss/plugins/mov_plugin.py
+++ b/floss/plugins/mov_plugin.py
@@ -1,0 +1,46 @@
+import envi
+import viv_utils
+
+import plugin_object
+import floss.interfaces as interfaces
+
+
+class MovPlugin(plugin_object.GeneralPlugin):
+    """
+    Identify suspicious MOV instructions.
+    """
+    implements = [interfaces.DecodingRoutineIdentifier]
+    version = 1.0
+
+    def identify(self, vivisect_workspace, function_vas):
+        candidate_functions = {}
+        # walk over every instruction
+        for fva in function_vas:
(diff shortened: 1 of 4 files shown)


mr-tz:
mov instructions like this are also common in C standard functions, e.g. memcpy. Signatures for these would be good to have...

floss/plugins/mov_plugin.py
```diff
@@ -0,0 +1,46 @@
...
+    def identify(self, vivisect_workspace, function_vas):
+        candidate_functions = {}
+        # walk over every instruction
+        for fva in function_vas:
+            f = viv_utils.Function(vivisect_workspace, fva)
+            for n_bb in xrange(0, len(f.basic_blocks)):
+                bb = f.basic_blocks[n_bb]
+                try:
+                    instructions = bb.instructions
+                    for n_instr in xrange(0, len(bb.instructions)):
```
williballenthin: why not use `for insn in bb.instructions: ...`?

williballenthin:
looks like the tests fail on some platforms. perhaps we need to tweak the weighting?

mr-tz:
It works when the plugin actually is enabled :)

mr-tz:
closed via #238

PR #240, opened 2016-08-05
appveyor: attempt to build and dist 32 and 64bit archs

williballenthin:
*should* be able to use env variable `PYTHON_ARCH` to fetch current architecture. ref: https://github.com/ogrisel/python-appveyor-demo/blob/master/appveyor.yml#L18


williballenthin:
if appveyor CI passes, and files appear in S3, need to update links in readme

PR #252, opened 2016-10-19
added x64dbg output functions and command line option

sam-b:
Let me know if you want any changes/don't actually want this as a feature :) 

Doing FLOSS -> IDA -> x64dbg only transfers decoded strings so I stuck with that as annotating stack strings is somewhat clunky without the concept of local vars. Any feedback is much appreciated.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -6,6 +6,7 @@ import sys
 import mmap
 import string
 import logging
+import json
 from time import time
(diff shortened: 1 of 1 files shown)


williballenthin:
this looks awesome, thank you!

clshack:
:+1:

PR #267, opened 2017-03-23
filter out long strings

mr-tz:
I propose this sanity check to filter out strings longer than `MAX_STRING_LENGTH`. I think 95% of decoded strings should be less than 100 characters anyway. So, `MAX_STRING_LENGTH = 2048` is a very generous margin. We could even lower the boundary, if you agree.
This is a quick fix for cases when emulation goes out of control.


williballenthin:
i like it, looks good!

mr-tz:
Would you see a benefit in keeping the substring up to the max length instead? I'm undecided between the two options.

PR #270, opened 2017-04-18
Filter FP strings

mr-tz:
Creating this PR to open the discussion.
The goal of these filters is to reduce the FP rate. However, it is possible that we filter out true positive strings. Should we add a command-line argument to let users disable filtering (enable it by default)? We then probably should move the filtering to the print_* functions.

### DIFF
--- a/floss/utils.py
+++ b/floss/utils.py
@@ -7,6 +7,8 @@ from collections import OrderedDict
 ONE_MB = 1024 * 1024
 STACK_MEM_NAME = "[stack]"
 
+MAX_STRING_LENGTH = 2048
+
 
 def makeEmulator(vw):
     """
--- a/floss/utils.py
+++ b/floss/utils.py
@@ -79,8 +79,8 @@ def hex(i):
     return "0x%X" % (i)
 
 
-FP_FILTER_PREFIXES = re.compile(r"^.?((p|P|0)?VA)|0\\A|P?\\A|\[A")  # remove string prefixes: pVA, VA, 0VA, etc.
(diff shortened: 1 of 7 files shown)


williballenthin:
changes look good. i like how were are separating out the specific details into readable functions and namespacing them.

williballenthin:
> Should we add a command-line argument to let users disable filtering (enable it by default)

exactly

> We then probably should move the filtering to the print_* functions.

ok, seems like a fair place to put this

floss/utils.py
```diff
@@ -76,3 +77,26 @@ def get_vivisect_meta_info(vw, selected_functions):
...
+    return FP_FILTER_CHARS.match(s)
+
+
+def strip_string(s):
+    """
+    Return string stripped from false positive (FP) pre- or suffixes.
+    :param s: input string
+    :return: string stripped from FP pre- or suffixes
+    """
+    return re.sub(FP_FILTER_PREFIXES, "", re.sub(FP_FILTER_SUFFIXES, "", s))
```
mr-tz: See 8df5b2f

mr-tz:
If we move filtering to the print functions other tools that use FLOSS won't benefit from this change. I lean towards keeping the filtering in the `extract_*` methods. Alternatively, we could add a separate `filter_strings` function that takes all decoded/stack strings and returns the filtered and stripped strings.

mr-tz:
I'm not sure if it's my machine, but FLOSS runs very slow (>7 minutes) for me at the moment...

mr-tz:
@williballenthin should we bump up the version (the API changed a little bit) and merge?

williballenthin:
yup, lets do it

PR #292, opened 2018-07-01
Introduce more granularity to setting log levels

capnspacehook:
As discussed in https://github.com/fireeye/flare-floss/issues/291, the current method of setting Floss's logging levels when used as a library is inadequate. Either the user has to manually set the 20 or so loggers Floss uses to the level they desire, or use the floss.main.set_logging_level() function, which has the unpleasant side effect of removing all of the root logger's handlers. 

This creates a new function, floss.main.set_log_config(), which has the exact same interface as floss.main.set_logging_level(), but sets ALL of Floss's loggers to the specified level. set_logging_level() is changed to call set_log_config() internally to remove code duplication.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -207,7 +207,15 @@ def make_parser():
 
     return parser
 
-def set_log_config(should_debug=False, should_verbose=False):
+def set_logging_levels(should_debug=False, should_verbose=False):
+    """
+    Sets the logging levels of each of Floss's loggers individually. 
+    Recomended to use if Floss is being used as a library, and your 
+    project has its own logging set up. If both parameters 'should_debug'
+    and 'should_verbose' are false, the logging level will be set to ERROR.
+    :param should_debug: set logging level to DEBUG
+    :param should_verbose: set logging level to INFO
+    """
     log_level = None
     emulator_driver_level = None
 
@@ -300,7 +308,15 @@ def set_log_config(should_debug=False, should_verbose=False):
     logging.getLogger("floss.plugins.mov_plugin.MovPlugin").setLevel(log_level)
         
 
-def set_logging_level(should_debug=False, should_verbose=False):
+def set_log_config(should_debug=False, should_verbose=False):
+    """
+    Removes root logging handlers, and sets Floss's logging level.
+    Recomended to use if Floss is being used in a standalone script, or 
+    your project doesn't have any loggers. If both parameters 'should_debug'
+    and 'should_verbose' are false, the logging level will be set to ERROR.
+    :param should_debug: set logging level to DEBUG
+    :param should_verbose: set logging level to INFO
+    """
     # reset .basicConfig root handler
     # via: http://stackoverflow.com/a/2588054
     root = logging.getLogger()
@@ -314,7 +330,8 @@ def set_logging_level(should_debug=False, should_verbose=False):
         logging.basicConfig(level=logging.INFO)
     else:
         logging.basicConfig(level=logging.WARNING)
-        set_log_config(should_debug, should_verbose)
+
+    set_logging_levels(should_debug, should_verbose)
 
 
 def parse_functions_option(functions_option):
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -222,31 +314,7 @@ def set_logging_level(should_debug=False, should_verbose=False):
...
-        # ignore messages like:
-        # WARNING:vtrace.platforms.win32:LoadLibrary C:\Users\USERNA~1\AppData\Local\Temp\_MEI21~1\vtrace\platforms\windll\amd64\symsrv.dll: [Error 126] The specified module could not be found
-        # WARNING:vtrace.platforms.win32:LoadLibrary C:\Users\USERNA~1\AppData\Local\Temp\_MEI21~1\vtrace\platforms\windll\amd64\dbghelp.dll: [Error 126] The specified module could not be found
-        logging.getLogger('vtrace.platforms.win32').setLevel(logging.ERROR)
-
-        # ignore messages like:
-        # WARNING:plugins.arithmetic_plugin.XORPlugin:identify: Invalid instruction encountered in basic block, skipping: 0x4a0637
-        logging.getLogger("floss.plugins.arithmetic_plugin.XORPlugin").setLevel(logging.ERROR)
-        logging.getLogger("floss.plugins.arithmetic_plugin.ShiftPlugin").setLevel(logging.ERROR)
+        set_log_config(should_debug, should_verbose)
```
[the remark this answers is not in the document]
capnspacehook: The only reason I didn't call set_log_config() from every case is because I figured logging.BasicConfig() would do the same thing, and it'd be redundant. set_log_config() is called only if logging.BasicConfig() sets the loggers to the WARNING level, because set_log_config() sets most of the loggers  to ERROR level, and what's not explicitly set is already set by logging.BasicConfig(). Calling set_log_config() when logging.BasicConfig() has already set the loggers to DEBUG or INFO level would just accomplish the same thing twice it'd think.

floss/main.py
```diff
@@ -207,6 +207,98 @@ def make_parser():
 
     return parser
 
+def set_log_config(should_debug=False, should_verbose=False):
```
williballenthin: This function sets logger levels, but doesn't do any other configuration. The other function (currently: `set_logging_level`) both sets logger levels and does configuration. This naming is confusing to me, and I think it should be swapped:

  - `set_log_levels`: this function
  - `set_log_config`: the other function, calls `logging.basicConfig` and also `set_log_levels`

floss/main.py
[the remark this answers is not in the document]
williballenthin: I think that this routine should be called from all cases (or de-dented) for symmetry. Maybe there is some duplication, but I believe this is ok, since otherwise, its confusing why the final case is special.

FYI, this type of reasoning is the perfect type of explanation to go into code comments.

floss/main.py
```diff
@@ -207,6 +207,98 @@ def make_parser():
...
+    # ignore messages like:
+    # DEBUG: get_meta_data: Function at 0x00401000 has meta data: Size: 177
+    logging.getLogger("floss.plugins.function_meta_data_plugin.FunctionSizePlugin").setLevel(log_level)
+
+    # ignore messages like:
+    # DEBUG: identify: suspicious MOV instruction at 0x00401017 in function 0x00401000: mov byte [edx],al
+    logging.getLogger("floss.plugins.mov_plugin.MovPlugin").setLevel(log_level)
+        
 
 def set_logging_level(should_debug=False, should_verbose=False):
```
williballenthin: let's change the name of this routine, as described above.

floss/main.py
capnspacehook: Done.

floss/main.py
capnspacehook: Yep, done.

williballenthin:
LGTM!

thanks @capnspacehook!

capnspacehook:
Lol, you merged right as I was typing this :joy:

As I was testing generating script files for idapro, radare2 and x64dbg, and I noticed that there was a few print statements that were messing with my logging. Specifically, in the floss.main.create_(x64dbg/idapro/radare2)_database() functions. Could there possibly be a parameter that controls whether the print statement is logged instead?

PR #301, opened 2019-01-08
Added an option to set the maximum number of times emulation can revisits addresses.

BenjaminSoelberg:
Added an option to set the maximum number of times emulation can revisits revisits addresses.

Increasing it improves string decoding within loops and
complex flows, but takes longer.

Also improved the string de-duplication process (still not perfect).

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -55,7 +55,7 @@ def hex(i):
     return "0x%X" % (i)
 
 
-def decode_strings(vw, decoding_functions_candidates, min_length, no_filter=False, max_instruction_count=20000):
+def decode_strings(vw, decoding_functions_candidates, min_length, no_filter=False, max_instruction_count=20000, max_hits=1):
     """
     FLOSS string decoding algorithm
     :param vw: vivisect workspace
@@ -63,13 +63,14 @@ def decode_strings(vw, decoding_functions_candidates, min_length, no_filter=Fals
     :param min_length: minimum string length
     :param no_filter: do not filter decoded strings
     :param max_instruction_count: The maximum number of instructions to emulate per function.
+    :param max_hits: The maximum number of hits per address
     :return: list of decoded strings ([DecodedString])
     """
     decoded_strings = []
     function_index = viv_utils.InstructionFunctionIndex(vw)
     # TODO pass function list instead of identification manager
     for fva, _ in decoding_functions_candidates.get_top_candidate_functions(10):
-        for ctx in string_decoder.extract_decoding_contexts(vw, fva):
+        for ctx in string_decoder.extract_decoding_contexts(vw, fva, max_hits):
             for delta in string_decoder.emulate_decoding_routine(vw, function_index, fva, ctx, max_instruction_count):
                 for delta_bytes in string_decoder.extract_delta_bytes(delta, ctx.decoded_at_va, fva):
                     for decoded_string in string_decoder.extract_strings(delta_bytes, min_length, no_filter):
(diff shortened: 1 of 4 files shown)


floss/main.py
```diff
@@ -941,7 +950,10 @@ def main(argv=None):
             print_identification_results(sample_file_path, decoding_functions_candidates)
 
         floss_logger.info("Decoding strings...")
-        decoded_strings = decode_strings(vw, decoding_functions_candidates, min_length, options.no_filter, options.max_instruction_count)
+        decoded_strings = decode_strings(vw, decoding_functions_candidates, min_length, options.no_filter,
+                                         options.max_instruction_count, options.max_address_revisits + 1)
+        # TODO: The de-duplication process isn't perfect as it is done here and in print_decoding_results and
+        # TODO: all of them on non-sanitized strings.
```
williballenthin: would you open issues against this repo for outstanding fixes?

floss/main.py
```diff
@@ -420,7 +423,7 @@ def filter_unique_decoded(decoded_strings):
     unique_values = set()
     originals = []
     for decoded in decoded_strings:
-        hashable = (decoded.va, decoded.s, decoded.decoded_at_va, decoded.fva)
+        hashable = (decoded.s, decoded.decoded_at_va, decoded.fva)
```
williballenthin: i presume that you made this fix because in practice our routine was *too* conservative in detecting duplicates?

floss/main.py
```diff
@@ -499,6 +502,12 @@ def print_decoding_results(decoded_strings, group_functions, quiet=False, expert
                     print("\nDecoding function at 0x%X (decoded %d strings)" % (fva, len_ds))
                 print_decoded_strings(grouped_strings, quiet=quiet, expert=expert)
     else:
+        if not expert:
+            seen = set()
+            decoded_strings = [x for x in decoded_strings if not (x.s in seen or seen.add(x.s))]
+        if not quiet:
+            print("\nFLOSS decoded %d strings" % len(decoded_strings))
```
williballenthin: at first i was confused, but now i see what you're doing here. nice!

floss/main.py
```diff
@@ -499,6 +502,12 @@ def print_decoding_results(decoded_strings, group_functions, quiet=False, expert
                     print("\nDecoding function at 0x%X (decoded %d strings)" % (fva, len_ds))
                 print_decoded_strings(grouped_strings, quiet=quiet, expert=expert)
     else:
+        if not expert:
+            seen = set()
+            decoded_strings = [x for x in decoded_strings if not (x.s in seen or seen.add(x.s))]
```
williballenthin: this is not the most readable implementation, but i appreciate how concise it is :-)

williballenthin:
awesome! thanks @BenjaminSoelberg!

floss/main.py
mr-tz: Removing the string va makes sense, although I'm not quite sure this filters out many duplicates.

floss/main.py
```diff
@@ -146,7 +147,9 @@ def make_parser():
                       help="do not filter deobfuscated strings (may result in many false positive strings)",
                       action="store_true")
     parser.add_option("--max-instruction-count", dest="max_instruction_count", type=int, default=20000,
-                      help="maximum number of instructions to emulate per function")
+                      help="maximum number of instructions to emulate per function (default is 20000)")
+    parser.add_option("--max-address-revisits", dest="max_address_revisits", type=int, default=0,
```
mr-tz: What's the reason this defaults to 0 and you later add 1 to it? As opposed to making the default value 1 and using that?

mr-tz:
Thank you for the improvements!

floss/main.py
williballenthin: my interpretation was that viv takes `max_hits` while the cli option is `max_revisits`, where the latter is `max_hits - 1`. keeping the terms the same would be more consistent, though i'm not sure which is easier to explain to a user on the cli.

floss/main.py
BenjaminSoelberg: Sure. I'll even fix it myself :-)

floss/main.py
BenjaminSoelberg: The reason for not having decode.va part of the hash is that even so two equal strings are decoded by the same function and at the same decoded.decoded_at_va they will have diffrent decoded.va addresses. I actually have no idea why. This only happens if the decoding is done in a loop which is why it is exposed when using the --max-address-revisits. Do you have any insight into why vivisect does this ?

floss/main.py
BenjaminSoelberg: I agree, not pretty but still pretty in its own way. There might be a better way to implement this de-keep-the-order-duplication

floss/main.py
BenjaminSoelberg: I really struggled to find a sane name.

PR #307, opened 2019-03-20
use complementary idb for further analysis

williballenthin:
via @BenjaminSoelberg


floss/main.py
```diff
@@ -922,7 +933,20 @@ def print_file_meta_info(vw, selected_functions):
...
+        for fva in api.idautils.Functions():
+            function_name = api.idc.GetFunctionName(fva)
+            if vw.isFunction(fva):
+                vw.makeName(fva, function_name)
+                floss_logger.debug('IDA Pro name imported : 0x%x:%s' % (fva, function_name))
+            else:
+                floss_logger.debug('IDA Pro name ignored  : 0x%x:%s' % (fva, function_name))
+
+
+def load_workspace(sample_file_path, complementary_idb_file_path, save_workspace):
```
williballenthin: lets make the idb path a kwarg, since its optional.

floss/main.py
```diff
@@ -955,13 +988,13 @@ def load_shellcode_workspace(sample_file_path, save_workspace, shellcode_ep_in,
                                            save_workspace, sample_file_path)
 
 
-def load_vw(sample_file_path, save_workspace, verbose, is_shellcode, shellcode_entry_point, shellcode_base):
+def load_vw(sample_file_path, complementary_idb_file_path, save_workspace, verbose, is_shellcode, shellcode_entry_point, shellcode_base):
```
williballenthin: my ultimate goal is to read the vw directly from an idb, rather  than from the complementary idb. we should detect the file type here and dispatch to an appropriate loader.

floss/main.py
```diff
@@ -1015,8 +1052,8 @@ def main(argv=None):
         return 1
 
     try:
-        vw = load_vw(sample_file_path, options.save_workspace, options.verbose, options.is_shellcode,
-                     options.shellcode_entry_point, options.shellcode_base)
+        vw = load_vw(sample_file_path, complementary_idb_file_path, options.save_workspace, options.verbose, options.is_shellcode,
```
williballenthin: this line is getting long, lets reformat it

floss/main.py
```diff
@@ -1015,8 +1052,8 @@ def main(argv=None):
...
+                     options.shellcode_entry_point, options.shellcode_base, )
```
williballenthin: remove trailing slash

floss/plugins/library_function_plugin.py
```diff
@@ -12,12 +12,13 @@ class FunctionIsLibraryPlugin(plugin_object.GeneralPlugin):
...
     def identify(self, vivisect_workspace, function_vas):
+        function_vas = function_vas - set(vivisect_workspace.getEntryPoints())
         candidate_functions = {}
         for fva in function_vas:
             fname = vivisect_workspace.getName(fva)
-            default_name = "sub_%.8x" % fva
-            if fname != default_name:
-                self.d("Identified %s at VA 0x%08X " % (fname, fva))
+            default_name = "sub_"
+            if fname is not None and not fname.startswith(default_name):
```
williballenthin: lets use `default_prefix` for this variable, instead

williballenthin:
@BenjaminSoelberg  added a bunch of comments. overall, this looks very good.

before we merge this, i'd like to try to get floss loading idbs directly, rather than as complementary files. 

there's an existing vivisect loader that works within IDA [here](https://github.com/williballenthin/viv-utils/blob/master/viv_utils/idaloader.py). however, i'm not sure that it'll work with python-idb right now. worst case, we can copy the code over and modify it within the floss project.

williballenthin:
i've been updating viv-utils to load an idb via python-idb. to do so, i had to add a bunch of new routines to python-idb. however, i've found some data structures that have changed recently([python-idb#63](https://github.com/williballenthin/python-idb/issues/63)), so i'll need to make some updates there before i can continue.

floss/main.py
```diff
@@ -311,12 +314,15 @@ def set_logging_levels(should_debug=False, should_verbose=False):
...
+
+    logging.getLogger("idb.netnode").setLevel(logging.WARN)
+    logging.getLogger("viv_utils.idaloader").setLevel(logging.WARN)
+
 
 def set_log_config(should_debug=False, should_verbose=False):
     """
     Removes root logging handlers, and sets Floss's logging level.
-    Recomended to use if Floss is being used in a standalone script, or 
+    Recomended to use if Floss is being used in a standalone script, or
```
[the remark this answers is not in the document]
BenjaminSoelberg: https://github.com/BenjaminSoelberg/flare-floss/commit/3dceb593f36123530e91007b410ca56502f18854

floss/main.py
```diff
@@ -922,7 +933,20 @@ def print_file_meta_info(vw, selected_functions):
         floss_logger.error("Failed to print vivisect analysis information: {0}".format(e.message))
 
 
-def load_workspace(sample_file_path, save_workspace):
+def populate_function_names(vw, complementary_idb_file_path):
```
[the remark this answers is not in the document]
BenjaminSoelberg: https://github.com/BenjaminSoelberg/flare-floss/commit/3dceb593f36123530e91007b410ca56502f18854

floss/main.py
BenjaminSoelberg: https://github.com/BenjaminSoelberg/flare-floss/commit/3dceb593f36123530e91007b410ca56502f18854

floss/main.py
BenjaminSoelberg: https://github.com/BenjaminSoelberg/flare-floss/commit/3dceb593f36123530e91007b410ca56502f18854

floss/plugins/library_function_plugin.py
BenjaminSoelberg: https://github.com/BenjaminSoelberg/flare-floss/commit/3dceb593f36123530e91007b410ca56502f18854

floss/main.py
BenjaminSoelberg: I agree, that would be a much better option.
I'll add it when https://github.com/williballenthin/python-idb has the required support.

floss/main.py
BenjaminSoelberg: I'm not sure that my Python foo is strong enough here.
Can you give an example on how that would look like ?

floss/main.py
BenjaminSoelberg: Also we would have to ensure that we could extract the 'original' binary from the idb to ensure the normal strings function would still work. I don't think the MZ and PE header are fully preserved but I might be wrong here.

BenjaminSoelberg:
The travis CI build seems to fail here.
I think a few things are the problem here, but I am unsure why it fails.
To make this build on my machine I had to pip install :
python-idb from https://github.com/williballenthin/python-idb/zipball/master
vivisect-vstruct from https://github.com/williballenthin/vivisect-vstruct/zipball/master
using virtualenv.

A new release of python-idb seems necessary ?
And vivisect-vstruct seems a bit more problematic to me as it seems to be an older version ?

Also I need to add tests for the new functionality in floss in order to ensure the quality of the new functionality. Naturally this should have been done beforehand.

Any advice would be much appreciated.

floss/api_hooks.py
```diff
@@ -118,66 +120,102 @@ def dumpState(self, emu):
...
     v = emu.readMemoryFormat(emu.getStackCounter(), "<P")[0]
     emu.setStackCounter(emu.getStackCounter() + pointerSize(emu))
     return v
 
 
-class GetProcessHeapHook(viv_utils.emulator_drivers.Hook):
-    '''
-    Hook and handle calls to GetProcessHeap, returning 0.
-    '''
+class LenientNamedHook(Hook):
```
williballenthin: i like this. nice job!

floss/api_hooks.py
```diff
@@ -118,66 +120,102 @@ def dumpState(self, emu):
...
+    def make_shortname(callname):
+        shortname = callname.lower()
+        shortname = shortname[shortname.find(".") + 1:]  # Note that if no dot is found, it will copy whole string
+        while shortname.startswith("_"):
+            shortname = shortname[1:]
+
+        return shortname
+
+
+class GetProcessHeapHook(LenientNamedHook):
```
williballenthin: thanks

floss/main.py
BenjaminSoelberg: https://github.com/BenjaminSoelberg/flare-floss/commit/d9d5f46090f312e62a49d58c4abcc4e1c417d8b1

XVilka:
Is this PR abandoned?

BenjaminSoelberg:
> Is this PR abandoned?

No not really. I have just been busy and have had some problems making it ready for build. But thank you for reminding me. I will work on finishing it.

XVilka:
Looking forward to it.

mr-tz:
Closing this stale PR. I've incorporated some of the API hook improvements in #546.

Issue #304, opened 2019-01-09
A domain/ip filter option would be great


BenjaminSoelberg:
May I suggest to add a string filtering option to the CLI. A build-in domain/ip filter could be provided.

Ana06:
You can do this combining floss with [grep](https://en.wikipedia.org/wiki/Grep). For example:

`floss my-file.exe | grep -E "([12]{0,1}[0-9]{1,2}\.){3}[12]{0,1}[0-9]{1,2}"`

which returns things like:
```
0.0.0.0
196.5.7.44
255.255.255.0
168.255.298.21
```

Those kind of regular expressions could be added in floss under a `suspicious` option which includes things like:

- urls
- IPs
- domains
- files
- registers

Even if the list is not perfect, it would save you some time as if the strings list is too long you most likely have to filter the output with grep (or in other way) for those cases.

I do not think though that adding an option only for domain/ip is a good idea, because if there was an option for every type of important string, the options list would get to long. I prefer to have just one option for everything. :thinking: 

@BenjaminSoelberg, @williballenthin what do you think? can I send a PR for the `suspicious` option?

mr-tz:
We've discussed this a while ago in #163. Back then we decided to not invent another "interesting strings database". However, I like the straight-forward solution provided in #312. Thanks, Ana!
@williballenthin thoughts?

williballenthin:
I agree that the implementation in #312 is an easy addition. 

It may be worth merging because it does add a concrete, useful feature to the tool.

My primary concern is about the maintenance required to maintain and update this "database". In practice, its probably not *hard*, though it duplicates work done elsewhere (TODO: where? I claim it does exist, but where can we actually look for prior art?).

williballenthin:
Here's a potentially alternative implementation: use [StringSifter](https://github.com/fireeye/stringsifter) to decide what should be highlighted.

the benefit is that we can rely on work done elsewhere and leave it to them to be experts.

the con is that it won't be easy to tweak (e.g. if StringSifter may not think `kernel32.dll` is _interesting_, but if its decoded as a stack string, it probably still is for FLOSS).

maybe we don't use StringSifter for filtering, but just for ordering (by default). 

Though it may appear so, I promise I'm not trying to push only FireEye tech on FLOSS. Just seemed like a relevant project that would be easy to integrate (`import stringsifter`).

Ana06:
@williballenthin 

> My primary concern is about the maintenance required to maintain and update this "database"

[...]

PR #312, opened 2019-10-02
Add an only-suspicious-option to filter strings

Ana06:
Add an `only-suspicious-option` which filters potential domains, urls, files (without including the Windows ones) and registers.

It may be possible to improve the regular expressions. I tried to keep the regular expressions simple (i.e. it returns strings similar to IPs that are not IPs such as `299.0.0.0`) to avoid making the code difficult to read/modify. I also made them strict (i.e. it doesn't return files with short names) to avoid returning too many strings. We may miss some strings, but we may also get the most important strings really fast. I tried it in a few files and it seems to work quite well. :tada: The list of Windows files is also not complete. I am not sure if we can get a complete list of files somewhere. :thinking: 

I need to write a test and documentation for the new option, but I would like to receive feedback about the feature before doing so.

Example:

```
> python floss/main.py --only-suspicious-strings ana_malware.exe
FLOSS static ASCII strings
anaisgreat.dll
\system32\anaisgreat.exe
http://anaisgreat.com/file.exe
192.178.63.1

FLOSS static UTF-16 strings


FLOSS decoded 2 strings
ana06.exe

FLOSS extracted 2 stackstrings
ana06.exe

```
Closes https://github.com/fireeye/flare-floss/issues/304

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -4,6 +4,7 @@
 
 from __future__ import print_function
 import os
+import re
 import sys
 import mmap
 import json
@@ -41,6 +42,16 @@ MAX_FILE_SIZE = 16 * MEGABYTE
 SUPPORTED_FILE_MAGIC = set(["MZ"])
 
 MIN_STRING_LENGTH_DEFAULT = 4
+
+WINDOWS_FILES = ['KERNEL32.DLL', 'GDI32.DLL', 'USER32.DLL', 'COMCTL32.DLL',
+                 'COMDLG32.DLL', 'WS2_32.DLL', 'ADVAPI32.DLL', 'NETAPI32.DLL',
+                 'OLE32.DLL', 'SHELL32.DLL', 'WUSER32.DLL', 'PSAPI.DLL',
+                 'MSVCRT.DLL', 'CRTDLL.DLL', 'MSCOREE.DLL']
+
+IS_IP = "^([12]{,1}\d{1,2}\.){3}[12]{,1}\d{1,2}"
+IS_DOMAIN_FILE = "^[\\\\\w\/:@\.-]{5,256}\.[a-zA-Z0-9]{1,4}"
+IS_REGISTER = "^(HKEY_LOCAL_MACHINE\\\\|HKLM\\\\).+$"
+IS_SUSPICIOUS = "(%s)|(%s)|(%s)" % (IS_IP, IS_DOMAIN_FILE, IS_REGISTER)
 
 
 class LoadNotSupportedError(Exception):
@@ -168,6 +179,8 @@ def make_parser():
                                 help="do not show decoded strings")
     extraction_group.add_option("--no-stack-strings", dest="no_stack_strings", action="store_true",
                                 help="do not show stackstrings")
+    extraction_group.add_option("--only-suspicious-strings", dest="only_suspicious_strings", action="store_true",
+                                help="only shows strings that look like an IP address")
     parser.add_option_group(extraction_group)
 
     format_group = OptionGroup(parser, "Format Options")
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -483,7 +496,7 @@ def print_identification_results(sample_file_path, decoder_results):
             headers=["address", "score"]))
 
 
-def print_decoding_results(decoded_strings, group_functions, quiet=False, expert=False):
+def print_decoding_results(decoded_strings, group_functions, quiet=False, expert=False, only_suspicious_strings=False):
```
williballenthin: these prototypes are getting a bit out of control (not an issue specific to this PR). lots of options and lots of possible combinations.

@mr-tz:

the code can be made more succinct if we use a config object (e.g. a dict of options); however, this doesn't mitigate the number of things to track.

we may want to spend a little time thinking about the different classes of output options and see if we can refactor here. the goal to make things concise and easy to understand.

for example, there may be multiple types of filter that we do (suspicious, static strings, unique, etc.) and maybe we could configure and implement these in a similar way. probably just needs some time dedicated to this.

floss/main.py
```diff
@@ -42,6 +43,16 @@
 
 MIN_STRING_LENGTH_DEFAULT = 4
 
+WINDOWS_FILES = ['KERNEL32.DLL', 'GDI32.DLL', 'USER32.DLL', 'COMCTL32.DLL',
+                 'COMDLG32.DLL', 'WS2_32.DLL', 'ADVAPI32.DLL', 'NETAPI32.DLL',
+                 'OLE32.DLL', 'SHELL32.DLL', 'WUSER32.DLL', 'PSAPI.DLL',
+                 'MSVCRT.DLL', 'CRTDLL.DLL', 'MSCOREE.DLL']
```
williballenthin: this is a good initial list. how did you pick these? should we add a comment saying "These are referenced in PMA" or similar?

floss/main.py
```diff
@@ -42,6 +43,16 @@
...
+
+IS_IP = "^([12]{,1}\d{1,2}\.){3}[12]{,1}\d{1,2}"
+IS_DOMAIN_FILE = "^[\\\\\w\/:@\.-]{5,256}\.[a-zA-Z0-9]{1,4}"
```
williballenthin: can you add comments that describe:

  1. how these work?
  2. examples of what they should match?

the IP regex is neat - it took me a minute or two to figure out how it works.

floss/main.py
```diff
@@ -42,6 +43,16 @@
...
+IS_REGISTER = "^(HKEY_LOCAL_MACHINE\\\\|HKLM\\\\).+$"
```
williballenthin: should be `IS_REGISTRY_HIVE`

also include:
  - HKCU
  - HKU

williballenthin: should anchoring at the start/end of the strings be consistent across all these patterns?

floss/main.py
williballenthin: this should be a Set rather than List

williballenthin:
I think the implementation here is fine. I'm not 100% convinced on the feature, but I look forward to more discussion on it. I'll add my thoughts to #304

floss/main.py
Ana06: @williballenthin 
> this is a good initial list. how did you pick these? should we add a comment saying "These are referenced in PMA" or similar?

Wikipedia + some that tend to appear in some files I tried this with. As said in the PR description:
> The list of Windows files is also not complete. I am not sure if we can get a complete list of files somewhere.

Ana06:
I haven't update the PR, because it seems it is not clear if this is the best solution. In case we decide this is wanted, I will update the PR. :wink: 

I have also realized of a few cases that are not cover and that would be really easy to add. For example, port numbers are not currently included.

mr-tz:
I propose to close this and come back to it once we've addressed the Python 3 cutover and #328

PR #321, opened 2020-04-15
add option to generate simple JSON report

b0urb0n:
Add `-o`/`--output-json` option to write the results of the analysis to a file in JSON format.

### DIFF
--- a/floss/decoding_manager.py
+++ b/floss/decoding_manager.py
@@ -21,10 +21,10 @@ MAX_MAPS_SIZE = 1024 * 1024 * 100  # 100MB max memory allocated in an emulator i
 DecodedString = namedtuple("DecodedString", ["va", "s", "decoded_at_va", "fva", "characteristics"])
 
 
-class LocationType(Enum):
-    STACK = 1
-    GLOBAL = 2
-    HEAP = 3
+class LocationType(str, Enum):
+    STACK = 'STACK'
+    GLOBAL = 'GLOBAL'
+    HEAP = 'HEAP'
 
 
 def is_import(emu, va):
--- a/floss/main.py
+++ b/floss/main.py
@@ -850,20 +850,23 @@ def create_r2_script(sample_file_path, r2_script_file, decoded_strings, stack_st
     # TODO return, catch exception in main()
 
 
-def create_json_output(json_file_path, sample_file_path, decoded_strings, stack_strings):
+def create_json_output(options, sample_file_path, decoded_strings, stack_strings):
     """
     Create a report of the analysis performed by FLOSS
-    :param json_file_path: path to write the report
+    :param options: parsed options
     :param sample_file_path: path of the sample analyzed
     :param decoded_strings: list of decoded strings ([DecodedString])
     :param stack_strings: list of stack strings ([StackString])
     """
-    results = {'stack_strings': [sanitize_string_for_printing(ss.s) for ss in stack_strings],
+    strings = {'stack_strings': [sanitize_string_for_printing(ss.s) for ss in stack_strings],
                'decoded_strings': [sanitize_string_for_printing(ds.s) for ds in decoded_strings]}
-    report = {'file_path': sample_file_path, 'results': results}
+    metadata = {'file_path': sample_file_path,
+                'stack_strings': not options.no_stack_strings,
+                'decoded_strings': not options.no_decoded_strings,
+                'static_strings': not options.no_static_strings}
     try:
-        with open(json_file_path, 'w') as f:
-            json.dump(report, f)
+        with open(options.json_file_path, 'w') as f:
+            json.dump({'metadata': metadata, 'strings': strings}, f)
     except Exception:
         raise
 
@@ -1114,7 +1117,7 @@ def main(argv=None):
         print("\nFinished execution after %f seconds" % (time1 - time0))
 
     if options.json_output_file:
-        create_json_output(options.json_output_file, sample_file_path, decoded_strings, stack_strings)
+        create_json_output(options, sample_file_path, decoded_strings, stack_strings)
         floss_logger.info("Wrote JSON file to %s\n" % options.json_output_file)
 
     return 0
--- a/floss/main.py
+++ b/floss/main.py
@@ -868,7 +868,7 @@ def create_json_output(options, sample_file_path, decoded_strings, stack_strings
                 'static_strings': not options.no_static_strings}
     report = {'metadata': metadata, 'strings': strings}
     try:
-        with open(options.json_file_path, 'w') as f:
+        with open(options.json_output_file, 'w') as f:
             json.dump(report, f, iterable_as_array=True)
     except Exception:
         raise
--- a/floss/main.py
+++ b/floss/main.py
@@ -80,6 +80,17 @@ def decode_strings(vw, decoding_functions_candidates, min_length, no_filter=Fals
     return decoded_strings
 
 
+def sanitize_strings_iterator(str_coll):
+    """
+    Iterate a collection and yield sanitized strings (uses sanitize_string_for_printing)
+    :param str_coll: collection of strings to be sanitized
+    :return: a sanitized string
+    """
+    for s_obj in str_coll:
+        s = getattr(s_obj, 's', s_obj)  # Use .s attribute from each namedtuple if possible
+        yield sanitize_string_for_printing(s)
+
+
 def sanitize_string_for_printing(s):
     """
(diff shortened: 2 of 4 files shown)


b0urb0n:
One thing I would like opinions on is whether or not to include static strings in the report.  In my experience, stack strings are hit and miss in terms of value.  This uncertainty in their value and the ease in which they can be reproduced lead me to omit them from the report.

Including them would make for a more complete report at the potential cost of consuming lots of memory.

b0urb0n:
Currently, the additional option behaves as:

```
(venv) 16:09 bourbon@scruffy [~/Development/code/flare-floss] $ floss --no-static-strings -o report.json /Users/bourbon/Desktop/malware/e2887a26ec82e8557b00ceb92e3b0ac5885933d9f66322a6ed7c8153abfe91d0

FLOSS decoded 5 strings
Runtime Error!
Program:
<program name unknown>
R6002
- floating point not loaded

FLOSS extracted 4 stackstrings
Fuck
KERNEL32.dll
q&>D
GetProcessHeap

Finished execution after 14.471431 seconds
(venv) 16:10 bourbon@scruffy [~/Development/code/flare-floss] $ cat report.json | json_pp
{
   "results" : {
      "decoded_strings" : [
         "Runtime Error!",
         "Program: ",
         "<program name unknown>",
         "R6002",
         "- floating point not loaded"
      ],
      "stack_strings" : [
         "Fuck",
         "KERNEL32.dll",
         "q&>D",
         "GetProcessHeap"
      ]
   },
   "file_path" : "/Users/bourbon/Desktop/malware/e2887a26ec82e8557b00ceb92e3b0ac5885933d9f66322a6ed7c8153abfe91d0"
}
```

mr-tz:
Sounds like a great additional feature!

I suggest that the JSON output respects the user-provided string types. By default it includes all strings and if needed types can be disabled via `--no-static-strings` etc.

Thoughts about adding the other available information such as offsets, decoding function VA, etc.?

b0urb0n:
> output respects the user-provided string types

I was thinking the report would always include keys for everything so the structure of the document stays constant. 

I think adding all the information we can would be beneficial. I'm curious how that should be done. 

My first thought was to have the report contain everything regardless of user input. This would prevent the user from ever needing to re-analyze. You can sort of do this already by saving the Vivisect workspace. The downside here is the potential size of the report. 

Somewhere in between probably makes most sense where we include strings the user asked for and some additional stuff that is commonly used (offsets, IDA/radare scripts?, Etc.)

mr-tz:
I see your point in having a consistent output document. However, different behavior between console output and JSON file could be confusing to users. For example "why does my JSON contain all strings if I only wanted decoded strings?"

One solution could be to add a "meta" part with boolean fields like `contains-static-strings`, etc.

b0urb0n:
I was playing around with that idea as well. I think a report manifest is a good idea.

b0urb0n:
The output report now respects the `--no-*-strings` flags and generates a metadata section that contains date of the analysis, the file path, and a manifest of string types.

```
$ floss --no-static-strings -o report.json /Users/bourbon/Desktop/malware/e2887a26ec82e8557b00ceb92e3b0ac5885933d9f66322a6ed7c8153abfe91d0

FLOSS decoded 5 strings
Runtime Error!
Program:
<program name unknown>
R6002
- floating point not loaded

FLOSS extracted 4 stackstrings
Fuck
KERNEL32.dll
q&>D
GetProcessHeap

Finished execution after 15.183981 seconds

$ cat report.json | json_pp
{
   "strings" : {
      "stack_strings" : [
         {
            "frame_offset" : 4,
            "sp" : 3216244720,
            "pc" : 4201201,
            "offset" : 8,
            "init_sp" : 3216244736,
            "s" : "Fuck",
            "fva" : 4201168
         },
         ...
      ],
      "decoded_strings" : [
         {
            "va" : 3216244572,
            "fva" : 4214304,
            "s" : "Runtime Error!",
            "characteristics" : {
               "location_type" : "STACK"
            },
            "decoded_at_va" : 4208482
         },
         ...
      ],
      "static_strings" : []
   },
   "metadata" : {
      "file_path" : "/Users/bourbon/Desktop/malware/e2887a26ec82e8557b00ceb92e3b0ac5885933d9f66322a6ed7c8153abfe91d0",
      "stack_strings" : true,
      "date" : "2020-04-21T11:48:20.874652",
      "decoded_strings" : true,
      "static_strings" : false
   }
}
```

floss/main.py
```diff
@@ -850,70 +852,62 @@ def create_r2_script(sample_file_path, r2_script_file, decoded_strings, stack_st
...
 
-def create_json_output(json_file_path, sample_file_path, decoded_strings, stack_strings):
+def create_json_output(options, sample_file_path, decoded_strings, stack_strings, static_strings):
     """
     Create a report of the analysis performed by FLOSS
-    :param json_file_path: path to write the report
+    :param options: parsed options
     :param sample_file_path: path of the sample analyzed
     :param decoded_strings: list of decoded strings ([DecodedString])
     :param stack_strings: list of stack strings ([StackString])
```
mr-tz, line 861: `:param static_strings` is missing here

floss/decoding_manager.py
```diff
@@ -21,10 +21,10 @@
 DecodedString = namedtuple("DecodedString", ["va", "s", "decoded_at_va", "fva", "characteristics"])
 
 
-class LocationType(Enum):
-    STACK = 1
-    GLOBAL = 2
-    HEAP = 3
+class LocationType(str, Enum):
```
mr-tz, line 24: neat!

floss/main.py
```diff
@@ -850,70 +852,62 @@ def create_r2_script(sample_file_path, r2_script_file, decoded_strings, stack_st
...
     """
-    results = {'stack_strings': [sanitize_string_for_printing(ss.s) for ss in stack_strings],
-               'decoded_strings': [sanitize_string_for_printing(ds.s) for ds in decoded_strings]}
-    report = {'file_path': sample_file_path, 'results': results}
+    strings = {'stack_strings': stack_strings,
```
mr-tz, line 864: Have you explored if strings need to be sanitized here?

floss/main.py
```diff
@@ -1027,7 +1021,18 @@ def main(argv=None):
...
         if not options.no_static_strings and not options.functions:
             floss_logger.info("Extracting static strings...")
-            print_static_strings(sample_file_path, min_length=min_length, quiet=options.quiet)
+            if os.path.getsize(sample_file_path) > sys.maxsize:
+                floss_logger.warning("File too large, strings listings may be truncated.")
+                floss_logger.warning("FLOSS cannot handle files larger than 4GB on 32bit systems.")
+
+            file_buf = get_file_as_mmap(sample_file_path)
+            print_static_strings(file_buf, min_length=min_length, quiet=options.quiet)
+            static_ascii_strings = strings.extract_ascii_strings(file_buf, min_length)
```
mr-tz, line 1031: can you extract these once and then use for printing and JSON exporting?

b0urb0n: My thought was to try to keep the memory footprint as small as possible with static strings. There can be a lot of them I chose to use a generator for printing and then create a new one for use with `chain` for JSON serialization. The use of a generator is also what brought me to choose `simplejson` over `json` from the standard lib.

I considered using `itertools.tee` to "fork" the generator but, after reading the documentation, if you let one of the "forks" get too far ahead of the other, you don't end up saving any memory and it becomes more efficient to just use lists or create two generators. 

Ultimately, I'd be looking to you to make the final call on this because I'm torn.

floss/main.py
b0urb0n: I gave it some thought and played around with turning the sanitize functions into generators (again to try to keep memory footprint low) but didn't end up finishing that because I couldn't come up with a good way to accomplish it.

I'll spend some more time thinking about how to solve that problem.

floss/main.py
mr-tz: Gotcha, thanks! Your explanation makes sense.

b0urb0n:
@mr-tz, making the requested changes to the `sanitize_strings_for*` functions is still on my radar.  I've been busier than normal during the COVID-19 times.  I'll get to this soon:tm:

mr-tz:
Awesome, thanks for the update and your work here!

floss/main.py
b0urb0n: Resolved by 2521339

mr-tz:
Thank you, @b0urb0n!

PR #329, opened 2020-06-11
Packaging and imports

b0urb0n:
The motivation for this PR was to streamline the install process and enable me to install FLOSS with:
```
pip install https://github.com/williballenthin/vivisect/zipball/master
pip install https://github.com/fireeye/flare-floss/archive/<tag>.zip
# or `python setup.py install`
```

What this PR changes:
* Fixed all package and relative imports
* Consolidated some reused constants into `const.py`
* Updated `setup.py` to enable running tests via `python setup.py pytest`
* ~~Created package-level `logger` and CLI-specific log-handler~~


floss/main.py
```diff
@@ -17,32 +18,19 @@
...
-MEGABYTE = 1024 * KILOBYTE
-MAX_FILE_SIZE = 16 * MEGABYTE
-
-SUPPORTED_FILE_MAGIC = set(["MZ"])
-
-MIN_STRING_LENGTH_DEFAULT = 4
+handler = logging.StreamHandler()
+handler.setLevel(logging.INFO)
+pkg_logger.addHandler(handler)
+logger = logging.getLogger(__name__)
```
williballenthin, line 33: i think there is a `configure_logging` (or similarly named) function below, in which we do all the logging config. lets keep all these logic in the same place, so its easy to track. i'm aware that some users do rely on this function, so i'd prefer to avoid breaking the interface, if possible.

floss/main.py
```diff
@@ -127,25 +111,25 @@ def get_all_plugins():
...
+        ps.append(function_meta_data_plugin.FunctionArgumentCountPlugin())
+        ps.append(function_meta_data_plugin.FunctionIsThunkPlugin())
+        ps.append(function_meta_data_plugin.FunctionBlockCountPlugin())
+        ps.append(function_meta_data_plugin.FunctionInstructionCountPlugin())
+        ps.append(function_meta_data_plugin.FunctionSizePlugin())
+        ps.append(function_meta_data_plugin.FunctionRecursivePlugin())
+        ps.append(library_function_plugin.FunctionIsLibraryPlugin())
+        ps.append(arithmetic_plugin.XORPlugin())
+        ps.append(arithmetic_plugin.ShiftPlugin())
+        ps.append(mov_plugin.MovPlugin())
```
williballenthin, line 124: these changes are valid syntactically, so no problem there.

for this project, since we have so many imports, we've been keeping the module prefix on most elements so that its clear where code/objects are coming from. all things being equal, let's keep it this way for consistency.

floss/utils.py
```diff
@@ -33,7 +33,7 @@ def removeStackMemory(emu):
             emu.setMemorySnap(memory_snap)
             emu.stack_map_base = None
             return
-    raise Exception  # STACK_MEM_NAME not in memory map
+    raise ValueError("`STACK_MEM_NAME` not in memory map")
```
williballenthin, line 36: good catch

setup.py
```diff
@@ -45,6 +43,12 @@
             "floss=floss.main:main",
         ]
     },
+    setup_requires=[
+        'pytest-runner',
+    ],
+    tests_require=[
+        'pytest',
+    ],
```
williballenthin, line 51: thanks!

floss/main.py
```diff
@@ -2,13 +2,14 @@
...
 
-from __future__ import print_function
 import os
 import sys
 import mmap
 import string
 import logging
 import datetime
+
+from base64 import b64encode
```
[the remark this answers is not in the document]
williballenthin: no way you could have known this unless you sat in Moritz and I's 1:1 calls :-)


though, as a sidebar, i do maintain a personal style guide here:
https://docs.google.com/document/d/1iRpeg-w4DtibwytUyC_dDT7IGhNGBP25-nQfuBa-Fyk/edit?usp=sharing

b0urb0n:
@williballenthin I'll try and get these to you today!

floss/__init__.py
```diff
@@ -1,6 +1,12 @@
...
-from utils import ONE_MB
-from utils import STACK_MEM_NAME
-from utils import makeEmulator
-from utils import removeStackMemory
+import logging
+
+from . import version
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
b0urb0n: If you always use `logging.getLogger(__name__)`, the loggers will be kept separate and will propagate to the top-level logger.  This is why I've imported the package `logger` in the CLI app and set a `SteamHandler` to the root `logger` instead of just the `main` module's `logger`.  Having a single root-level package `logger` also allows folks who use `floss` as a module to control FLOSS' logging from one place.

The package-level logger will be named `floss` and the CLI app logger is named `floss.main` because of how `__name__` works.  Log events created in `floss.main` will naturally propagate to ancestor loggers (`floss` in this case).

I'm happy to change this if you want it changed but this is how I've handled logging in very large projects and this is also what the logging library documentation recommends in the [advanced logging tutorial](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial).

I'll look to you for the final call here.

b0urb0n:
Looks like the build is failing because of the pinned vivisect library not being satisfied and not being in PyPI.  This dependency should be handled by the following line in your `.appveyor.yml`:
```
    - "%PYTHON%\\Scripts\\pip install https://github.com/williballenthin/vivisect/zipball/master"
```
However, your fork of Vivisect is only `version='0.0.20200225',` according to `setup.py` which is ~4 months older than the pinned version.

williballenthin:
ugh, i keep making this mistake - making a GH release but forgetting to update the setup.py file. let me do that now.

williballenthin:
ok, updated vivisect/setup.py. would you mind trying again?

b0urb0n:
I merged in your changes from `master`.  That kicked off another CI/CD task.

PR #336, opened 2020-07-06
Fix --ida, broken in IDA >= 7.4

### LINKED ISSUE #335: --ida option doesn't work with IDA 7.4
Some of the used methods have been renamed in IDA 7.4: https://www.hex-rays.com/products/ida/support/ida74_idapython_no_bc695_porting_guide.shtml

It is not difficult to fix this, the question is: do we want to keep this option?

If Floss could run in Python3, a better approach would be to just provide a json output, as an ida script/plunging could call floss. Taking into account that this is not an option and that this script is really useful, I would fix it. Most likely we want to have a plugin for < 7.3 and one for >= 7.4.

Related to https://github.com/fireeye/flare-floss/issues/328

Ana06:
Some of the used methods have been renamed in IDA 7.4:
https://www.hex-rays.com/products/ida/support/ida74_idapython_no_bc695_porting_guide.shtml
Use the new methods to be able to run the generated script in IDA >= 7.4.

Python3 is the default in IDA 7.4. Adapt the script to work in Python3 as well.

Closes #335

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -611,7 +611,6 @@ def create_ida_script_content(sample_file_path, decoded_strings, stack_strings):
 
 
 def AppendComment(ea, s, repeatable=False):
-    # see williutils and http://blogs.norman.com/2011/security-research/improving-ida-analysis-of-x64-exception-handling
     if repeatable:
         string = RptCmt(ea)
     else:
--- a/floss/main.py
+++ b/floss/main.py
@@ -592,20 +592,20 @@ def create_ida_script_content(sample_file_path, decoded_strings, stack_strings):
         if ds.s != "":
             sanitized_string = sanitize_string_for_script(ds.s)
             if ds.characteristics["location_type"] == LocationType.GLOBAL:
-                main_commands.append("print(\"FLOSS: string \\\"%s\\\" at global VA 0x%X\")" % (sanitized_string, ds.va))
-                main_commands.append("AppendComment(%d, \"FLOSS: %s\", True)" % (ds.va, sanitized_string))
+                main_commands.append('print("FLOSS: string \\"%s\\" at global VA 0x%X")' % (sanitized_string, ds.va))
+                main_commands.append('AppendComment(%d, "FLOSS: %s", True)' % (ds.va, sanitized_string))
             else:
-                main_commands.append("print(\"FLOSS: string \\\"%s\\\" decoded at VA 0x%X\")" % (sanitized_string, ds.decoded_at_va))
-                main_commands.append("AppendComment(%d, \"FLOSS: %s\")" % (ds.decoded_at_va, sanitized_string))
-    main_commands.append("print(\"Imported decoded strings from FLOSS\")")
+                main_commands.append('print("FLOSS: string \\"%s\\" decoded at VA 0x%X")' % (sanitized_string, ds.decoded_at_va))
+                main_commands.append('AppendComment(%d, "FLOSS: %s")' % (ds.decoded_at_va, sanitized_string))
+    main_commands.append('print("Imported decoded strings from FLOSS")')
 
     ss_len = 0
     for ss in stack_strings:
         if ss.s != "":
             sanitized_string = sanitize_string_for_script(ss.s)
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -592,65 +592,58 @@ def create_ida_script_content(sample_file_path, decoded_strings, stack_strings):
         if ds.s != "":
             sanitized_string = sanitize_string_for_script(ds.s)
             if ds.characteristics["location_type"] == LocationType.GLOBAL:
-                main_commands.append("print \"FLOSS: string \\\"%s\\\" at global VA 0x%X\"" % (sanitized_string, ds.va))
+                main_commands.append("print(\"FLOSS: string \\\"%s\\\" at global VA 0x%X\")" % (sanitized_string, ds.va))
                 main_commands.append("AppendComment(%d, \"FLOSS: %s\", True)" % (ds.va, sanitized_string))
             else:
-                main_commands.append("print \"FLOSS: string \\\"%s\\\" decoded at VA 0x%X\"" % (sanitized_string, ds.decoded_at_va))
+                main_commands.append("print(\"FLOSS: string \\\"%s\\\" decoded at VA 0x%X\")" % (sanitized_string, ds.decoded_at_va))
```
williballenthin, line 598: it might be easier to use single quotes here to reduce the escaping

floss/main.py
```diff
@@ -592,65 +592,58 @@ def create_ida_script_content(sample_file_path, decoded_strings, stack_strings):
...
-    print "Failed to append stackstring comment \\\"%%s\\\" at stack frame offset 0x%%X in function 0x%%X" %% (s, frame_offset, fva)
+    print("Failed to append stackstring comment \\\"%%s\\\" at stack frame offset 0x%%X in function 0x%%X" %% (s, frame_offset, fva))
 
 
 def main():
-    print "Annotating %d strings from FLOSS for %s"
+    print("Annotating %d strings from FLOSS for %s")
     %s
-    Refresh()
+    ida_kernwin.refresh_idaview_anyway
```
williballenthin, line 646: should this be a function call?

floss/main.py
```diff
@@ -592,65 +592,58 @@ def create_ida_script_content(sample_file_path, decoded_strings, stack_strings):
...
     for ss in stack_strings:
         if ss.s != "":
             sanitized_string = sanitize_string_for_script(ss.s)
             main_commands.append("AppendLvarComment(%d, %d, \"FLOSS stackstring: %s\", True)" % (ss.fva, ss.frame_offset, sanitized_string))
             ss_len += 1
-    main_commands.append("print \"Imported stackstrings from FLOSS\"")
+    main_commands.append("print(\"Imported stackstrings from FLOSS\")")
 
-    script_content = """from idc import RptCmt, Comment, MakeRptCmt, MakeComm, GetFrame, GetFrameLvarSize, GetMemberComment, SetMemberComment, Refresh
+    script_content = """get_cmt, set_cmt, get_func_attr, get_member_cmt, set_member_cmt
```
mr-tz, line 610: should these be imported here?

Ana06: It seems it is not needed to import them 🙄

floss/main.py
Ana06: yes 👍 👍 it was failing, but It didn't raise an error in IDA and I hadn't realized

floss/main.py
```diff
@@ -592,65 +592,56 @@ def create_ida_script_content(sample_file_path, decoded_strings, stack_strings):
         if ds.s != "":
             sanitized_string = sanitize_string_for_script(ds.s)
             if ds.characteristics["location_type"] == LocationType.GLOBAL:
-                main_commands.append("print \"FLOSS: string \\\"%s\\\" at global VA 0x%X\"" % (sanitized_string, ds.va))
-                main_commands.append("AppendComment(%d, \"FLOSS: %s\", True)" % (ds.va, sanitized_string))
+                main_commands.append('print("FLOSS: string \\"%s\\" at global VA 0x%X")' % (sanitized_string, ds.va))
+                main_commands.append('AppendComment(%d, "FLOSS: %s", True)' % (ds.va, sanitized_string))
```
williballenthin, line 596: nice, much easier to read

Ana06:
We may want to release FLOSS 1.7 🤔

Issue #347, opened 2021-01-02
Strings.py --- TypeError: %b requires a bytes-like object, or an object that implements __bytes__, not 'str'


jedimasterbot:
The regex is bytes based on the Unicode strings. If remove the bytes type regex to simple then **TypeError: cannot use a string pattern on a bytes-like object** occurs as the file is read in form of bytes.

jedimasterbot:
Found the fix, made the pull request waiting for approval

williballenthin:
sounds like you're trying to use this with python3? python2 doesn't make a distinction between str/bytes. at the moment, floss only works under python2.

PR #348, opened 2021-01-02
Update strings.py

jedimasterbot:
Fix for the issue  #347, Converting all the regex into bytes based and the issue is resolved, checked for python version 3.73

https://github.com/fireeye/flare-floss/issues/347

### DIFF
--- a/floss/strings.py
+++ b/floss/strings.py
@@ -4,9 +4,9 @@ import re
 from collections import namedtuple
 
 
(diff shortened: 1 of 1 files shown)


williballenthin (approved):
this is a good fix, and we'll accept it. 

keep in mind, though, that floss is currently python2-only, so i dont expect that just these fixes are sufficient to run floss under python3.

PR #368, opened 2021-05-26
update ci build

mr-tz:
current CI build fails, see build Action

```
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007f06624a3700 (most recent call first):
<no Python frame>
```


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
```
mr-tz, line 1: unrelated addition

mr-tz:
Who can make sense of this? All my research lead nowhere. Build and spec are almost identical to `capa` which works fine.

williballenthin:
https://github.com/pyinstaller/pyinstaller/issues/5438

williballenthin:
image

the `encodings` module does seem to be present i8n `base_library.zip`.

williballenthin:
see #371 that duplicates and extends this PR

PR #371, opened 2021-06-09
update ci build extending #368

williballenthin:
supersedes #368

i took the nuclear option: i regenerated the pyinstaller spec using the 4.3 release, verified that it produced a working ELF binary, and then copied our configuration into it one-by-one. ultimately, im not actually sure where our problem lies - but it works now.


williballenthin:
image
this section here passes in an additional parameter `a.datas` that we didn't have, so i wonder if that's why the `encodings` module wasn't available? just a guess, not going to dig any deeper.

williballenthin:
image
w00t

mr-tz:
w00t w00t! I really appreciate you investigating this! This is really cool and I'm relieved it's working now - even without exactly knowing what went wrong.

.github/pyinstaller/floss.spec
```diff
@@ -1,177 +1,83 @@
...
+    upx_exclude=[],
+    runtime_tmpdir=None,
+    console=True,
+)
 
+# enable the following to debug the contents of the .exe
+# writes to ./dist/floss-dat
+coll = COLLECT(
+    exe, a.binaries, a.zipfiles, a.datas, strip=None, upx=True, name="floss-dat"
+)
```
mr-tz, line 83:
```suggestion
)

```

.github/workflows/build.yml
```diff
@@ -1,6 +1,9 @@
 name: build
 
 on:
+  # for testing
+  pull_request:
+    branches: [master]
```
mr-tz, line 6: remove before merge once working

mr-tz (approved):

.github/workflows/build.yml
williballenthin:
```suggestion
```

Issue #385, opened 2021-07-21
Provide json as base output


mr-tz:
FLOSS should use a standardized json base output to enhance interoperability with external tools and ease the various output modes (cmd line, IDA script, etc.).

related: #188

mr-tz:
As a counter point (?) it would be nice if FLOSS would output strings as it encounters/deobfuscates them.

williballenthin:
> As a counter point (?) it would be nice if FLOSS would output strings as it encounters/deobfuscates them.

we could use logging status messages to STDERR for this and keep the final output to STDOUT at the end

PR #398, opened 2021-07-21
standardized result document

### LINKED ISSUE #397: use dataclasses as appropriate
this are useful for simple product types (like namedtuples/dicts)

https://docs.python.org/3/library/dataclasses.html

with backport to 3.6 here: https://pypi.org/project/dataclasses/

williballenthin:
refactor to use a common, well-typed "result document". notably, this result document has the same structure as the json report. as described in #385, this json report will the primary mechanism for integration. so, this result document forces us to "dog feed" our design; if floss core is missing something from the result document, then other integrations may as well.

closes #397 (dataclasses)

### DIFF
--- a/floss/decoding_manager.py
+++ b/floss/decoding_manager.py
@@ -1,7 +1,6 @@
 # Copyright (C) 2017 FireEye, Inc. All Rights Reserved.
 
 import logging
-from enum import Enum
 from collections import namedtuple
 
 import viv_utils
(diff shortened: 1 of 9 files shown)


williballenthin:
@mr-tz  review and merge OR when i next go to dev i'll mark this as draft again and keep going from here.

i think we'll end up accumulating fairly large PRs if we wait each feature to be implemented (lots of stuff to fix along the way). thats why i'm ok with incremental PRs that are "better than before but not perfect".

mr-tz (approved):
nice, no objections here besides the failing code style and tests

PR #421, opened 2021-08-25
Use relative path for testfiles submodule

calfr:
With a similar rational to https://github.com/fireeye/capa/issues/255, looks to change the path for submodules from absolute (which forces a ssh based clone) to relative.

Note a downside mentioned within the issue stating that developers would need to fork the testfiles repository as well while developing.


williballenthin:
thank you!

PR #452, opened 2021-10-22
add tightloop string extraction

mr-tz:
Adds tightstring extraction:
- identify functions with tight loops
- emulate similar to stack strings algorithm
- extracts tight string contexts after each tight loop executed
- filters out strings that were present before loop executed

There's still room for improvement but in preliminary tests the algorithm successfully extracts the desired strings.

This PR also adds a couple of random unrelated improvements (sorry for the mix up).

### DIFF
--- a/floss/features/extract.py
+++ b/floss/features/extract.py
@@ -24,7 +24,7 @@ SECURITY_COOKIE_BYTES_DELTA = 0x40
 
 SHIFT_ROTATE_INS = (INS_SHL, INS_SHR, INS_ROL, INS_ROR)
 
-logger = logger = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
 
 
 def extract_insn_nzxor(f, bb, insn):
--- a/floss/features/extract.py
+++ b/floss/features/extract.py
@@ -1,16 +1,30 @@
+import logging
+
 import envi
 import networkx
 import vivisect
 from networkx import strongly_connected_components
 from envi.archs.i386.opconst import INS_MOV, INS_ROL, INS_ROR, INS_SHL, INS_SHR, INS_XOR
 
-from floss.features.features import Mov, Loop, Nzxor, Shift, CallsTo, NzxorLoop, TightLoop, NzxorTightLoop
+from floss.features.features import (
+    Mov,
+    Loop,
+    Nzxor,
+    Shift,
+    CallsTo,
+    NzxorLoop,
+    TightLoop,
+    KindaTightLoop,
+    NzxorTightLoop,
+)
 
 # security cookie checks may perform non-zeroing XORs, these are expected within a certain
 # byte range within the first and returning basic blocks, this helps to reduce FP features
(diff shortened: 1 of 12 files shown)


floss/features/extract.py
```diff
@@ -1,17 +1,31 @@
...
+    NzxorTightLoop,
+)
 
 # security cookie checks may perform non-zeroing XORs, these are expected within a certain
 # byte range within the first and returning basic blocks, this helps to reduce FP features
 SECURITY_COOKIE_BYTES_DELTA = 0x40
 
 SHIFT_ROTATE_INS = (INS_SHL, INS_SHR, INS_ROL, INS_ROR)
 
+logger = logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
mr-tz: whoops, good catches :)

floss/features/extract.py
```diff
@@ -97,13 +111,58 @@ def extract_function_calls_to(f):
...
+            continue
+
+        # find the intermediate "loop" block that makes this "kinda" tight
+        loop_bb = None
+        for suc in succs:
+            if bb.va in [s.va for s in f.get_bb_succs(suc)]:
+                # successor must only have one successor itself
+                # successor can be the block itself (that's fine)
+                # without these rules it leaves room for weird corner cases
+                if len(list(f.get_bb_succs(suc))) == 1 or bb.va == suc.va:
```
williballenthin, line 136: we should save off `f.get_bb_succs(suc)` in a temporary variable here to reduce the number of times this is computed

floss/features/extract.py
```diff
@@ -97,13 +111,58 @@ def extract_function_calls_to(f):
...
+    Yields tight loop features in the provided function
+    Algorithm by Blaine S.
+    """
+    for bb in f.basic_blocks:
+        # skip first and last BBs
+        if bb.va == f.va or bb.va + bb.size == f.basic_blocks[-1].va:
+            continue
+
+        # skip blocks that don't have exactly 2 successors
+        succs = list(f.get_bb_succs(bb))
```
williballenthin, line 125: there are a lot of calls to `f.get_bb_succs(...)` in this function. we should cache the results, perhaps explicitly in a local variable here or alternatively wrap the method with a caching decorator as described [here](https://github.com/williballenthin/viv-utils/pull/20#discussion_r734582384)

floss/features/extract.py
```diff
@@ -97,13 +111,58 @@ def extract_function_calls_to(f):
...
+        if len(succs) != 2:
+            continue
+
+        # find the intermediate "loop" block that makes this "kinda" tight
+        loop_bb = None
+        for suc in succs:
+            if bb.va in [s.va for s in f.get_bb_succs(suc)]:
```
williballenthin, line 132:
```py
if bb.va in {s.va for s in f.get_bb_succs(suc)}:
```

we should use a set here for *probably* faster lookups

williballenthin:
@mr-tz  do you have a good test case for tight loops?

PR #458, opened 2021-11-05
extract: use viv-utils.CFG instead of bb.succs

williballenthin:
use `CFG` to compute and access basic block successors. if nothing else, this is about 40% faster than BB succs (which we can remove from viv-utils, again).

also, add documentation explaining the patterns we're looking for.

tested with [b0b758e680e652144a78a7ddecc027d4868c1dc3d8d7d611ec4d3798358b0ce5](https://www.virustotal.com/gui/file/b0b758e680e652144a78a7ddecc027d4868c1dc3d8d7d611ec4d3798358b0ce5) and it seems to extract reasonable strings. 

image

### DIFF
--- a/floss/features/extract.py
+++ b/floss/features/extract.py
@@ -3,6 +3,7 @@ import logging
 import envi
 import networkx
 import vivisect
+import viv_utils
 from networkx import strongly_connected_components
 from envi.archs.i386.opconst import INS_MOV, INS_ROL, INS_ROR, INS_SHL, INS_SHR, INS_XOR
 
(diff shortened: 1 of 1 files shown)


floss/features/extract.py
```diff
@@ -116,32 +117,75 @@ def extract_function_kinda_tight_loop(f):
...
+            for suc in succs:
+                suc_succs = [x for x in cfg.get_successor_basic_blocks(suc)]
+                if len(suc_succs) != 1:
+                    continue
+                if suc_succs[0] != bb.va:
+                    continue
+
+                loop_bb = suc_succs[0]
+                break
+
```
williballenthin, line 184: i split out the two cases here to make it more explicit how we check for them

floss/features/extract.py
```diff
@@ -116,32 +117,75 @@ def extract_function_kinda_tight_loop(f):
...
     Algorithm by Blaine S.
     """
+    cfg = viv_utils.CFG(f)
+
+    root_bb_va = cfg.get_root_basic_block()
+    leaf_bb_vas = {bb.va for bb in cfg.get_leaf_basic_blocks()}
+
     for bb in f.basic_blocks:
         # skip first and last BBs
-        if bb.va == f.va or bb.va + bb.size == f.basic_blocks[-1].va:
```
williballenthin, line 121: bug here that the "last" basic block may not be end of the function. 

now we use the "leaves" that should be the ends of the function.

floss/features/extract.py
```diff
@@ -116,32 +117,75 @@ def extract_function_kinda_tight_loop(f):
     Yields tight loop features in the provided function
     Algorithm by Blaine S.
     """
+    cfg = viv_utils.CFG(f)
```
mr-tz, line 120: I'm still wondering how I missed this 😄

mr-tz (approved):
awesome, thanks for the improvement!

PR #492, opened 2022-02-04
Change to ida7 and adjust to floss changing

harelon:
Changed to use the api of ida7.
Removed encoding and encoding errors because s is already a string.
Changed field names to the ones in the current floss package
Updated vivisect version to be compatible with the current viv-utils supported version


mr-tz:
Hey, thanks a lot! I was wondering if you were also looking at this because of the viv-utils updates :)
Was just looking at this script yesterday.

williballenthin (approved):

PR #539, opened 2022-05-13
doc: update copyright header

mr-tz:


mr-tz:
I'll update per file creation years.

williballenthin:
I don’t feel strongly about keeping these up to date or making more work for you. Just wondering if we should grep for all of them and update everything to 2022.

PR #546, opened 2022-05-24
feat: update and add hooks

mr-tz:
also brings in some ideas from #307

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -130,6 +130,7 @@ def make_parser(argv):
         "-n",
         "--minimum-length",
         dest="min_length",
+        type=int,
         default=MIN_STRING_LENGTH,
         help="minimum string length",
     )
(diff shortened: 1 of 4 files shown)


scripts/render-ida-import-script.py
```diff
@@ -72,7 +72,7 @@ def AppendComment(ea, string, repeatable=False):
     else:
         if string in current_string:  # ignore duplicates
             return
-        cmt = string + "\\n" + string
+        cmt = current_string + "\\n" + string
```
mr-tz, line 75: unrelated fix, sorry

floss/main.py
```diff
@@ -130,6 +130,7 @@ def make_parser(argv):
         "-n",
         "--minimum-length",
         dest="min_length",
+        type=int,
```
mr-tz, line 133: unrelated fix, sorry

williballenthin (approved):

Issue #560, opened 2022-06-22
sigs does not exist or cannot be accessed


mr-tz:
Basic install with pip and I get the following error:

```
OSError: signatures path .../envs/py37/lib/python3.7/site-packages/floss/../sigs does not exist or cannot be accessed
```

See same issue in capa at https://github.com/mandiant/capa/issues/682

rholloway:
👍  same issue. The standalone binaries have sigs functioning and no error but 2.0.0 on Pypi is missing sigs directory entirely in the package

PR #561, opened 2022-06-23
Moving sigs directory and add MANIFEST-file

3c7:
Potential fix for #560, hope i haven't overlooked any hardcoded `sigs` paths.

- Moved `sigs` to `floss` directory
- Addes MANIFEST.in file to include README, LICENSE and sigs


mr-tz:
Thanks for the solution idea. To make the tool easy to use it makes a lot of sense to include the signatures in the default installation. I'll investigate this more this week.

mr-tz:
Can we accomplish the same and leave `sigs` where it currently resides?

3c7:
I don't think so (but am not 100% sure - not super familiar with including non-source files in python modules). During installation, only the `floss` directory will be installed into the `site-packages` directory. While the other files and directories can be included in the distribution files, they will be skipped during installation.

mr-tz:
Thanks, that makes sense. We can move the directory.

mr-tz:
handled via #578

PR #566, opened 2022-07-02
Feature/large file support

PurplProto:


PurplProto:
Unintentional PR. Intended for my master branch.

Happy to PR just the commit 14c2a69 if it would be accepted though. It appears it wouldn't be from the issues I've read around the file size limitation though.

williballenthin:
Hey @PurplProto 

We'd be happy to accept a PR that adds the large file flag! Thanks for taking the time to write up the code (and all the related CI/CD work in your repo :-).

We have the limitation there because in most cases, most analysts probably don't want to analyze huge files. But with such a flag, the user can assert they know what they're doing and understand why processing might take a long time. If the flag would be useful to you, its probably useful to other people too, so lets merge it.

Want to open a PR?

floss/main.py
```diff
@@ -556,9 +565,12 @@ def main(argv=None) -> int:
...
     ):
-        if os.path.getsize(sample) > MAX_FILE_SIZE:
-            logger.error("cannot deobfuscate strings from files larger than 0x%x bytes", MAX_FILE_SIZE)
-            return -1
+        if sample_size > MAX_FILE_SIZE:
+            if not args.large_file:
+                logger.error("cannot deobfuscate strings from files larger than 0x%x bytes", MAX_FILE_SIZE)
+                return -1
+            else:
+                logger.warning("a large file was provided with a size of %i bytes, this may take much more time to process and require more system resource", sample_size)
```
williballenthin, line 573: i like this message that indicates "even though you provided the flag, remember this might take a while"

floss/main.py
```diff
@@ -142,6 +142,13 @@ def make_parser(argv):
         help="minimum string length",
     )
 
+    parser.add_argument(
+        "-L",
+        "--large-file",
+        action="store_true",
+        help="allow processing files larger than {} MB".format(int(MAX_FILE_SIZE / MEGABYTE))
```
mr-tz, line 149: nice, please also add `if show_all_options else argparse.SUPPRESS` here so this shows up under the "advanced" options via `-H`

PurplProto: Fixed in #568 🙂

floss/main.py
PurplProto: Opened PR #568 ! 😊

PR #578, opened 2022-08-16
Package signatures

### LINKED ISSUE #560: sigs does not exist or cannot be accessed

mr-tz:
closes #560
based on  #561


mr-tz:
@3c7 does this solution based on your PR work for you?

3c7:
It works! However, `python3 setup.py build` raises an error because `floss.sigs` is not registered as a package. This can be fixed through adding `floss.sigs` as a package

*setup.py*
```python
packages=setuptools.find_packages().append("floss.sigs"),
```

or adding an empty `__init__.py` file into `floss/sigs`.

I'm not sure, if this is relevant, as the installation via `pip install git+https://github.com/mr-tz/flare-floss@pkg/signatures` also works flawlessly.

mr-tz:
Hm, for me `python setup.py build` seems to include the files?

```
...
copying floss\sigs\1_flare_msvc_rtf_32_64.sig -> build\lib\floss\sigs
copying floss\sigs\2_flare_msvc_atlmfc_32_64.sig -> build\lib\floss\sigs
copying floss\sigs\3_flare_common_libs.sig -> build\lib\floss\sigs
...
```

3c7:
Don't you get the below warning?

```
    ############################
    # Package would be ignored #
    ############################
    Python recognizes 'floss.sigs' as an importable package,
    but it is not listed in the `packages` configuration of setuptools.

    'floss.sigs' has been automatically added to the distribution only
    because it may contain data files, but this behavior is likely to change
    in future versions of setuptools (and therefore is considered deprecated).
```

However, for now the package works as-is.

mr-tz:
I did get the warning after updating `setuptools`. Thank you. How about now?

3c7:
Yep, that did the trick. Working as expected now.

mr-tz:
Thank you very much. I'll create a new release.

PR #642, opened 2023-03-02
Update tests.yml

ggold7046:
Updated deprecated actions.


williballenthin:
```
Error: Unable to resolve action `actions/checkout@v4`, unable to find version `v4`
```

https://github.com/mandiant/flare-floss/actions/runs/4312858250/jobs/7523938321

PR #646, opened 2023-03-03
Added text to subheadings that specifies the heading they come under.

### LINKED ISSUE #621: output headings may be confusing
today FLOSS outputs headings like the following to separate its results:

```console
-------------------------------
| FLOSS STATIC STRINGS (8188) |
-------------------------------
------------------------------
| FLOSS ASCII STRINGS (7298) |
------------------------------
!This program cannot be run in DOS mode.
```

when I ran FLOSS, I was confused by a few things:

  1. "FLOSS ASCII strings" are not decode strings but static strings
  2. the second section is a subsection of the first section

I think we should rename the titles to clarify this relationship. Perhaps we can use something like:

```console
-------------------------------
| FLOSS STATIC STRINGS (8188) |
-------------------------------
--------------------------------------
| FLOSS STATIC STRINGS: ASCII (7298) |
--------------------------------------
!This program cannot be run in DOS mode.
...
----------------------------------------
| FLOSS STATIC STRINGS: UTF-16LE (...) |
----------------------------------------
...
```

Shackles404:
Currently the subheadings for static strings look like:
```
-------------------------------
| FLOSS STATIC STRINGS (8188) |
-------------------------------
------------------------------
| FLOSS ASCII STRINGS (7298) |
------------------------------
!This program cannot be run in DOS mode.
```

Changed them to:

```
-------------------------------
| FLOSS STATIC STRINGS (3119) |
-------------------------------
--------------------------------------
| FLOSS STATIC STRINGS: ASCII (3119) |
--------------------------------------
...
--------------------------------------
| FLOSS STATIC STRINGS: UTF-16LE (0) |
--------------------------------------
```

I wonder though, could we also use another table format for the subheadings to make them stand out from the headings? I tried looking around in tabulate but didn't feel like any format would appear "lighter" than the ones currently in use for headings.

resolves #621

### DIFF
--- a/floss/render/default.py
+++ b/floss/render/default.py
@@ -134,7 +134,7 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
         unicode_offset_len = len(f"{unicode_strings[-1].offset}")
     offset_len = max(ascii_offset_len, unicode_offset_len)
 
-    render_heading("FLOSS ASCII STRINGS", len(ascii_strings), ostream, disable_headers)
+    render_heading("FLOSS STATIC STRINGS: ASCII", len(ascii_strings), ostream, disable_headers)
     for s in ascii_strings:
         if verbose == Verbosity.DEFAULT:
             ostream.writeln(s.string)
@@ -142,7 +142,7 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
             ostream.writeln(f"0x{s.offset:>0{offset_len}x} {s.string}")
     ostream.writeln("")
 
-    render_heading("FLOSS UTF-16LE STRINGS", len(unicode_strings), ostream, disable_headers)
+    render_heading("FLOSS STATIC STRINGS: UTF-16LE", len(unicode_strings), ostream, disable_headers)
     for s in unicode_strings:
         if verbose == Verbosity.DEFAULT:
             ostream.writeln(s.string)
--- a/floss/render/default.py
+++ b/floss/render/default.py
@@ -206,7 +206,8 @@ def render_decoded_strings(decoded_strings: List[DecodedString], ostream, verbos
 
 def render_heading(heading, n, ostream, disable_headers):
     """
-    example:
+    example::
+
         ===========================
         ‖ FLOSS TIGHT STRINGS (0) ‖
         ===========================
--- a/floss/render/default.py
+++ b/floss/render/default.py
@@ -134,7 +134,7 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
         unicode_offset_len = len(f"{unicode_strings[-1].offset}")
     offset_len = max(ascii_offset_len, unicode_offset_len)
 
-    render_heading("FLOSS STATIC STRINGS: ASCII", len(ascii_strings), ostream, disable_headers)
+    render_sub_heading("FLOSS STATIC STRINGS: ASCII", len(ascii_strings), ostream, disable_headers)
     for s in ascii_strings:
         if verbose == Verbosity.DEFAULT:
             ostream.writeln(s.string)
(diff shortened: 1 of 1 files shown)


williballenthin:
> I wonder though, could we also use another table format for the subheadings to make them stand out from the headings? I tried looking around in tabulate but didn't feel like any format would appear "lighter" than the ones currently in use for headings.

Good question, I see what you mean. Thanks for raising the issue and considering it a bit.

First, if we have a design that fits our needs that isn't in tabulate, that's ok. Tabulate just happened to be convenient for the original code.

I suppose we could use `=` instead of `-` on the top level headings, like:

```
===============================
| FLOSS STATIC STRINGS (3119) |
===============================
--------------------------------------
| FLOSS STATIC STRINGS: ASCII (3119) |
--------------------------------------
...
--------------------------------------
| FLOSS STATIC STRINGS: UTF-16LE (0) |
--------------------------------------
```

This seems slightly "denser" and maybe stands out a little bit more?

And/or we could use color to highlight things. Here are some examples:

### highlight the key part of the title
image


### highlight major title, mute minor title
image


### use shades of highlight to suggest priority
image

I'm not super keen on this last one because there aren't always a lot of colors available on older terminals. Also, without seeing the colors side by side (like above) it might be difficult to notice which is darker or lighter. Finally, we should consider how the colors look on both light and dark modes, which may invert the meaning of this heading style? Anyways, its still a design we could consider.



note: I think using color might be nice, but we should avoid using color to convey critical information since some users may have theme conflicts or accessibility issues. I don't think this feature is "critical" so the advice doesn't apply here, but I do want to call this out as a reminder since we're discussing colors.


What do you think @Shackles404?

Shackles404:
Thank you for the feedback and suggestions @williballenthin! I do like the idea of using different ASCII art more than changing colors as the terminal can be of any color. For example we could do:
```
==============================
‖ FLOSS STATIC STRINGS (862) ‖
==============================
+-----------------------------------+
| FLOSS STATIC STRINGS: ASCII (862) |
+-----------------------------------+
...
+------------------------------------+
| FLOSS STATIC STRINGS: UTF-16LE (0) |
+------------------------------------+
```
or maybe even 
```
==============================
‖ FLOSS STATIC STRINGS (862) ‖
==============================
+---------------------+
| ASCII STRINGS (862) |
+---------------------+
...
+----------------------+
| UTF-16LE STRINGS (0) |
+----------------------+
```
as the ASCII art might be enough to convey that the ASCII and UTF strings are static strings.

mr-tz:
Nice, I like the second suggestion. May be in conflict with #621 though... @williballenthin?!

williballenthin:
personally i prefer the first suggestion because it makes clear where the string is coming from (FLOSS static strings). i have this minor hestitation that we're not really differentiating the output strings from the program output, so there's this long shot scenario that a user gets confused whether the output is from FLOSS or the target .exe. using the term "FLOSS ..." makes it extra explicit.

then again, this is a pretty remote chance, so maybe its not a worthwhile consideration.

williballenthin:
oh, also, i prefer the first one because its clear that this is from "static strings" not from "stack strings" or "decoded strings" or whatever. and i personally thing this is pretty important.

Shackles404:
I can see why the first one is better. Plus it could also help when the ASCII list is long (which it frequently is) and user has to scroll down way below the main heading before the UTF subheading. It is probably better than the second one on the grounds that there is no possibility of any confusion of the type of string. I'll just add the first one then.

floss/render/default.py
```diff
@@ -205,12 +205,26 @@ def render_decoded_strings(decoded_strings: List[DecodedString], ostream, verbos
 
 
 def render_heading(heading, n, ostream, disable_headers):
+    """
+        example:
+        ===========================
+        ‖ FLOSS TIGHT STRINGS (0) ‖
+        ===========================
```
williballenthin, line 212: thanks for including an example here, great!

floss/render/default.py
```diff
@@ -205,12 +205,26 @@ def render_decoded_strings(decoded_strings: List[DecodedString], ostream, verbos
 
 
 def render_heading(heading, n, ostream, disable_headers):
+    """
+        example:
```
williballenthin, line 209:
```suggestion
    example:
```

williballenthin (approved):

williballenthin:
thank you @Shackles404!

PR #648, opened 2023-03-06
sigpaths parameter datatype

Shackles404:
The sigpaths parameter in the load_vw function expects a string, but is passed a list.
Changed:
```
def load_vw(
    sample_path: str,
    format: str,
    sigpaths: str,
    should_save_workspace: bool = False,
) -> VivWorkspace:
```
to:
```
def load_vw(
    sample_path: str,
    format: str,
    sigpaths: list,
    should_save_workspace: bool = False,
) -> VivWorkspace:
```
The paths are loaded by the get_signatures function, which returns a list.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -361,7 +361,7 @@ class Architecture(str, Enum):
 def load_vw(
     sample_path: str,
     format: str,
-    sigpaths: str,
+    sigpaths: list,
     should_save_workspace: bool = False,
 ) -> VivWorkspace:
     if format not in ("sc32", "sc64"):
--- a/floss/main.py
+++ b/floss/main.py
@@ -361,7 +361,7 @@ class Architecture(str, Enum):
 def load_vw(
     sample_path: str,
     format: str,
-    sigpaths: list,
+    sigpaths: List[str],
     should_save_workspace: bool = False,
 ) -> VivWorkspace:
     if format not in ("sc32", "sc64"):


Shackles404:
Also an unrelated question, is the is_workspace_file() function used for something currently?
```
def is_workspace_file(sample_file_path):
    """
    Return if input file is a vivisect workspace, based on file extension
    :param sample_file_path:
    :return: True if file extension is .viv, False otherwise
    """
    if os.path.splitext(sample_file_path)[1] == ".viv":
        return True
    return False
```
Should I clean it up to
```
def is_workspace_file(sample_file_path):
    """
    Return if input file is a vivisect workspace, based on file extension
    :param sample_file_path:
    :return: True if file extension is .viv, False otherwise
    """
    return os.path.splitext(sample_file_path)[1] == ".viv":
```

mr-tz:
Good catch. `is_workspace_file` can be removed as it seems unused.

floss/main.py
```diff
@@ -361,7 +361,7 @@ class Architecture(str, Enum):
 def load_vw(
     sample_path: str,
     format: str,
-    sigpaths: str,
+    sigpaths: list,
```
williballenthin, line 364:
```suggestion
    sigpaths: List[str],
```

Shackles404: Thank you!

williballenthin:
thanks!

PR #652, opened 2023-03-07
cleaning is_workspace_file function

Shackles404:
I am still not sure about if this function is useful, there seem to be no usages in the code... Is it useful when FLOSS is used as a module? I have cleaned it for now, but should I remove it completely?

The function:
```
def is_workspace_file(sample_file_path):
    """
    Return if input file is a vivisect workspace, based on file extension
    :param sample_file_path:
    :return: True if file extension is .viv, False otherwise
    """
    if os.path.splitext(sample_file_path)[1] == ".viv":
        return True
    return False
```
Original:
```
if os.path.splitext(sample_file_path)[1] == ".viv":
     return True
return False
```
Changed:
```
return os.path.splitext(sample_file_path)[1] == ".viv"
```

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -333,9 +333,7 @@ def is_workspace_file(sample_file_path):
     :param sample_file_path:
     :return: True if file extension is .viv, False otherwise
     """
-    if os.path.splitext(sample_file_path)[1] == ".viv":
-        return True
-    return False
+    return os.path.splitext(sample_file_path)[1] == ".viv"
 
 
 def is_supported_file_type(sample_file_path):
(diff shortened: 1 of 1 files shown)


williballenthin:
these changes are good, though i think that if the function isn't referenced within FLOSS, then we should just remove it completely. a more appropriate place for this functionality might be in [viv-utils](https://github.com/williballenthin/viv-utils) if someone needs it in a library. but, let's wait for someone to need it before adding code we need to maintain.

Shackles404:
Alright, I'll just remove it entirely. We could add it later if required (in FLOSS or any suitable library).

williballenthin (approved):

williballenthin:
thank you @Shackles404!

Issue #623, opened 2023-01-31
use color to differentiate results output


williballenthin:
we currently use a lot of ASCII art to break up the output in to sections and metadata. it can be a little overwhelming, especially when a user wants to find a string and work backwards. perhaps we can use a bit of color judiciuosly to highlight the interesting data and mute the rest:

<img width="452" alt="image" src="https://user-images.githubusercontent.com/156560/215755944-b3a37462-8fa0-48b1-b9e3-fa576cf0f0af.png">

Arker123:
Hi @williballenthin , can I work on this issue?

mr-tz:
Absolutely, I'll assign you. Thanks!

Arker123:
Hi @mr-tz  I'm wondering if you could provide some feedback on the colors. I've highlighted the strings. I've included the screenshots of the output, and I would appreciate any suggestions you might have.
<p float="left">
<img width="50%" src="https://user-images.githubusercontent.com/94680887/223024650-db6bc9fc-4692-4576-9533-44ecfe7c4318.png">
<img width="44%" src="https://user-images.githubusercontent.com/94680887/223024730-f41d96f5-bb4f-47d6-ad02-a7999af5939f.png">
</p>

mr-tz:
Cool, I'm all for colors, but would like to
* use them sparely, to focus just on the key parts, e.g., strings and maybe headings (see https://github.com/mandiant/flare-floss/pull/646#issuecomment-1455752036)
* have a solution that works on all configurations, especially light and dark backgrounds (maybe via a library?)

Arker123:
Thank you for the feedback and suggestion @mr-tz !

I've made the changes you suggested and the output now uses color sparingly to highlight only the key parts such as strings and headings. Additionally, I've tested the solution on both light and dark backgrounds to ensure that it works on all configurations.

I have some designs to select from:-
1)  Green Strings with Colored Heading and muted subheadings
<p float="left">
<img width="50%" src="https://user-images.githubusercontent.com/94680887/223335682-ef626ba5-0229-4e82-905a-97ee38b9f946.png">
<img width="49%" src="https://user-images.githubusercontent.com/94680887/223335698-9d1709f5-a356-49c1-bacd-168c692a30fc.png">
</p>


2) Blue Strings with Colored Heading and muted subheadings
<p float="left">
[...]

PR #653, opened 2023-03-09
Added colors to differentiate results output #623

Arker123:
This pull request adds color to the output of the program to make it easier to read and navigate. Specifically, it adds color to important elements such as headings and strings. The 'colorama' and 'termcolor' library has been used to achieve this.

Changes Made:

- Added the 'colorama' and 'termcolor' libraries as a dependency in setup.py
- Modified the code to use these libraries to add color to the output
- Added a command line argument --color to enable/disable color output
- Added support for the NO_COLOR environment variable as per the convention described in https://no-color.org/ via 'termcolor' library
- Works in both light and dark backgrounds

Testing:
  Tested the code on ubuntu 22.04 and windows 11

Screenshots:
<p float=left>

</p>
<p float=left>

</p>

Please review the code changes and provide feedback on the implementation. Let me know if there are any suggestions or concerns. Thank you!

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -266,10 +266,8 @@ def make_parser(argv):
 
 
 def set_color_config(color):
-    import codecs
-
-    codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp65001" else None)
-    codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp1252" else None)
+    import sys
+    sys.stdout.reconfigure(encoding='utf-8')
 
     if color == "always":
         colorama.init(strip=False)
--- a/floss/main.py
+++ b/floss/main.py
@@ -8,6 +8,7 @@ import logging
 import argparse
 import textwrap
 import contextlib
+import colorama
 from enum import Enum
 from time import time
 from typing import Set, List, Optional
@@ -253,11 +254,32 @@ def make_parser(argv):
     logging_group.add_argument(
         "-q", "--quiet", action="store_true", help="disable all status output on STDOUT except fatal errors"
     )
+    logging_group.add_argument(
+        "--color",
+        type=str,
+        choices=("auto", "always", "never"),
+        default="auto",
+        help="enable ANSI color codes in results, default: only during interactive session",
+    )
 
     return parser
 
 
-def set_log_config(debug, quiet):
+def set_log_config(debug, quiet, color):
+
+    if color == "always":
+        colorama.init(strip=False)
+    elif color == "auto":
+        # colorama will detect:
+        #  - when on Windows console, and fixup coloring, and
+        #  - when not an interactive session, and disable coloring
+        # renderers should use coloring and assume it will be stripped out if necessary.
+        colorama.init()
+    elif color == "never":
+        colorama.init(strip=True)
+    else:
+        raise RuntimeError("unexpected --color value: " + color)
+    
     if quiet:
         log_level = logging.WARNING
     elif debug >= DebugLevel.TRACE:
@@ -471,7 +493,7 @@ def main(argv=None) -> int:
         print(e)
         return -1
 
-    set_log_config(args.debug, args.quiet)
+    set_log_config(args.debug, args.quiet, args.color)
 
     # Since Python 3.8 cp65001 is an alias to utf_8, but not for Python < 3.8
     # TODO: remove this code when only supporting Python 3.8+
--- a/floss/main.py
+++ b/floss/main.py
@@ -269,6 +269,7 @@ def set_color_config(color):
     import codecs
 
     codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp65001" else None)
+    codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp1252" else None)
 
     if color == "always":
         colorama.init(strip=False)
--- a/floss/main.py
+++ b/floss/main.py
@@ -266,8 +266,7 @@ def make_parser(argv):
 
 
 def set_color_config(color):
-    import sys
-    sys.stdout.reconfigure(encoding='utf-8')
+    sys.stdout.reconfigure(encoding="utf-8")
 
     if color == "always":
         colorama.init(strip=False)
--- a/floss/main.py
+++ b/floss/main.py
@@ -266,7 +266,7 @@ def make_parser(argv):
 
 
 def set_color_config(color):
-    sys.stdout.reconfigure(encoding='utf-8')
+    sys.stdout.reconfigure(encoding="utf-8")
     colorama.just_fix_windows_console()
 
     if color == "always":
--- a/floss/main.py
+++ b/floss/main.py
@@ -266,6 +266,10 @@ def make_parser(argv):
 
 
 def set_color_config(color):
+    import codecs
+
+    codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp65001" else None)
+
     if color == "always":
         colorama.init(strip=False)
     elif color == "auto":
--- a/floss/main.py
+++ b/floss/main.py
@@ -10,7 +10,6 @@ import textwrap
 import contextlib
 from enum import Enum
 from time import time
-import sys
 from typing import Set, List, Optional
 
 import halo
--- a/floss/main.py
+++ b/floss/main.py
@@ -8,12 +8,12 @@ import logging
 import argparse
 import textwrap
 import contextlib
-import colorama
 from enum import Enum
 from time import time
 from typing import Set, List, Optional
 
 import halo
+import colorama
 import viv_utils
 import viv_utils.flirt
 from vivisect import VivWorkspace
@@ -265,8 +265,7 @@ def make_parser(argv):
     return parser
 
 
-def set_log_config(debug, quiet, color):
-
+def set_color_config(color):
     if color == "always":
         colorama.init(strip=False)
     elif color == "auto":
@@ -279,7 +278,9 @@ def set_log_config(debug, quiet, color):
         colorama.init(strip=True)
     else:
         raise RuntimeError("unexpected --color value: " + color)
-    
+
+
+def set_log_config(debug, quiet):
     if quiet:
         log_level = logging.WARNING
     elif debug >= DebugLevel.TRACE:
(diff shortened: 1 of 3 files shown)


williballenthin:
Would you please fix the code style violations identified in CI? they can be reproduced locally with:

```console
isort --profile black --length-sort --line-width 120 -c .
black -l 120 --check .
mypy --config-file .github/mypy/mypy.ini floss/ scripts/ tests/
```

floss/main.py
```diff
@@ -253,11 +254,32 @@ def make_parser(argv):
...
+    elif color == "auto":
+        # colorama will detect:
+        #  - when on Windows console, and fixup coloring, and
+        #  - when not an interactive session, and disable coloring
+        # renderers should use coloring and assume it will be stripped out if necessary.
+        colorama.init()
+    elif color == "never":
+        colorama.init(strip=True)
+    else:
+        raise RuntimeError("unexpected --color value: " + color)
```
williballenthin, line 281: would you move this into main, under the call to `set_log_config`? the coloring configuration isn't really part of logging (though it *does* have to do with program output, so there's some similarity) so i don't think it fits here.

alternatively, you could introduce a new function like `set_color_config` to hold this logic.

floss/render/default.py
```diff
@@ -134,20 +135,22 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
         unicode_offset_len = len(f"{unicode_strings[-1].offset}")
     offset_len = max(ascii_offset_len, unicode_offset_len)
 
-    render_sub_heading("FLOSS STATIC STRINGS: ASCII", len(ascii_strings), ostream, disable_headers)
+    render_sub_heading("FLOSS STATIC STRINGS: " + colored("ASCII", "blue", attrs=['bold']), len(ascii_strings), ostream, disable_headers)
```
williballenthin, line 138: the implementation here is good, and i think it can be even further polished by refactoring the common coloring configs into their own functions, like `def heading_style(str) -> str` and `def string_style(str) -> str`. the benefits are twofold:

  1. a bunch of lines are simplified, and
  2. if the config changes, like changing the color, its easy to do in one place, versus many.

if you agree, would you please update the PR with these changes?

floss/render/default.py
```diff
@@ -134,20 +135,22 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
...
     for s in ascii_strings:
+        str_ = colored(s.string, "green", attrs=['bold'])
         if verbose == Verbosity.DEFAULT:
-            ostream.writeln(s.string)
+            ostream.writeln(str_)
         else:
-            ostream.writeln(f"0x{s.offset:>0{offset_len}x} {s.string}")
+            ostream.writeln(f"0x{s.offset:>0{offset_len}x} {str_}")
```
williballenthin, line 144: nice

williballenthin (changes requested):

williballenthin (approved):
looks awesome, thank you @Arker123!

williballenthin:
hm CI failure: https://github.com/mandiant/flare-floss/actions/runs/4376107350/jobs/7657946700

mr-tz:
Hm, I think I've seen this before somewhere, but don't fully recall...

```
emulating function 0x401500 (call 1/1): 100%|##########| 20/20 [00:03<00:00,  6.40 functions/s]
INFO: floss: finished execution after 12.82 seconds
Traceback (most recent call last):
  File "main.py", line 700, in <module>
  File "main.py", line 694, in main
  File "main.py", line 476, in write
  File "colorama\ansitowin32.py", line 41, in write
  File "colorama\ansitowin32.py", line 162, in write
  File "colorama\ansitowin32.py", line 190, in write_and_convert
  File "colorama\ansitowin32.py", line 195, in write_plain_text
  File "colorama\ansitowin32.py", line 41, in write
  File "colorama\ansitowin32.py", line 162, in write
  File "colorama\ansitowin32.py", line 190, in write_and_convert
  File "colorama\ansitowin32.py", line 195, in write_plain_text
  File "encodings\cp1252.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode character '\u2016' in position 980: character maps to <undefined>
```

That character is "DOUBLE VERTICAL LINE", maybe from tabulate and conflicts with colorama?

In capa we have some code around codecs, but not sure this is it here, see https://github.com/mandiant/capa/blob/3dfb7beb6bcb73959c714a24b01566b969efc4f9/capa/main.py#L921

mr-tz:
Do you have a Window system to test?

setup.py
```diff
@@ -14,6 +14,9 @@
     "tqdm==4.65.0",
     "networkx==2.5.1",
     "halo==0.0.31",
+    "termcolor==2.2.0",
+    "colorama==0.4.4",
+    "types-colorama==0.4.15.8",
```
mr-tz, line 19: this should go into the dev section below

mr-tz (commented):
one little thing on the requirements and we'll have to see how to make this work on Windows

mr-tz:
The screenshots look great! Good work.

mr-tz:
Let's try `codecs.register(lambda name: codecs.lookup("utf-8") if name == "cp65001" else None)` (per the capa reference) for a potential quick win?

williballenthin:
in your tests above, which version of python are you using? looks like we're using python 3.8 in CI.

i notice in the stack trace that the encoding that's referenced is `cp1252` rather than `cp65001`. i guess we could try moritz's fix with that encoding, though im worried that we don't really understand what's wrong quite yet. but, a fix is a fix.

> `r.encode("utf-8").decode('ascii', 'ignore')`

i see how this can work, but i don't like it a lot because:

  1. the characters it can't display it just drops, which will look bad with the ascii art, and
  2. it hides the underlying issue

for right now, i'd like to keep looking for an explanation and fix. let me poke around a bit more and see if i can figure anything out. please feel free to do the same, but also, don't waste a lot of time and effort on this, unless you enjoy it and/or you're learning.

williballenthin:
(message received. i'll be able to respond hopefully within about two hours, otherwise in the morning. its at the top of my list!)

williballenthin:
first things first, i don't think i understand this very well. so, im hesitant about even trying to explain, but perhaps by trying i will learn myself, so here goes...

when you do `print("foo")` in python this ultimately calls some OS API to move the text of "foo" to the console (or whatever the standard output points to, might also be a file, a pipe, etc.). there has to be an agreement between python and the console on how "foo" is represented. this is the encoding.

on *nix systems, the encoding is basically always utf-8, so there's not much to decide.

on windows, things are complicated by history and compatibility and stuff. internally today, windows represents strings using wide characters (utf-16le), but that's not necessarily how data is passed from program to program (in this case, python to console). there's also some things in play around code pages. my specific knowledge here is fuzzy. but basically it has to do with legacy ways of encoding data, particularly before unicode was widespread.

the takeaway is that on windows, python has to try to guess what the agreement is for encoding output text.

the stack trace says:

```
  File "encodings\cp1252.py", line 19, in encode
UnicodeEncodeError: 'charmap' codec can't encode character '\u2016' in position 980: character maps to <undefined>
```

which hints that python is trying to use code page 1252 to encode the text when passing to the console. however, 1252 is a really small code page and doesn't have a way to represent the Unicode character "DOUBLE VERTICAL LINE". so when python tries to print the ASCII art table headings, it doesn't know how to represent this to the console and raises an exception.

by using `sys.stdout.reconfigure(encoding='utf-8')`, we are forcing python to use "utf-8" to encode text to stdout. utf-8 has a way to represent "DOUBLE VERTICAL LINE" so it doesn't raise an exception.

now, im not sure if this is correct, because im not sure if the console actually expects to receive utf-8. if it does want utf-8, this is a correct solution. but if not, the console may display garbage (broken font symbols, for example). i can't figure out an authoritative reference for this, especially without getting bogged down in decades of legacy cruft.

notably, the python documentation for [sys.stdout](https://docs.python.org/3/library/sys.html#sys.stderr) describe how python tries to guess the output encoding. it indicates that utf-8 is correct for console output, but if the output is redirected elsewhere, the current codepage setting is used - just like we're seeing here. so, my guess is that in Github Actions CI, they're redirecting terminal output, so python picks the current code page, which is cp1252 and is insufficient for our ASCII art, and things break.

i think this is probably a mostly safe solution, but there's the potential for trouble when FLOSS outputs utf-8 and either 1) the console doesn't expect utf-8 (but im not sure when this is), or 2) output is redirected to a file and when the file is subsequently consumed, the downstream application isn't expecting utf-8 (maybe ascii or cp1252). 

does that make any sense? can you see anything i've got wrong and/or suggest an alternative conclusion?

Shackles404:
It was me who added the double vertical line character... 
`heading = f"‖ {heading} ({n}) ‖"` 
How did it manage to break stuff though?

williballenthin:
yeah, i agree that its odd that CI worked fine when introducing the unicode characters and that things now fail in this separate PR. i wonder if colorama is doing any manipulation of the output encoding. i'll look into this.

williballenthin:
image

it does seem like colorama is doing some low level encoding stuff that might easily collide with the other recent changes.

setup.py
```diff
@@ -66,6 +66,9 @@
             # type stubs for mypy
             "types-PyYAML==6.0.10",
             "types-tabulate==0.9.0.1",
+            "termcolor==2.2.0",
+            "colorama==0.4.4",
```
williballenthin, line 70: these should remain in `requirements` while the `types-*` dependencies should stay here. type checking is only done during development/CI, that's why they're separated out. dependencies that are needed for FLOSS to run should be listed in `requirements`

williballenthin (changes requested):

floss/main.py
```diff
@@ -13,6 +13,7 @@
 from typing import Set, List, Optional
```
williballenthin, line 13:
```suggestion
import sys
from typing import Set, List, Optional
```

floss/main.py
```diff
@@ -253,10 +254,34 @@ def make_parser(argv):
...
+        type=str,
+        choices=("auto", "always", "never"),
+        default="auto",
+        help="enable ANSI color codes in results, default: only during interactive session",
+    )
 
     return parser
 
 
+def set_color_config(color):
```
williballenthin, line 268:
```suggestion
def set_color_config(color):
    sys.stdout.reconfigure(encoding='utf-8')
```

looks like we still need this to force the output encoding

floss/main.py
```diff
@@ -10,9 +10,11 @@
 import contextlib
 from enum import Enum
 from time import time
+import sys
```
williballenthin, line 13:
```suggestion
```

williballenthin:
image
woohoo

good to merge everyone?

mr-tz (approved):
great, thank you!

PR #664, opened 2023-03-21
language identification: initial GO binary detection

d01a:
An initial implementation of Language identification module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setuptools.setup(
     extras_require={
         "dev": [
             "pyyaml==6.0",
-            "pytest==7.3.0",
+            "pytest==7.3.1",
             "pytest-sugar==0.9.4",
             "pytest-instafail==0.5.0",
             "pytest-cov==4.0.0",


floss/main.py
```diff
@@ -283,6 +284,54 @@ def set_color_config(color):
...
+        b"\xfb\xff\xff\xff\x00\x00",
+        b"\xfa\xff\xff\xff\x00\x00",
+        b"\xf1\xff\xff\xff\x00\x00",
+    ]
+
+    for magic in go_magic:
+        for section in pe.sections:
+            section_va = section.VirtualAddress
+            section_size = section.SizeOfRawData
+            # TODO: we can customize the serch to start with .rdata section.
```
mr-tz, line 312: right, we can probably optimize and skip various sections here

mr-tz (commented):
neat, thanks, this is a good start towards golang binary detection

floss/main.py
d01a: I added search for the .data section first. if not available go through all sections

PR #665, opened 2023-03-22
Adds tightstrings to render/scripts

### LINKED ISSUE #559: Add tightstrings to scripts/render-* files
IDA handled via #580

ooprathamm:
Fixes #559

### DIFF
--- a/scripts/render-binja-import-script.py
+++ b/scripts/render-binja-import-script.py
@@ -50,22 +50,18 @@ def render_binja_script(result_document: ResultDocument) -> str:
                 main_commands.append('AppendComment(%d, "FLOSS: " + %s)' % (ds.decoded_at, b64))
     main_commands.append('print("Imported decoded strings from FLOSS")')
 
-    ss_len = 0
     for ss in result_document.strings.stack_strings:
         if ss.string != "":
             b64 = base64.b64encode(ss.string.encode("utf-8")).decode("ascii")
             b64 = 'base64.b64decode("%s").decode("utf-8")' % (b64)
             main_commands.append('AppendLvarComment(%d, "FLOSS stackstring: " + %s)' % (ss.function, b64))
-            ss_len += 1
     main_commands.append('print("Imported stackstrings from FLOSS")')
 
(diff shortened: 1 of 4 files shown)


scripts/render-r2-import-script.py
```diff
@@ -60,6 +60,12 @@ def render_r2_script(result_document: ResultDocument) -> str:
             sanitized_string = base64.b64encode(b'"FLOSS: %s"' % ss.string.encode("utf-8")).decode("ascii")
             main_commands.append("Ca -0x%x base64:%s @ %d" % (ss.frame_offset, sanitized_string, ss.function))
             ss_len += 1
+    ts_len = 0
```
mr-tz, line 63: here the lengths aren't used

ooprathamm: here we are also counting the number of stack strings (ss_len), so I thought this might be something added for future scope.
Should I remove both?

scripts/render-binja-import-script.py
```diff
@@ -109,7 +118,7 @@ def AppendLvarComment(fva, s):
 %s
 
 """ % (
-        len(result_document.strings.decoded_strings) + ss_len,
+        len(result_document.strings.decoded_strings) + ss_len + ts_len,
```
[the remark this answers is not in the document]
ooprathamm: No, empty strings are not added but looking at the code below I feel appending of statements in main_commands need restructuring due to sequence of statements in last part of binary ninja script.

* Results of floss test-decode-tightstring.exe
image
*Last part of bninja script 
image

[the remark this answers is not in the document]
mr-tz: Yes, we can remove the manual counts and just use `len(result_document.strings....)`

scripts/render-r2-import-script.py
mr-tz: yes, please, if needed we could use `len(result_document.strings....)`

mr-tz:
oh, can you please fix the black lint issues?

PR #667, opened 2023-03-24
Replace termcolor with rich #623

Arker123:
Changes:-
- Replacing termcolor with rich
- Removed --outfile cli parameter
- Moved write function inline

Screenshots :-

1) Command --> `floss -v [executable name]`
<p float="left">

</p>

2) Command --> `floss -v [executable name] --color never]`
<p float="left">

</p>

3) Command --> `floss -v [executable name] > file`

<p float="left">

</p>

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -448,7 +448,6 @@ def get_signatures(sigs_path):
 
 
 def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, outfile: Optional[str], color):
-    sys.stdout.reconfigure(encoding="utf-8")
     if json_:
         r = floss.render.json.render(results)
     else:
--- a/floss/main.py
+++ b/floss/main.py
@@ -448,6 +448,7 @@ def get_signatures(sigs_path):
 
 
 def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, outfile: Optional[str], color):
+    sys.stdout.reconfigure(encoding="utf-8")
     if json_:
         r = floss.render.json.render(results)
     else:
--- a/floss/main.py
+++ b/floss/main.py
@@ -235,12 +235,6 @@ def make_parser(argv):
         default=Verbosity.DEFAULT,
         help="enable verbose results, e.g. including function offsets (does not affect JSON output)",
     )
-    output_group.add_argument(
-        "-o",
-        "--outfile",
-        type=str,
-        help="write results to output file, instead of default STDOUT",
-    )
 
     logging_group = parser.add_argument_group("logging arguments")
     logging_group.add_argument(
@@ -447,17 +441,13 @@ def get_signatures(sigs_path):
     return paths
 
 
-def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, outfile: Optional[str], color):
+def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, color):
     if json_:
         r = floss.render.json.render(results)
     else:
         r = floss.render.default.render(results, verbose, quiet, color)
-    if outfile:
-        logger.info("writing results to %s", outfile)
-        with open(outfile, "wb") as f:
-            f.write(r.encode("utf-8"))
-    else:
-        print(r)
+
+    print(r)
 
 
 def main(argv=None) -> int:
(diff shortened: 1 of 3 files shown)


floss/main.py
```diff
@@ -270,18 +271,16 @@ def set_color_config(color):
     colorama.just_fix_windows_console()
 
     if color == "always":
-        colorama.init(strip=False)
```
williballenthin, line 273: lets remove the colorama import if its no longer used here. also the dependency in setup.py

floss/main.py
```diff
@@ -466,7 +465,7 @@ def get_signatures(sigs_path):
     return paths
 
 
-def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, outfile: Optional[str]):
+def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, outfile: Optional[str], console):
```
williballenthin, line 468: i wonder if we should remove the `--outfile` cli parameter and rely on people redirecting output like `floss > output.json`. there's getting to be a lot of arguments to this function.

floss/render/default.py
```diff
@@ -158,67 +164,66 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
...
-            render_sub_heading(" FUNCTION at " + heading_style(f"0x{fva:x}"), len(data), ostream, disable_headers)
+            render_sub_heading(" FUNCTION at " + heading_style(f"0x{fva:x}"), len(data), console, disable_headers)
             rows = []
             for ds in data:
                 if ds.address_type == AddressType.STACK:
-                    offset_string = "[stack]"
+                    offset_string = "Stack"
                 elif ds.address_type == AddressType.HEAP:
-                    offset_string = "[heap]"
+                    offset_string = "Heap"
```
williballenthin, line 211: can you explain this change?

floss/main.py
```diff
@@ -466,17 +441,13 @@ def get_signatures(sigs_path):
     return paths
 
 
-def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, outfile: Optional[str]):
+def write(results: ResultDocument, json_: bool, verbose: Verbosity, quiet: bool, color):
```
wballenthin, line 444: please provide a type hint for `color`

floss/render/default.py
```diff
@@ -158,67 +155,73 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
...
             strings_by_functions[ds.decoding_routine].append(ds)
 
         for fva, data in strings_by_functions.items():
-            render_sub_heading(" FUNCTION at " + heading_style(f"0x{fva:x}"), len(data), ostream, disable_headers)
+            render_sub_heading(" FUNCTION at " + heading_style(f"0x{fva:x}"), len(data), console, disable_headers)
             rows = []
             for ds in data:
                 if ds.address_type == AddressType.STACK:
-                    offset_string = "[stack]"
+                    offset_string = "\[Stack]"
```
wballenthin, line 207:
```suggestion
                    offset_string = "\[stack]"
```

floss/render/default.py
```diff
@@ -158,67 +155,73 @@ def render_staticstrings(strings, ostream, verbose, disable_headers):
...
                 elif ds.address_type == AddressType.HEAP:
-                    offset_string = "[heap]"
+                    offset_string = "\[Heap]"
```
wballenthin, line 209:
```suggestion
                    offset_string = "\[heap]"
```

setup.py
```diff
@@ -68,7 +67,6 @@
             # type stubs for mypy
             "types-PyYAML==6.0.10",
             "types-tabulate==0.9.0.1",
-            "types-colorama==0.4.15.8",
```
wballenthin, line 71: are there type hints for `rich` that we can include? or not necessary?

wballenthin (changes requested):
minor changes requested and then looks good to me!

floss/main.py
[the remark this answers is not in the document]
wballenthin: i like that. the logic of this function is trivial and yet it accepts so many parameters. moving it inline seems like a good simplification.

floss/main.py
wballenthin: we should ensure that when `--color` is not explicitly provided and the output is redirect that color is automatically disabled. i think `rich` does this, but please confirm.

mr-tz:
Also looks good to me overall! Thanks!
Can you share a screenshot of the new output here?

mr-tz (approved):
lgtm

Issue #658, opened 2023-03-14
fix mypy 1.1.1 fails


mr-tz:
```
Run mypy --config-file .github/mypy/mypy.ini floss/ scripts/ tests/
floss/strings.py:55: error: Too many positional arguments for "StaticString"  [misc]
floss/strings.py:82: error: Too many positional arguments for "StaticString"  [misc]
floss/utils.py:234: error: Too many positional arguments for "StaticString"  [misc]
floss/stackstrings.py:53: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
floss/stackstrings.py:181: error: Too many positional arguments for "StackString"  [misc]
floss/tightstrings.py:112: error: Too many positional arguments for "TightString"  [misc]
floss/string_decoder.py:165: error: Too many positional arguments for "DecodedString"  [misc]
Found 6 errors in 5 files (checked 36 source files)
```

ref: #656

samrath-sudesh-acharya:
@mr-tz can you assign me this issue

EmperialX:
@williballenthin  Thank you for letting me know, I understand @Arker123 is currently working on this issue and I respect that. Please let me know if there are any other issues or tasks that I can work on to contribute to the project Floss. I look forward to future opportunities. Thank you.

mr-tz:
Sure, please go ahead.

EmperialX:
Thank you , i will let you know my progress soon as possible.

williballenthin:
@EmperialX as you work on this, would you please also add the `--check-untyped-defs` option to the mypy config here: https://github.com/mandiant/flare-floss/blob/8eed8c653d765192ee66cd4308beaeedd362ba42/.github/workflows/tests.yml#L31

this may add some more items to fix, but is a much more thorough way of running mypy. thank you!

berkaygunduzz:
I think I fixed the error in the PR [#670](https://github.com/mandiant/flare-floss/pull/670)

williballenthin:
@Aayush-Goel-04 go ahead and take a stab, either continuing the existing work or starting from scratch. Open a PR when you have it in good working shape and we'll take a look.

PR #670, opened 2023-04-02
Fix mypy 1.1.1 integration error of PR #656

berkaygunduzz:
Specified argument's keywords when using string classes defined in 'floss/results'

PR #671, opened 2023-04-03
Fix mypy 1.1.1 failures for issue #658 with add --check-untyped-defs 

EmperialX:
fixed mypy 1.1.1 with changes in workflow and added the --check-untyped-defs option to the mypy config

closes issue https://github.com/mandiant/flare-floss/issues/658


mr-tz:
Note that this is already being worked on in #670.

EmperialX:
ok

PR #681, opened 2023-04-07
Remove unused class Architecture

symbolicvoid:
I don't think the class `Architecture` is used. Also, we could use Vivisect to get the architecture of the file if required.


mr-tz (approved):
Good catch, thanks!

PR #683, opened 2023-04-10
Added rich as default traceback handler #623

Arker123:
This pull request adds "rich" as the default traceback handler. This will enhance the visual presentation of tracebacks and make it easier for developers to read and understand them. Rich tracebacks are easier to read and show more code than standard Python tracebacks.

Here are some Screenshots:-
<p float="left">

</p>

To demonstrate the tracebacks, I intentionally made a small mistake by changing the function name from 'make_parser()' to 'make_parserr()', which caused a NameError because 'make_parserr()' was not defined.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -14,9 +14,9 @@ from typing import Set, List, Optional
 
 import halo
 import viv_utils
+import rich.traceback
 import viv_utils.flirt
 from vivisect import VivWorkspace
-from rich.traceback import install
 
 import floss.utils
 import floss.results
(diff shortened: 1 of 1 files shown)


floss/main.py
```diff
@@ -56,6 +57,9 @@
 
 logger = floss.logging_.getLogger("floss")
 
+# use rich as default Traceback handler
+install(show_locals=True)
+
```
williballenthin, line 62: please invoke this from within `main`. otherwise, this configuration is set whenever `floss.main` is imported, which is a problem if when floss is used as a library, since it should be up to the application to decide how to handle output/logging.

williballenthin: also, lets use the fully qualified path `rich.traceback.install` so thats its more clear where the name comes from.

williballenthin (changes requested):
i love the new functionality - it looks great. thanks for taking the time to open this PR.

i have a couple comments that i hope are easy to address. then, happy to merge.

williballenthin (approved):

williballenthin:
thank you!

PR #695, opened 2023-04-29
Language Identification module

d01a:


floss/main.py
```diff
@@ -259,6 +260,67 @@ def make_parser(argv):
     return parser
 
 
+def is_go_bin(sample: str) -> bool:
```
mr-tz, line 263: this and related code could ideally go into a separate file

mr-tz (commented):
nice, this is a good start! next I'd like to see tests for this (and a reorg into a separate file out of main)

d01a:
Referring to the discussion, should it exit after saying that GO strings not handled or just a warning and continue?  I will try to continue and add string extractor but just to close this for now.

gmr-tz:
A warning and continuing sound good. We can still try to do more work but this sets better expectations with the user.

d01a:
What else i can do?

floss/language_identifier.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (C) 2023 Mandiant, Inc. All Rights Reserved.
+import pefile
```
williballenthin, line 2: i think this needs to be added to setup.py

floss/language_identifier.py
```diff
@@ -0,0 +1,68 @@
...
+    go_magic = [
+        b"\xf0\xff\xff\xff\x00\x00",
+        b"\xfb\xff\xff\xff\x00\x00",
+        b"\xfa\xff\xff\xff\x00\x00",
+        b"\xf1\xff\xff\xff\x00\x00",
+    ]
+
+    # look for the .rdata section first
+    for section in pe.sections:
+        if ".rdata" in section.Name.decode("utf-8").rstrip("\x00"):
```
williballenthin, line 33: please wrap this in a try/except because this decoding might easily fail on corrupt PE files
```suggestion
        try:
            section_name = section.Name.partition(b"\x00")[0].decode("utf-8")
        except UnicodeDecodeError:
            continue
        if ".rdata" in section_name:
```

floss/language_identifier.py
```diff
@@ -0,0 +1,68 @@
...
+            section_va = section.VirtualAddress
+            section_size = section.SizeOfRawData
+            for magic in go_magic:
+                if magic in section.get_data(section_va, section_size):
+                    pclntab_va = section.get_data(section_va, section_size).index(magic) + section_va
+                    if verify_pclntab(section, pclntab_va):
+                        return True
```
williballenthin, line 40: it might be expensive to fetch the section data (in practice maybe not that bad, but still), so lets pull the computation out of the inner loop

```suggestion
            section_data = section.get_data(section_va, section_size)
            for magic in go_magic:
                if magic in section_data:
                    pclntab_va = section_data.index(magic) + section_va
                    if verify_pclntab(section, pclntab_va):
                        return True
```

floss/language_identifier.py
```diff
@@ -0,0 +1,68 @@
...
+
+    for magic in go_magic:
+        for section in pe.sections:
+            section_va = section.VirtualAddress
+            section_size = section.SizeOfRawData
+            if magic in section.get_data(section_va, section_size):
+                pclntab_va = section.get_data(section_va, section_size).index(magic) + section_va
+                if verify_pclntab(section, pclntab_va):
+                    # just for testing
+                    return True
```
williballenthin, line 53:
```suggestion
            section_data = section.get_data(section_va, section_size)
            if magic in section_data:
                pclntab_va = section_data.index(magic) + section_va
                if verify_pclntab(section, pclntab_va):
                    # just for testing
                    return True
```

floss/language_identifier.py
```diff
@@ -0,0 +1,68 @@
...
+    Parse headers of pclntab to verify it is legit
+    used in go parser itself https://go.dev/src/debug/gosym/pclntab.go
+    """
+    pc_quanum = section.get_data(pclntab_va + 6, 1)
+    pointer_size = section.get_data(pclntab_va + 7, 1)
+    if (pc_quanum != b"\x01" and pc_quanum != b"\x02" and pc_quanum != b"\x04") or (
+        pointer_size != b"\x04" and pointer_size != b"\x08"
+    ):
+        return False
+    return True
```
williballenthin, line 68:
```suggestion
    pc_quanum = section.get_data(pclntab_va + 6, 1)[0]
    pointer_size = section.get_data(pclntab_va + 7, 1)[0]
    return pc_quanum in {1, 2, 4} and pointer_size in {4, 8}:
        return False
```

floss/main.py
```diff
@@ -528,6 +529,10 @@ def main(argv=None) -> int:
     # 3. tight strings
     # 4. decoded strings
 
+    if is_go_bin(sample=sample):
+        logger.warning("GO Binary Detected, GOLang binaries are not supported yet. Results may be inaccurate.")
+        logger.warning("GO: Proceeding with analysis may take a long time.")
```
williballenthin, line 534:
```suggestion
        logger.warning("Go Binary Detected, Go binaries are not supported yet. Results may be inaccurate.")
        logger.warning("Go: Proceeding with analysis may take a long time.")
```

The language name is "Go", which can be difficult to Google for, since its such as short name, hence the convention that people sometimes use "golang". but we should use "Go" here because its the correct name.

mr-tz:
Great, I think we'd now like to see tests showing this works as expected and can merge then.

d01a:
hey, i just added some binaries to `flare-floss-testfiles` repo. idk if that what you meant by testing but if you want me to try it on some go sample, i got some go malwares i can test and attach screenshots. If something else, please explain it.

d01a:
I added the test file. There is nothing more to do in the PR in the other repo, i think.

tests/test_language_id.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from floss.language_identifier import is_go_bin
+
+
+@pytest.mark.parametrize("binary_file, expected_result", [
+    ("./tests/data/src/go-hello/bin/go-hello.exe", True),
+    ("./tests/data/src/go-hello/bin/go-hello", False),
```
mr-tz: should this be an expected fail for now, but actually `True`?

tests/data
mr-tz, line 1: perfect, thanks!

floss/language_identifier.py
```diff
@@ -0,0 +1,70 @@
...
+    it checks the magic header of the pclntab structure -pcHeader-
+    the magic values varies through the version
+    reference:
+    https://github.com/0xjiayu/go_parser/blob/865359c297257e00165beb1683ef6a679edc2c7f/pclntbl.py#L46
+    """
+    try:
+        pe = pefile.PE(sample)
+    except:
+        logger.debug("NOT valid PE header")
+        return False
```
mr-tz, line 22: see the tests issue, we should at least catch FileNotFound here specifically and maybe also pefile errors.

tests/test_language_id.py
```diff
@@ -0,0 +1,19 @@
...
+
+from floss.language_identifier import is_go_bin
+
+
+@pytest.mark.parametrize(
+    "binary_file, expected_result",
+    [
+        ("./tests/data/src/go-hello/bin/go-hello.exe", True),
+        ("./tests/data/src/go-hello/bin/go-hello", False),
+        ("test-decode-to-stack.exe", False),
```
mr-tz, line 11: shouldn't this be also in the same directory as the other binaries? do we need to assert, paths exist?

tests/test_language_id.py
d01a: Yes, it should be True. but elf not handled. maybe i will add it.

tests/test_language_id.py
d01a: I added the file check. in test script and in try/except

tests/test_language_id.py
mr-tz: let's comment it out with a note on that for now

mr-tz: let's comment it out with a note on that for now

floss/language_identifier.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Check if the binary given is compiled with Go compiler or not
+    it checks the magic header of the pclntab structure -pcHeader-
+    the magic values varies through the version
+    reference:
+    https://github.com/0xjiayu/go_parser/blob/865359c297257e00165beb1683ef6a679edc2c7f/pclntbl.py#L46
+    """
+    try:
+        pe = pefile.PE(sample)
+    except:
```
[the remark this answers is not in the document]
d01a: After some googling, i tried to handle it. If you see a specific case, tell me. 
The `Exception` case is raised as it is a generic exception, idk if this is the right thing but that's what i think of it.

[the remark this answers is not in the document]
d01a: @mr-tz can you take a look

mr-tz (approved):
lgtm

floss/language_identifier.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def verify_pclntab(section, pclntab_va: int) -> bool:
+    """
+    Parse headers of pclntab to verify it is legit
+    used in go parser itself https://go.dev/src/debug/gosym/pclntab.go
+    """
+    pc_quanum = section.get_data(pclntab_va + 6, 1)[0]
+    pointer_size = section.get_data(pclntab_va + 7, 1)[0]
+    return True if pc_quanum in {1, 2, 4} and pointer_size in {4, 8} else False
```
williballenthin, line 78: wrap this in a try/except to handle malformed pe files where this data is not accessible. in that case, return False

williballenthin (changes requested):
some minor change requests but i really like what i see here. thanks for the hard work and i'm excited to get this merged shortly. seems like we should be able to get this done this week!

@Arker123 is also going to take a look shortly and will provide feedback.

floss/language_identifier.py
williballenthin: not strictly required, but seems like a good idea

floss/language_identifier.py
```diff
@@ -0,0 +1,78 @@
...
+        b"\xf1\xff\xff\xff\x00\x00",
+    ]
+
+    # look for the .rdata section first
+    for section in pe.sections:
+        try:
+            section_name = section.Name.partition(b"\x00")[0].decode("utf-8")
+        except UnicodeDecodeError:
+            continue
+        if ".rdata" in section_name:
```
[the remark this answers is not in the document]
d01a: I think you are right but two levels back? as if it is one level it will execute only one iteration and if the section not `.data` will return false. If 2 levels, it will return false if all the section checked. 
why `in` instead of `==`, I read a blog for a long time ago and he mentioned that the section name could be not strictly equal what we expect, it might have a garbage byte that will make the equality not true. It is better to use it anyway as the comparison will be small so will not affect the performance.

floss/language_identifier.py
d01a: I will do it. thank you.
Seems like I should think more about exceptions lol.

floss/language_identifier.py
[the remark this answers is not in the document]
d01a: But what about the loop? I guess it will return after only one iteration if it unindented one level.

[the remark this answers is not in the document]
williballenthin: in which case, we only expect there to be a single section named .rdata, so we only expect the loop to even run once, so we can have the return in the inner branch, and in the outer scope just let things fall through. so maybe just remove the second return?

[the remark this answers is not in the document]
williballenthin:
```suggestion
        if ".rdata" == section_name:
```

floss/language_identifier.py
```diff
@@ -0,0 +1,78 @@
...
+            section_va = section.VirtualAddress
+            section_size = section.SizeOfRawData
+            section_data = section.get_data(section_va, section_size)
+            for magic in go_magic:
+                if magic in section_data:
+                    pclntab_va = section_data.index(magic) + section_va
+                    if verify_pclntab(section, pclntab_va):
+                        return True
+            return False
```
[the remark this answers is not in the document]
williballenthin:
```suggestion
```

floss/language_identifier.py
```diff
@@ -0,0 +1,77 @@
...
+            section_va = section.VirtualAddress
+            section_size = section.SizeOfRawData
+            section_data = section.get_data(section_va, section_size)
+            if magic in section_data:
+                pclntab_va = section_data.index(magic) + section_va
+                if verify_pclntab(section, pclntab_va):
+                    # just for testing
+                    return True
+    return False
+
```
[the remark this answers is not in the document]
d01a: It's a good feature. I thought about it when I saw it used in DiE signatures. But I didn't believe that it would make any difference, the core thing is extracting the strings which did not change over the versions. 
But as you have implemented it, it will be a nice addition.

williballenthin (approved):
woohoo! thank you @d01a

mr-tz:
Thanks, everyone! Great to see this merged and kick start language specific string extractions.

PR #744, opened 2023-05-19
Allow running floss using `python -m floss`

Dobatymo:
This doesn't need any path setup like the floss entrypoint does.
see https://github.com/psf/black for an example of how this is done in other popular projects.


williballenthin (approved):
good idea, thank you!

Issue #762, opened 2023-05-25
Implement Language Identification Module for Rust


Arker123:
Create an efficient language identification module in Rust that enables accurate language detection and classification.

PR #775, opened 2023-06-01
Language Identification module for Rust #762

Arker123:
This PR introduces a Rust module for language identification. It enables easy classification of programming languages used in Rust projects, enhancing code analysis and understanding.

Changes made:
- Added Rust identification module
- Added Database file of Rust hashes that maps them to the associated rust version
- Added extract_rust_hashes.py file to generate the Database file that maps Rust hashes to their versions

Screenshots:
image

Please review the code changes and provide feedback on the implementation. Let me know if there are any suggestions or concerns. Thank you!


floss/language_identifier.py
```diff
@@ -1,13 +1,61 @@
...
 
 logger = floss.logging_.getLogger(__name__)
 
 
+def identify_language(sample: str) -> str:
+    """
+    Identify the language of the binary given
+    """
+    if is_rust_bin(sample):
+        logger.warning("Rust Binary Detected, Rust binaries are not supported yet. Results may be inaccurate.")
```
williballenthin, line 19:
```suggestion
        logger.warning("Rust binary detected, Rust binaries are not supported yet. Results may be inaccurate.")
```

floss/language_identifier.py
```diff
@@ -1,13 +1,61 @@
...
+    """
+
+    rust_commit_hash = {}
+    # Load the rust version database
+    with open(os.path.join(os.path.dirname(__file__), "rust_version_database.json"), "r") as in_handle:
+        rust_commit_hash = json.load(in_handle)
+
+    # Check if the binary contains the rustc/commit-hash string
+    regex_hash = re.compile(r"rustc/.*[\\\/]library")
+    regex_version = re.compile(r"rustc/[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2}")
```
williballenthin, line 43: please add a comment showing some examples of strings that would match these regular expressions

floss/language_identifier.py
```diff
@@ -1,13 +1,61 @@
...
+
+    with open(sample, "rb") as f:
+        data = f.read()
+        for string, type, span, is_interesting in b2s.extract_all_strings(data, only_interesting=True):
```
williballenthin, line 47: please use existing strings implementation rather that introducing a new dependency:
https://github.com/mandiant/flare-floss/blob/master/floss/strings.py

setup.py
```diff
@@ -16,6 +16,7 @@
     "halo==0.0.31",
     "rich==13.3.4",
     "pefile>=2022.5.30",
+    "binary2strings==0.1.9",
```
williballenthin, line 19: see above, please dont introduce this new dependency

williballenthin:
```suggestion
```

floss/rust_version_database.json
```diff
@@ -0,0 +1 @@
...": "1.58.1", "02072b482a8b5357f7fb5e5637444ae30e423c40": "1.58.0", "f1edd0429582dd29cccacaf50fd134b05593bd9c": "1.57.0", "fc50f328b0353b285421b8ff5d4100966387a997": "1.34.1", "91856ed52c58aa5ba66a015354d1cc69e9779bdf": "1.34.0", "2aa4c46cfdd726e97360c2734835aa3515e8c858": "1.33.0", "9fda7c2237db910e41d6a712e9a2139b352e558b": "1.32.0", "b6c32da9b0481e3e9d737153286b3ff8aa39a22c": "1.31.1", "abe02cefd6cd1916df62ad7dc80161bea50b72e8": "1.31.0", "1433507eba7d1a114e4c6f27ae0e1a74f60f20de": "1.30.1", "da5f414c2c0bfe5198934493f04c676e2b23ff2e": "1.30.0", "17a9dc7513b9fea883dc9505f09f97c63d1d601b": "1.29.2", "b801ae66425cf7c3c71052b19ef8f145b0d0513d": "1.29.1", "4d90ac38c0b61bb69470b61ea2cccea0df48d9e5": "1.24.0", "766bd11c8a3c019ca53febdcd77b2215379dd67d": "1.23.0", "05e2e1c41414e8fc73d0f267ea8dab1a3eeeaa99": "1.22.1", "328886ba2eaf0d3ac7e78ef3ba27eb296d0af3c0": "1.22.0", "3b72af97e42989b2fe104d8edbaee123cdf7c58f": "1.21.0", "f3d6973f41a7d1fb83029c9c0ceaf0f5d4fd7208": "1.20.0", "0ade339411587887bf01bcfa2e9ae4414c8900d4": "1.19.0", "03fc9d622e0ea26a3d37f5ab030737fcca6928b9": "1.18.0", "56124baa9e73f28c0709e59e74783cf234a978cf": "1.17.0", "30cf806ef8881c41821fbd43e5cf3699c5290c16": "1.16.0"}
```
williballenthin, line 1: i think we'll need to update the pyinstaller config to reference this data file.

williballenthin: we should also add a resource that explains where the file came from and how to regenerate this file.

williballenthin (changes requested):
nice work!

floss/rust_version_database.json
williballenthin: using a Python file is a fine idea. i think we should keep the file along the language identification file, so this is the appropriate place today. we can move it fairly easily.

mr-tz: Yes, you can add the script to the `scripts` directory.

floss/language_identifier.py
```diff
@@ -1,13 +1,65 @@
...
+        return "rust"
+    elif is_go_bin(sample):
+        logger.warning("Go Binary Detected, Go binaries are not supported yet. Results may be inaccurate.")
+        logger.warning("Go: Proceeding with analysis may take a long time.")
+        return "go"
+    else:
+        return "unknown"
+
+
+def is_rust_bin(static_strings: Iterable[StaticString]) -> bool:
```
mr-tz, line 34: what about the other strings you've identified for potential ID, should we include them here or leave for now?

floss/main.py
```diff
@@ -546,6 +542,12 @@ def main(argv=None) -> int:
         results.strings.static_strings = static_strings
         results.metadata.runtime.static_strings = get_runtime_diff(interim)
         interim = time()
+    else:
+        with open(sample, "rb") as f:
+            with contextlib.closing(mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ)) as buf:
+                static_strings = list(extract_ascii_unicode_strings(buf, args.min_length))
+
+    language = identify_language(sample=sample, static_strings=static_strings)
```
mr-tz, line 550: if we need these now anyway, let's refactor the logic to always extract and check the language first

then, if static strings are enabled, we can also add them to the results data

floss/rust_version_database.py
```diff
@@ -0,0 +1,137 @@
...
+#
+#   4) Identify the specific Rust version you are interested in and note down its corresponding commit hash.
+#
+#   5) Update the file by adding or updating the commit hash for the respective Rust version.
+#
+#############################################################################################
+
+
+rust_commit_hash = {
+    "59eed8a2aac0230a8b53e89d4e99d55912ba6b35": "1.56.1",
```
mr-tz, line 26: just a nit, but sort these by version value in ascending or descending order

floss/rust_version_database.py
```diff
@@ -0,0 +1,137 @@
+#############################################################################################
```
mr-tz, line 1: for new files, please also add the copyright header as seen in other files

mr-tz (commented):
Nice work! A few comments and questions inline.

floss/language_identifier.py
williballenthin: do the commit strings show up in every one of the test files? or is there a delta between what the commit strings and the other strings cover?

floss/main.py
```diff
@@ -523,26 +523,24 @@ def main(argv=None) -> int:
     interim = time0
     sample_size = os.path.getsize(sample)
 
+    with open(sample, "rb") as f:
+        with contextlib.closing(mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ)) as buf:
```
williballenthin, line 527: im not sure this is the right set of flags on both Linux and Windows. my best guess for this in the past is here: https://github.com/mandiant/flare-floss/blob/25b6f85dd45f7a47bcc90f3c94b328fe2477f0b0/floss/qs/main.py#L610-L620

williballenthin (approved):
nice work!

floss/language_identifier.py
```diff
@@ -1,13 +1,65 @@
...
+    regex_hash = re.compile(r"rustc/.*[\\\/]library")
+
+    # matches strings like "rustc/version/library" e.g. "rustc/1.54.0/library"
+    regex_version = re.compile(r"rustc/[0-9]{1,2}\.[0-9]{1,2}\.[0-9]{1,2}")
+
+    for static_string_obj in static_strings:
+        string = static_string_obj.string
+        if regex_hash.search(string):
+            for commit_hash in rust_commit_hash.keys():
+                if commit_hash in string:
```
mr-tz, line 52: small optimization, `re.match` to extract the hash and verify it's in the database dictionary -- we're probably fine as is, though

floss/main.py
mr-tz: this is what FLOSS used before, but would be good if @Arker123 can confirm here this actually works as expected :)

scripts/extract_rust_hashes.py
```diff
@@ -0,0 +1,85 @@
...
+See the License for the specific language governing permissions and limitations under the License.
+"""
+
+import subprocess
+
+import requests
+from bs4 import BeautifulSoup
+
+page_number = 1
+rust_hases = {}
```
mr-tz, line 28: typo `hases` vs. `hashes`

mr-tz (commented):
nice, I think we're almost good to merge, thanks!

mr-tz (approved):
Great, thanks! First week crushed 🥳

PR #790, opened 2023-06-06
Dotnet binary identification

### LINKED ISSUE #352: Indicate .NET, packed or other binaries FLOSS doesn't handle well
Warn users similar to capa. Arose in issue #346

symbolicvoid:
Added dotnet binary detection to the `language_identifier` module.

partially fixes #352

### DIFF
--- a/floss/language_identifier.py
+++ b/floss/language_identifier.py
@@ -171,4 +171,4 @@ def is_dotnet_bin(pe: pefile.PE) -> bool:
     except IndexError:
         return False
 
-    return dir_entry.Size != 0
+    return dir_entry.Size != 0 and dir_entry.VirtualAddress != 0
--- a/floss/language_identifier.py
+++ b/floss/language_identifier.py
@@ -14,10 +14,10 @@ logger = floss.logging_.getLogger(__name__)
 
 
 class Language(Enum):
-    rust = "rust"
-    go = "go"
-    dotnet = "dotnet"
-    unknown = "unknown"
+    RUST = "Rust"
+    GO = "Go"
+    DOTNET = ".NET"
+    UNKNOWN = "unknown"
 
 
 def identify_language(sample: str, static_strings: Iterable[StaticString]) -> Language:
@@ -27,25 +27,25 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> La
     if is_rust_bin(static_strings):
         logger.warning("Rust Binary Detected, Rust binaries are not supported yet. Results may be inaccurate.")
         logger.warning("Rust: Proceeding with analysis may take a long time.")
-        return Language.rust
+        return Language.RUST
 
     # Open the file as PE for further checks
     try:
         pe = pefile.PE(sample)
     except pefile.PEFormatError as err:
         logger.debug(f"NOT valid PE header: {err}")
-        return Language.unknown
+        return Language.UNKNOWN
 
     if is_go_bin(pe):
         logger.warning("Go Binary Detected, Go binaries are not supported yet. Results may be inaccurate.")
         logger.warning("Go: Proceeding with analysis may take a long time.")
-        return Language.go
+        return Language.GO
     elif is_dotnet_bin(pe):
         logger.warning(".net Binary Detected, .net binaries are not supported yet. Results may be inaccurate.")
         logger.warning(".net: Deobfuscation of strings from .net binaries is not supported yet.")
-        return Language.dotnet
+        return Language.DOTNET
     else:
-        return Language.unknown
+        return Language.UNKNOWN
 
 
 def is_rust_bin(static_strings: Iterable[StaticString]) -> bool:
--- a/floss/language_identifier.py
+++ b/floss/language_identifier.py
@@ -4,6 +4,7 @@ import re
 from typing import Iterable
 
 import pefile
+import dnfile
 
 import floss.logging_
 from floss.results import StaticString
@@ -25,6 +26,10 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> st
         logger.warning("Go Binary Detected, Go binaries are not supported yet. Results may be inaccurate.")
         logger.warning("Go: Proceeding with analysis may take a long time.")
         return "go"
+    elif is_dotnet_bin(sample):
+        logger.warning(".net Binary Detected, .net binaries are not supported yet. Results may be inaccurate.")
+        logger.warning(".net: Proceeding with analysis may take a long time.")
+        return "dotnet"
     else:
         return "unknown"
 
@@ -149,3 +154,11 @@ def verify_pclntab(section, pclntab_va: int) -> bool:
         logger.error("Error parsing pclntab header")
         return False
     return True if pc_quanum in {1, 2, 4} and pointer_size in {4, 8} else False
+
+
+def is_dotnet_bin(sample: str) -> bool:
+    """
+    The 'net' attribute of a dnPE object is null if the file is not a dotnet binary.
+    """
+    file = dnfile.dnPE(sample)
+    return bool(file.net)
--- a/floss/language_identifier.py
+++ b/floss/language_identifier.py
@@ -8,7 +8,6 @@ import pefile
 
 import floss.logging_
 from floss.results import StaticString
-from floss.strings import extract_ascii_unicode_strings
 from floss.rust_version_database import rust_commit_hash
 
 logger = floss.logging_.getLogger(__name__)
@@ -31,29 +30,22 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> La
         return Language.rust
 
     # Open the file as PE for further checks
-    pe = open_pe_file(sample)
-
-    if pe is None:
+    try:
+        pe = pefile.PE(sample)
+    except pefile.PEFormatError as err:
+        logger.debug(f"NOT valid PE header: {err}")
         return Language.unknown
-    elif is_go_bin(pe):
+
+    if is_go_bin(pe):
         logger.warning("Go Binary Detected, Go binaries are not supported yet. Results may be inaccurate.")
         logger.warning("Go: Proceeding with analysis may take a long time.")
         return Language.go
     elif is_dotnet_bin(pe):
         logger.warning(".net Binary Detected, .net binaries are not supported yet. Results may be inaccurate.")
-        logger.warning(".net: Proceeding with analysis may take a long time.")
+        logger.warning(".net: Deobfuscation of strings from .net binaries is not supported yet.")
         return Language.dotnet
     else:
         return Language.unknown
-
-
-def open_pe_file(path: str) -> pefile.PE | None:
-    try:
-        pe = pefile.PE(path)
-    except pefile.PEFormatError as err:
-        logger.debug(f"NOT valid PE header: {err}")
-        return None
-    return pe
 
 
 def is_rust_bin(static_strings: Iterable[StaticString]) -> bool:
--- a/floss/language_identifier.py
+++ b/floss/language_identifier.py
@@ -27,13 +27,7 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> st
         pe = pefile.PE(sample)
     except pefile.PEFormatError as err:
         logger.debug(f"NOT valid PE header: {err}")
-        return "Unknown"
-    except IOError as err:
-        logger.error(f"File does not exist or cannot be accessed: {err}")
-        return "Unknown"
-    except Exception as err:
-        logger.error(f"Unexpected error: {err}")
-        raise
+        return "unknown"
 
     if is_go_bin(pe):
         logger.warning("Go Binary Detected, Go binaries are not supported yet. Results may be inaccurate.")
--- a/floss/language_identifier.py
+++ b/floss/language_identifier.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2023 Mandiant, Inc. All Rights Reserved.
 
 import re
+from enum import Enum
 from typing import Iterable
 
 import pefile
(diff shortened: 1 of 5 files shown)


mr-tz:
Cool, thanks! Have you looked into how dnfile does the .NET detection? Maybe we can do it easily based on pefile.

For the PE check, we could add a function to first validate that we're looking at a proper executable.

symbolicvoid:
I did look at dnfile, but I don't really understand how it does the detection. I did consider using pefile to do the detection before, should I try to implement this using pefile? Might take me time to understand how dnfile does it though.

floss/language_identifier.py
```diff
@@ -149,3 +154,11 @@ def verify_pclntab(section, pclntab_va: int) -> bool:
...
         return False
     return True if pc_quanum in {1, 2, 4} and pointer_size in {4, 8} else False
+
+
+def is_dotnet_bin(sample: str) -> bool:
+    """
+    The 'net' attribute of a dnPE object is null if the file is not a dotnet binary.
+    """
+    file = dnfile.dnPE(sample)
+    return bool(file.net)
```
williballenthin, line 164: for reference, this property is populated here:

https://github.com/malwarefrank/dnfile/blob/527d57033720482c22451be448b743f3590ca7d0/src/dnfile/__init__.py#L173-L181

williballenthin:
> I don't really understand how it does the detection.

the detection is based on the presence of the `pe.net` property:

https://github.com/mandiant/flare-floss/pull/790/files/ca647d9bc4801c1e82cf8d9679ca994288f12af8#diff-0aaada7d6f6bc2b49663b4b4d03fc4ee53257871d51392ad44722c8d9066e3dbR164

if the property is present and not empty/None, then its considered a .NET module. this property is populated here:
https://github.com/malwarefrank/dnfile/blob/527d57033720482c22451be448b743f3590ca7d0/src/dnfile/__init__.py#L173-L181
image


basically, check to see if there's a `IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR` structure, and if so, use that. so, i think we can implement this using pure pefile in just about the name number of lines (perhaps 10 or so).

does that help?

williballenthin:
```py
            try:
                directory_index = pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR"]
                dir_entry = pe.OPTIONAL_HEADER.DATA_DIRECTORY[directory_index]
            except IndexError:
                return False
            return dir_entry is not None and maybe a little more validation
```

it might even boil down to something this short

symbolicvoid:
Thank you! This is very helpful. I agree we should not add another dependency for now (considering we aren't really analyzing the dotnet file yet either). I will try to implement this using pefile.

williballenthin:
awesome, thanks!

symbolicvoid:
> it might even boil down to something this short.

Turns out its very similar to this! I managed to make it work with this:

```python
    try:
        directory_index = pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR"]
        dir_entry = pe.OPTIONAL_HEADER.DATA_DIRECTORY[directory_index]
    except IndexError:
        return False

    return dir_entry.Size != 0
```

If the file is not a .net file, pefile still seems to represent its `IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR` but with a size of 0.

image_2023-06-07_130747960

 But in case of a .net file the size of the Structure we get from `OPTIONAL_HEADER.DATA_DIRECTORY` is non-zero.

image_2023-06-07_130356395

I tested it with a .net file from dnfile-testfiles and some non-dotnet files, should I try it on some more executables? Also, I don't know if this will ever raise `IndexError`, should we return false when any error occurs?

williballenthin:
```
isort --profile black --length-sort --line-width 120 --skip-glob "*_pb2.py" . ; \
black -l 120 --extend-exclude ".*_pb2.py" . ; \
ruff check --config .github/ruff.toml . ; \
pycodestyle --exclude="*_pb2.py" --show-source floss/ scripts/ tests/ ; \
mypy --config-file .github/mypy/mypy.ini --check-untyped-defs floss/ scripts/ tests/
```

symbolicvoid:
> 2. add a sample .NET file to https://github.com/mandiant/flare-floss-testfiles

Can we reuse the samples from [dnfile-testfiles](https://github.com/malwarefrank/dnfile-testfiles) or should I create one?

symbolicvoid:
Also, I modified the code to open the pefile before checking for go and .net and after checking for rust. While it does break the if..elif block, we don't need pefile to check for rust binary. Is there some other way I should order it?
```python
    if is_rust_bin(static_strings):
     
    try:
        pe = pefile.PE(sample)
    except pefile.PEFormatError as err:
        
    except IOError as err:
       
    except Exception as err:
        

    if is_go_bin(pe):
        
    elif is_dotnet_bin(pe):
      
```

williballenthin:
> Can we reuse the samples from [dnfile-testfiles](https://github.com/malwarefrank/dnfile-testfiles) or should I create one?

yes, but please copy the file into the floss-testfiles repository, since we don't currently have dnfile-testfiles as a submodule

symbolicvoid:
I have never written any tests so far, so I am a bit confused on what to do here...
I thought we could move the opening of the pe file into a function:

```python
def open_pe_file(path: str) -> pefile.PE | None:
    """
    Takes file path and returns the PE object for the file.
    """
    try:
        pe = pefile.PE(path)
    except pefile.PEFormatError as err:
        logger.debug(f"NOT valid PE header: {err}")
        return None
    except IOError as err:
        logger.error(f"File does not exist or cannot be accessed: {err}")
        return None
    except Exception as err:
        logger.error(f"Unexpected error: {err}")
        raise

    return pe
```

Now this function will return None if it couldn't open a file, and we can check this in the `identify_language` function.
```python
def identify_language(sample: str, static_strings: Iterable[StaticString]) -> str:
    """
    Identify the language of the binary given
    """
    if is_rust_bin(static_strings):

    # Open the file as PE for further checks
    pe = open_pe_file(sample)

    if pe is None:
        return "Unknown"
    elif is_go_bin(pe):

    elif is_dotnet_bin(pe):

    else:
        return "unknown"
```

However, I don't know how to handle it within the tests, do we check for None there too? Is there some error to be raised when the file is None?

```python
def test_go_binary_detection(binary_file, expected_result):
    CD = os.path.dirname(__file__)
    abs_path = os.path.normpath(os.path.join(CD, binary_file))
    # check if the file exists
    assert os.path.exists(abs_path) == True, f"File {binary_file} does not exist"

    pe = open_pe_file(abs_path)
    # Check for None here?

    is_go_binary = is_go_bin(pe)
    # Check the expected result
    assert is_go_binary == expected_result, f"Expected: {expected_result}, Actual: {is_go_binary}"
```

floss/language_identifier.py
```diff
@@ -149,3 +156,18 @@ def verify_pclntab(section, pclntab_va: int) -> bool:
...
+    Checks the IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR entry in the OPTIONAL_HEADER of the file.
+    If the entry is not found, or if its size is 0, the file is not a .net file.
+    """
+    try:
+        directory_index = pefile.DIRECTORY_ENTRY["IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR"]
+        dir_entry = pe.OPTIONAL_HEADER.DATA_DIRECTORY[directory_index]
+    except IndexError:
+        return False
+
+    return dir_entry.Size != 0
```
mr-tz, line 173: not sure if this really improves things, but `dir_entry.Size` and `dir_entry.VirtualAddress` should be non zero.

floss/language_identifier.py
```diff
@@ -21,10 +21,28 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> st
...
         logger.warning("Rust: Proceeding with analysis may take a long time.")
         return "rust"
-    elif is_go_bin(sample):
+
+    # Open the file as PE for further checks
+    try:
+        pe = pefile.PE(sample)
+    except pefile.PEFormatError as err:
+        logger.debug(f"NOT valid PE header: {err}")
+        return "Unknown"
```
mr-tz, line 30:
```suggestion
        return "unknown"
```

floss/language_identifier.py
```diff
@@ -21,10 +21,28 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> st
...
+    except IOError as err:
+        logger.error(f"File does not exist or cannot be accessed: {err}")
+        return "Unknown"
```
mr-tz, line 33: This is handled when we parse the arguments, so can be removed here.
```suggestion
```

floss/language_identifier.py
```diff
@@ -21,10 +21,28 @@ def identify_language(sample: str, static_strings: Iterable[StaticString]) -> st
...
+    except Exception as err:
+        logger.error(f"Unexpected error: {err}")
+        raise
```
mr-tz, line 36:
```suggestion
```

mr-tz: good time to clean up this code (I know you just copy-pasted from below)

mr-tz:
Each test should focus on one thing so you can assume the file exists and test that it is / is not .NET.

You could then add another test to verify the behavior of the program when a file cannot be opened.

symbolicvoid:
Uh, how can I make `open_pe_file` return PE or None?

williballenthin:
>Uh, how can I make open_pe_file return PE or None?

I dont quite understand the question. can you rephrase it with a bit more detail, and maybe what you've tried so far?

symbolicvoid:
> I dont quite understand the question. can you rephrase it with a bit more detail, and maybe what you've tried so far?

Apologies for the confusion. So I moved the opening of the pe file into a function `open_pe_file`:

```python
def open_pe_file(path: str) -> pefile.PE | None:
    try:
        pe = pefile.PE(path)
    except pefile.PEFormatError as err:
        logger.debug(f"NOT valid PE header: {err}")
        return None
    return pe
```

Now the function return type `pefile.PE | None` causes CI/CD issues. This return type was recommended to me by pycharm as the function can return either a PE object, or None. So what return type should I use here?

The reason why I moved this into a separate function is to allow tests to run on individual binary check functions. Previously, `is_go_bin` took the file path as input, now it takes a PE object. This means we need to modify the test cases so that it passes a PE object of the file instead of the path. Previous:

```python
def test_go_binary_detection(binary_file, expected_result):
    CD = os.path.dirname(__file__)
    abs_path = os.path.normpath(os.path.join(CD, binary_file))
    # check if the file exists
    assert os.path.exists(abs_path) == True, f"File {binary_file} does not exist"

    is_go_binary = is_go_bin(abs_path)
    # Check the expected result
    assert is_go_binary == expected_result, f"Expected: {expected_result}, Actual: {is_go_binary}"
```

So instead of having to open the file and handle errors in the test script, I thought we could move the opening of the file into a function, import that function and use it instead:

```python
def test_go_binary_detection(binary_file, expected_result):
    CD = os.path.dirname(__file__)
    abs_path = os.path.normpath(os.path.join(CD, binary_file))
    # check if the file exists
    assert os.path.exists(abs_path) == True, f"File {binary_file} does not exist"

    pe = open_pe_file(abs_path)

    is_go_binary = is_go_bin(pe)
    # Check the expected result
    assert is_go_binary == expected_result, f"Expected: {expected_result}, Actual: {is_go_binary}"
```

Another issue here is that the `pe` variable can be `None`, meaning it can break the functions as we do the `None` check in the `identify_language` function, not the `is_go_bin` function.

```python
def identify_language(sample: str, static_strings: Iterable[StaticString]) -> Language:
   
    if is_rust_bin(static_strings):

    # Open the file as PE for further checks
    pe = open_pe_file(sample)

    if pe is None:
        return Language.unknown 
    elif is_go_bin(pe):
        
    elif is_dotnet_bin(pe):
        
    else:
        return Language.unknown
```

While we could move the `None` check into those functions, I do have a question. Could we just test the entire `language_identify` function instead of checking individual functions? However we might need to provide the `static_strings` parameter from the test script..

```python
def identify_language(sample: str, static_strings: Iterable[StaticString]) -> Language:
```

mr-tz:
I think you can rebase to master (`git rebase master` after fetch/pull locally on that branch) or update the submodule in this PR (`git add tests/data`).

symbolicvoid:
> I think you can rebase to master (`git rebase master` after fetch/pull locally on that branch) or update the submodule in this PR (`git add tests/data`).

Sorry I couldn't make that work.. I added the submodule again with `git submodule add`, please take a look if this is correct!

mr-tz:
No worries, I think I can fix that if needed. It'll probably be later today or tomorrow though.

floss/language_identifier.py
```diff
@@ -1,32 +1,51 @@
...
 
 logger = floss.logging_.getLogger(__name__)
 
 
-def identify_language(sample: str, static_strings: Iterable[StaticString]) -> str:
+class Language(Enum):
+    rust = "rust"
+    go = "go"
+    dotnet = "dotnet"
+    unknown = "unknown"
```
williballenthin, line 20: this is a great improvement, thank you!

williballenthin, line 20:
```suggestion
    RUST = "Rust"
    GO = "Go"
    DOTNET = ".NET"
    UNKNOWN = "unknown"
```

williballenthin: minor style suggestions. recommend using UPPERCASE for constants

floss/language_identifier.py
williballenthin:
```suggestion
    return dir_entry.Size != 0 and dir_entry.VirtualAddress != 0
```

tests/test_language_id.py
```diff
@@ -2,24 +2,28 @@
...
     [
-        ("data/src/go-hello/bin/go-hello.exe", True),
-        # ("data/src/go-hello/bin/go-hello", True), should be true, but it fails as elf file format is not supported
-        ("data/test-decode-to-stack.exe", False),
-        ("data/src/shellcode-stackstrings/bin/shellcode-stackstrings.bin", False),
+        ("data/src/go-hello/bin/go-hello.exe", Language.go),
+        ("data/src/rust-hello/bin/rust-hello.exe", Language.rust),
+        ("data/test-decode-to-stack.exe", Language.unknown),
+        ("data/src/dotnet-hello/bin/dotnet-hello.exe", Language.dotnet),
+        ("data/src/shellcode-stackstrings/bin/shellcode-stackstrings.bin", Language.unknown),
```
williballenthin, line 16: ❤️

williballenthin (approved):
great work @symbolicvoid!

one naming suggestion pending. if you want to address - we'll hold off merging. otherwise, we can merge as is. let us know when we should pull the trigger.

williballenthin:
also, please update the changelog so that we can recognize your work during the next release

symbolicvoid:
> also, please update the changelog so that we can recognize your work during the next release

I don't think we have a changelog in this project, but I made the name changes you suggested, so this PR should be ready to merge!

williballenthin:
ah, right 🤦🏼  well we probably should have one, but in the meantime, we'll be sure to include your effort in the release notes, thanks again!

mr-tz:
Great job, thanks @symbolicvoid!

Issue #805, opened 2023-06-14
Analyzing the .text Segment in a Go Binary


Arker123:
This issue involves the analysis of the .text segment within a Go binary. The focus is on understanding the strings contained in the .text segment.

This issue is a continuation of the discussion in https://github.com/mandiant/flare-floss/issues/779

mr-tz:
Cool, I think that's a good start for an optimization for short strings.

A few follow up questions:
1. Have you looked at other binaries? Here the instruction is `cmp` but we want to handle others, e.g., `mov` as well.
2. What about strings that are longer than 8/16 characters? I think that case is actually easier (that's why I suggested to skip case 3 originally in https://github.com/mandiant/flare-floss/issues/779#issuecomment-1581047424).

mr-tz:
Have you checked out the code of various Go binaries (different source) to see how strings are stored and accessed?

williballenthin:
"strings in the .text segment"

from here: https://stackoverflow.com/a/35342389
we are linked to this commit in the Go compiler: https://github.com/golang/go/commit/69cd91a5981c49eaaa59b33196bdb5586c18d289
which introduces an optimization in Go 1.5 for comparing a string with a constant string. in this scenario. the constant string is converted to a stack-allocated temporary byte[] variable, which leads to inlining like seen in the above examples. im not sure yet where the decision to use 8 vs 16 bytes as the max inline size is, but its possible thats its far away in the compiler source.

anyways, this clarifies which strings we expect to see in this pattern: string constants used during map lookup, constant concatenation, and comparisons
https://github.com/golang/go/blob/c3db64c0f45e8f2d75c5b59401e0fc925701b6f4/src/runtime/string.go#L146-L149

mr-tz:
> In versions preceding Go 1.6, while analyzing the .text segment, it is observed that the reference to the string exists, but the string length information is not readily accessible.

As per our discussion we're not targeting versions <1.6 to focus on more recent versions. But please add an issue with some details like sample (hash) and what you've learned so far.

mr-tz:
Interesting, which samples have you analyzed above?

mr-tz:
Great, would love to see the current performance (via the analysis script) on
1. our testfiles across Go versions and
2. a bunch of random Go samples

Ideally we also script/automate this and integrate into our tests (CI), see #807

PR #792, opened 2023-06-09
Extracting language specific strings from golang #779

Arker123:
This pull request (PR) aims to enhance the codebase by implementing a feature to extract language-specific strings from Go (Golang) binaries. 

TODO:-
- Extract strings from the .text segment.
- Add comprehensive tests for various versions of Go.

PR #810, opened 2023-06-15
Added go binary analysis file

Arker123:
This PR adds an "analyze.py" file to the repository. 
https://github.com/mandiant/flare-floss/issues/805#issuecomment-1592519918


scripts/analyse-go-binary.py
```diff
@@ -0,0 +1,200 @@
...
+        longest_off = this_off
+        longest_size = off - this_off
+
+    if longest_size > 0:
+        return (longest_off, longest_size)
+
+    return (None, 0)
+
+
+def extract_go_strings(
```
mr-tz, line 61: let's keep the functionality separate and reuse the Go extractor code

scripts/analyse-go-binary.py
```diff
@@ -0,0 +1,200 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+final_size = 0
+
+MIN_STR_LEN = 4
+
+
+def find_longest_string(bindata):
```
mr-tz, line 23: I feel this is limiting us to a subset of all the strings. Why not look for all the strings (maybe in specific regions of the file like .data/.rdata) and calculate what we cover? Goal should be at least to recover 80% of all the strings in the file.

scripts/analyse-go-binary.py
```diff
@@ -0,0 +1,200 @@
...
+    final_length = 0
+
+    for strings_obj in static_strings:
+        string = strings_obj.string
+        final_length += len(string)
+
+    try:
+        print(final_length * 100 / final_size)
+    except:
+        pass
```
mr-tz, line 196: why would this throw an exception, let's handle it differently if final_size is 0, for example

mr-tz (changes requested):
We'll also want to add a way to see which strings we've missed. One naive idea for that is to create
1. a file with all the static strings
2. a file where the strings we've extracted are removed (overwritten with a known character)

Then we could do a diff on the files and see what we've missed.

scripts/analyse-go-binary.py
mr-tz: instead of working off the master branch, you can branch off / merge into the branch from #792

Issue #798, opened 2023-06-12
use pathlib.Path to represent file system paths instead of strings


williballenthin:
throughout the codebase, replace string type variables with `pathlib.Path` type variables that represent file system paths. these objects have convenient methods/properties for common operations, as well as make clear how the data should be used (as a file system path, not a string to display to the user, or whatever).

mr-tz:
Closing as complete. There may be a few outstanding instances which we can capture via `flake8-use-pathlib` (see #825).
Thanks, @DiegoRomeo!

PR #819, opened 2023-06-26
change os.path to pathlib.Path #798

DiegoRomeo:


williballenthin:
thank you @DiegoRomeo! i will do a close review of this PR in the morning. i can see you're already on the right track of using pathlib. 

i'm general, i'm likely to provide some feedback suggesting that pathlib.Path instances be passed throughout the program whenever possible, so there should be as few conversions to/from str as you can. also, it may help to introduce type annotations to make the types of function arguments and return values more clear. again, i can provide more specifics in the morning.

floss/main.py
```diff
@@ -336,7 +337,8 @@ def is_supported_file_type(sample_file_path):
     :param sample_file_path:
     :return: True if file type is supported, False otherwise
     """
-    with open(sample_file_path, "rb") as f:
+    sample_file_path = Path(sample_file_path)
```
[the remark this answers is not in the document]
williballenthin: also suggest adding type annotations to function arg and return value.

floss/main.py
```diff
@@ -407,30 +409,31 @@ def get_default_root() -> str:
         # so we'll fetch this attribute dynamically.
         return getattr(sys, "_MEIPASS")
     else:
-        return os.path.join(os.path.dirname(__file__))
+        return str(Path(__file__).resolve().parent)
```
williballenthin, line 412: return pathlib.Path


...and so forth.

floss/main.py
```diff
@@ -407,30 +409,31 @@ def get_default_root() -> str:
...
 
 
 def get_signatures(sigs_path):
```
williballenthin, line 415: this function should accept a pathlib.Path

and add type hints

floss/main.py
```diff
@@ -442,7 +445,8 @@ def get_static_strings(sample: str, min_length: int) -> list:
     """
     Returns list of static strings from the file which are above the minimum length
     """
-    with open(sample, "rb") as f:
+    sample_path = Path(sample)
```
williballenthin, line 448: the function should accept a pathlib.Path already

floss/main.py
```diff
@@ -537,7 +541,7 @@ def main(argv=None) -> int:
 
     time0 = time()
     interim = time0
-    sample_size = os.path.getsize(sample)
+    sample_size = Path(sample).stat().st_size
```
williballenthin, line 544: this is a good update

setup.py
```diff
@@ -21,13 +21,15 @@
 # this sets __version__
 # via: http://stackoverflow.com/a/7071358/87207
 # and: http://stackoverflow.com/a/2073599/87207
-with open(os.path.join("floss", "version.py"), "r") as f:
+file_path = Path("floss") / "version.py"
+with file_path.open("r") as f:
     exec(f.read())
```
williballenthin, line 26: can probably also do:

```py
exec(file_path.read_text())
```

tests/fixtures.py
```diff
@@ -1,18 +1,20 @@
...
-CD = os.path.dirname(__file__)
+CD = Path(__file__).resolve().parent
 
 
 @pytest.fixture
 def exefile() -> str:
     # decode-in-place is among the fastest samples in data/src
-    return os.path.join(CD, "data", "src", "decode-in-place", "bin", "test-decode-in-place.exe")
+    path = CD / "data" / "src" / "decode-in-place" / "bin" / "test-decode-in-place.exe"
+    return str(path)
```
williballenthin, line 14: return pathlib.Path

williballenthin (changes requested):
as described yesterday, definitely on the right track. please continue to make more updates that use pathlib.Path more consistently - we should use this whenever possible and avoid using str as arguments and return values from functions. updating the type hints along the way can make it easier, especially if you're using mypy to highlight potential problems.

does the above make sense? can i clarify anything?

DiegoRomeo:
Ok thanks for your support and feedback! 
I am already working to change all the string paths used as arguments and return values (and also the type hints).
While working on main.py I noticed some functions from external libraries like viv_utils that only accept string arguments for paths, so I convertet the pathlib.Path back to a string type.
Also I noticed that there is a dataclass "Metadata" (which is in floss/results.py): should I change the argument type to Path?

tests/fixtures.py
DiegoRomeo: I am not sure on how to change this code because the parser tries to convert the sample path to an argparse.FileType and does some string related operations that are not allowed with pathlib.Path objects.

floss/main.py
```diff
@@ -533,11 +534,11 @@ def main(argv=None) -> int:
 
         return 0
 
-    results = ResultDocument(metadata=Metadata(file_path=sample, min_length=args.min_length), analysis=analysis)
+    results = ResultDocument(metadata=Metadata(file_path=str(sample), min_length=args.min_length), analysis=analysis)
```
williballenthin, line 537: this is fine

tests/fixtures.py
williballenthin: ah, ok, i understand how this might interact weirdly with pytest and/or argparse. no worries.

williballenthin (approved):
this is great!

do you think there is any outstanding work @DiegoRomeo? or are we ready to merge?

DiegoRomeo:
Just used isort on changed files to improve code style

Issue #830, opened 2023-07-12
Add Language-Specific Command-Line Flag for Go Binary


Arker123:
In scenarios where our identifier might produce false positives, we want to further enhance the user experience. To achieve this, we plan to provide users with the flexibility to manually select the appropriate languages for their binaries.

- Implement a new command-line flag, such as --language go, that allows users to specify Go as the language of the binary.

mr-tz:
`--language [go, rust, dotnet]` could make sense, and maybe `--language none` to skip language identification

alternatives:
- `--lang`
- `--format`
- `--extract`

stevemk14ebr:
I think it should be feasible to signature some common internal routines in each of these languages to suggest/auto apply a flag such as this

Arker123:
Thank you for sharing your thoughts. We actually have a script in place for this [here](https://github.com/mandiant/flare-floss/blob/master/floss/language/identify.py). The current issue was raised to address potential false positives with the identifier, aiming to enhance the user experience. To mitigate this, users have the flexibility to manually select the languages. I will certainly be more cautious when describing the issue in the future.

stevemk14ebr:
Thank you for letting me know, I was not aware! 

I took a look at the identification used for go just now, it may be nice to extend the Go detections to include binaries with custom magics. This can be the case for garble obfuscated Go binaries. Here https://github.com/mandiant/flare-floss/blob/master/floss/language/identify.py#L85

Using strings from the linked go runtime may be good to signature off of in addition, or perhaps signature scans on runtime code for each arch x version. A strings based approach may be sufficient though.

PR #833, opened 2023-07-13
 Add Language-Specific Command-Line Flag for Go Binary #830

Arker123:
This pull request introduces a language-specific command-line flag, "--extract", to enhance the functionality of our program. The flag allows users to specify the language they want to extract information from, with options for "go", "rust", and "dotnet". Additionally, the flag supports an "unknown" option.

Usage example; `floss -e <language> <binary>`
Resolves https://github.com/mandiant/flare-floss/issues/830

### DIFF
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -505,6 +505,10 @@ def find_string_blob_range(pe: pefile.PE, struct_strings: List[StructString]) ->
 
     # pick the mid string, so that we avoid any junk data on the edges of the string blob
     run_mid = (run_start + run_end) // 2
+
+    if run_mid == 0:
+        raise ValueError("no string blob found")
+
     instance = struct_strings[run_mid]
 
     s = read_struct_string(pe, instance)
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -505,7 +505,6 @@ def find_string_blob_range(pe: pefile.PE, struct_strings: List[StructString]) ->
 
     # pick the mid string, so that we avoid any junk data on the edges of the string blob
     run_mid = (run_start + run_end) // 2
-
     instance = struct_strings[run_mid]
 
     s = read_struct_string(pe, instance)
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -506,9 +506,6 @@ def find_string_blob_range(pe: pefile.PE, struct_strings: List[StructString]) ->
     # pick the mid string, so that we avoid any junk data on the edges of the string blob
     run_mid = (run_start + run_end) // 2
 
-    if run_mid == 0:
-        raise ValueError("no string blob found")
-
     instance = struct_strings[run_mid]
 
     s = read_struct_string(pe, instance)
@@ -573,6 +570,9 @@ def get_string_blob_strings(pe: pefile.PE, min_length) -> Iterable[StaticString]
 
     with floss.utils.timing("find struct string candidates"):
         struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
+
+    if struct_strings == []:
+        raise ValueError("no struct string candidates found")
 
     with floss.utils.timing("find string blob"):
         string_blob_start, string_blob_end = find_string_blob_range(pe, struct_strings)
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -572,7 +572,7 @@ def get_string_blob_strings(pe: pefile.PE, min_length) -> Iterable[StaticString]
         struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
 
     if struct_strings == []:
-        raise ValueError("no struct string candidates found")
+        raise ValueError("No struct string candidates found. Is this a Go binary?")
 
     with floss.utils.timing("find string blob"):
         string_blob_start, string_blob_end = find_string_blob_range(pe, struct_strings)
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -569,10 +569,14 @@ def get_string_blob_strings(pe: pefile.PE, min_length) -> Iterable[StaticString]
     image_base = pe.OPTIONAL_HEADER.ImageBase
 
     with floss.utils.timing("find struct string candidates"):
-        struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
+        try:
+            struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
+            if struct_strings == []:
+                raise ValueError("No struct string candidates found. Is this a Go binary?")
 
-    if struct_strings == []:
-        raise ValueError("No struct string candidates found. Is this a Go binary?")
+        except ValueError as e:
+            logger.warning("Failed to find struct string candidates: %s", e)
+            return
 
     with floss.utils.timing("find string blob"):
         string_blob_start, string_blob_end = find_string_blob_range(pe, struct_strings)
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -569,13 +569,9 @@ def get_string_blob_strings(pe: pefile.PE, min_length) -> Iterable[StaticString]
     image_base = pe.OPTIONAL_HEADER.ImageBase
 
     with floss.utils.timing("find struct string candidates"):
-        try:
-            struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
-            if struct_strings == []:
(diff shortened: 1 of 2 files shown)


floss/language/go/extract.py
```diff
@@ -505,6 +505,10 @@ def find_string_blob_range(pe: pefile.PE, struct_strings: List[StructString]) ->
 
     # pick the mid string, so that we avoid any junk data on the edges of the string blob
     run_mid = (run_start + run_end) // 2
+
+    if run_mid == 0:
+        raise ValueError("no string blob found")
```
mr-tz, line 510: this should be caught somewhere or handled differently

floss/main.py
```diff
@@ -151,6 +151,17 @@ def make_parser(argv):
         help="path to sample to analyze",
     )
 
+    language_group = parser.add_argument_group("language arguments")
+    language_group.add_argument(
+        "-e",
+        "--extract",
```
mr-tz, line 157: out my proposals in issue 830 I like `extract` the least, @williballenthin thoughts?

floss/main.py
```diff
@@ -151,6 +151,17 @@ def make_parser(argv):
         help="path to sample to analyze",
     )
 
+    language_group = parser.add_argument_group("language arguments")
```
mr-tz, line 154: we don't need a new group here, just an added argument

floss/main.py
```diff
@@ -151,6 +151,17 @@ def make_parser(argv):
...
+    language_group.add_argument(
+        "-e",
+        "--extract",
+        dest="language",
+        type=str,
+        choices=[l.value for l in Language],
+        default=[],
```
mr-tz, line 161: use "" or "none" here?

floss/main.py
Arker123: Actually, I was considering adding the "--language" option, but it seems that the "-l" flag is already in use. Do you have any suggestions?

Arker123: Regarding the flag options, what are your thoughts on using both '-g' and '--language' together? Do you think it would be feasible or could it lead to confusion?

mr-tz: We could just use `--language` without a short argument.

williballenthin: yeah, i think we should reserve short arguments for those that will be used very often or are very obvious/expected. `--language` seems like an advanced feature, so lets restrict it to a long argument.

mr-tz: good, and then add it to the advanced options, like `"--format"`

floss/language/go/extract.py
```diff
@@ -570,6 +571,9 @@ def get_string_blob_strings(pe: pefile.PE, min_length) -> Iterable[StaticString]
     with floss.utils.timing("find struct string candidates"):
         struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
 
+    if struct_strings == []:
+        raise ValueError("no struct string candidates found")
```
mr-tz, line 575: this is still not caught/handled

Arker123: Could you please take a look at the latest commit? I made some updates to address the issue. If you still notice any unresolved concerns, I would greatly appreciate your suggestions on how to handle it properly. Thanks😄

floss/main.py
```diff
@@ -197,6 +186,14 @@ def make_parser(argv):
         default="auto",
         help="select sample format, %s" % format_help if show_all_options else argparse.SUPPRESS,
     )
+    advanced_group.add_argument(
+        "--language",
+        dest="language",
```
mr-tz, line 191:
```suggestion
```

floss/main.py
```diff
@@ -534,7 +542,7 @@ def main(argv=None) -> int:
     lang_id = identify_language(sample, static_strings)
 
     # set language configurations
-    if lang_id == Language.GO:
+    if (lang_id == Language.GO and args.language == "") or args.language == Language.GO.value:
         results.metadata.language = Language.GO.value
```
mr-tz, line 546:
```suggestion
    if args.language == Language.GO.value or lang_id == Language.GO:
        results.metadata.language = Language.GO.value
```

mr-tz: would this work?

floss/main.py
```diff
@@ -186,6 +186,14 @@ def make_parser(argv):
         default="auto",
         help="select sample format, %s" % format_help if show_all_options else argparse.SUPPRESS,
     )
+    advanced_group.add_argument(
+        "--language",
+        dest="language",
+        type=str,
+        choices=[l.value for l in Language],
```
mr-tz, line 193: what about `none` to force no language ID?

mr-tz (changes requested):

floss/language/go/extract.py
mr-tz: we need to catch the exception or return and empty tuple and see if that works

what happens otherwise when run on a binary where the string blob is not found?

floss/main.py
Arker123: Considering the possibility of false positives, what if our language identifier incorrectly detects a binary as Go, even though it isn't?

mr-tz: then user's have to specify `--language none` explicitly?

the conditions should be similar to your code here, or are they not?

Arker123: > then user's have to specify `--language none` explicitly?

Indeed, if there are false positives, they have the option to do so.

> the conditions should be similar to your code here, or are they not?

I may be wrong, but the code appears to already enable users to specify the language explicitly using the --language flag and the ability to override the automatic detection if necessary?

floss/language/go/extract.py
Arker123: > what happens otherwise when run on a binary where the string blob is not found?

image

floss/language/go/extract.py
```diff
@@ -505,6 +505,7 @@ def find_string_blob_range(pe: pefile.PE, struct_strings: List[StructString]) ->
 
     # pick the mid string, so that we avoid any junk data on the edges of the string blob
     run_mid = (run_start + run_end) // 2
+
```
mr-tz, line 508:
```suggestion
```

floss/language/go/extract.py
```diff
@@ -568,7 +569,14 @@ def get_string_blob_strings(pe: pefile.PE, min_length) -> Iterable[StaticString]
...
     with floss.utils.timing("find struct string candidates"):
-        struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
+        try:
+            struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
+            if struct_strings == []:
+                raise ValueError("No struct string candidates found. Is this a Go binary?")
+
+        except ValueError as e:
+            logger.warning("Failed to find struct string candidates: %s", e)
+            return
```
mr-tz, line 579:
```suggestion
        struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
        if struct_strings == []:
            logger.warning("Failed to find struct string candidates: Is this a Go binary?")
            return
```

floss/main.py
mr-tz: my second comment is regarding the logic of the `if` here, my goal was to simplify it

mr-tz: the same then applies to the ones below

mr-tz (changes requested):

floss/language/go/extract.py
mr-tz: exactly, we want to handle this nicely

floss/main.py
Arker123: I wanted to provide a contradictory example:
```
if args.language == Language.GO.value or lang_id == Language.GO:
    results.metadata.language = Language.GO.value
```

Consider a scenario where we have a Rust binary that is incorrectly identified as Go by our identification script. However, the user is aware that it is a Rust binary. In this case, the `lang_id` variable would be `True`. Since there is an "OR" statement in the expression, it will evaluate to `True`.

I believed this example could be relevant in highlighting how the expression may lead to inaccurate language detection under specific circumstances.

I am open to further discussion. Thanks😄

mr-tz: good point, let's leave as is

floss/main.py
```diff
@@ -186,6 +186,13 @@ def make_parser(argv):
         default="auto",
         help="select sample format, %s" % format_help if show_all_options else argparse.SUPPRESS,
     )
+    advanced_group.add_argument(
+        "--language",
+        type=str,
+        choices=[l.value for l in Language if l != Language.UNKNOWN] + ["none"],
+        default="",
+        help="language of the sample",
```
mr-tz, line 194:
```suggestion
        help="language of the sample" if show_all_options else argparse.SUPPRESS,
```

mr-tz (approved):
thanks

mr-tz:
good to merge, @Arker123?

Issue #807, opened 2023-06-14
Add test cases to CI


mr-tz:
To ensure our algorithms work across a variety of Go samples:

```[tasklist]
### Tasks
- [x] know expected strings for all tracked Go versions: https://github.com/mandiant/flare-floss-testfiles/tree/master/src/go-binaries-all-versions
- [x] add test samples of Go sample where source is unknown
- [x] using script from #806 ensure we at least cover x% of strings
```

mr-tz:
@Arker123, can you please make sure all points are covered here and add tests if necessary?

mr-tz:
we should add tests for the Go samples @Arker123 generated for various Go versions

Arker123:
Hey, during testing, I made a discovery. Our program successfully extracted the following strings from the sample code used for compiling the binaries with a version greater than or equal to go1.12:

    "Something is wrong with your computer"
    "You Cracked it, A Hero is born"
    "Don't Worry, Relax, Chill and Try harder"
    "Enter Password:"

These strings were successfully identified and extracted by our program.

Arker123:
However, I must inform you that this functionality doesn't seem to work for binaries compiled with go1.11

Arker123:
Additionally, I was wondering if you have any suggestions on how I can add a test for the binaries located at https://github.com/Arker123/flare-floss-testfiles/tree/master/src/go-binaries-all-versions/bin. The challenge is that these binaries are actually compressed in a zip file. It's an interesting hurdle to overcome! 😅

Arker123:
During the development of the Go coverage script, the value we needed to determine was previously determined through guessing. 
![image](https://github.com/mandiant/flare-floss/assets/94680887/6e6905a7-f72d-46a6-af5b-234ade0a13bd)

To visualize this process, I plotted the results on a graph.
To determine the appropriate value, I adopted a brute force approach and experimented with different ranges. 
Initially, I used a range with start = 10, end = 1000, and step = 200. The graph below shows the results of this range:

![image](https://github.com/mandiant/flare-floss/assets/94680887/0a7e5c49-9dbf-4f48-87cd-03c967341cc3)

After narrowing down the range, we further experimented with start = 1650, end = 3200, and step = 50. The graph below illustrates the results obtained from this range:
![image](https://github.com/mandiant/flare-floss/assets/94680887/d4d7a0ac-9d34-4d3e-be1f-2653b9d5ea55)

[...]

PR #834, opened 2023-07-14
Update go coverage script

Arker123:
See: https://github.com/mandiant/flare-floss/issues/807#issuecomment-1636087673

### DIFF
--- a/floss/language/go/coverage.py
+++ b/floss/language/go/coverage.py
@@ -80,7 +80,6 @@ def get_extract_stats(pe, all_ss_strings: List[StaticString], go_strings, min_le
             continue
 
         if len(s.string) <= 2800:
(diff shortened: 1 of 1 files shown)


floss/language/go/coverage.py
```diff
@@ -79,8 +79,7 @@ def get_extract_stats(pe, all_ss_strings: List[StaticString], go_strings, min_le
         if secname != ".rdata":
             continue
 
-        if len(s.string) <= 30:
-            # guessed value right now
+        if len(s.string) <= 2800:
```
mr-tz, line 82: you should add a comment and reference to your research here on why this value was chosen

mr-tz (approved):
great, thanks

Issue #791, opened 2023-06-06
extract strings from binaries compiled from Rust


ghost:

wballenthin:
background research: 
  - https://research.checkpoint.com/2023/rust-binary-analysis-feature-by-feature/


(please update this comment directly with more references, if you'd like)

Arker123:
Initially, Rust strings appear as follows:
![Screenshot (39)](https://github.com/mandiant/flare-floss/assets/94680887/e02ff36f-751d-4683-ac92-59275b46bf3d)
Here are some key points to note:
- Rust strings terminate with a null character.
- The string extends its length by appending `00` in order to fit within an 8-byte block (for amd64 architecture)
- Following the string, there is a reference to the address of the string and its length.

Arker123:
During my examination of the binary, I came across 3 interesting types of strings:

1.
```
00000001400BD6E0  61 74 74 65 6D 70 74 20  74 6F 20 63 61 6C 63 75  attempt to calcu
00000001400BD6F0  6C 61 74 65 20 74 68 65  20 72 65 6D 61 69 6E 64  late the remaind
00000001400BD700  65 72 20 77 69 74 68 20  61 20 64 69 76 69 73 6F  er with a diviso
00000001400BD710  72 20 6F 66 20 7A 65 72  6F 2F 72 75 73 74 63 2F  r of zero/rustc/
00000001400BD720  38 34 63 38 39 38 64 36  35 61 64 66 32 66 33 39  84c898d65adf2f39
00000001400BD730  61 35 61 39 38 35 30 37  66 31 66 65 30 63 65 31  a5a98507f1fe0ce1
00000001400BD740  30 61 32 62 38 64 62 63  5C 6C 69 62 72 61 72 79  0a2b8dbc\library
00000001400BD750  5C 63 6F 72 65 5C 73 72  63 5C 73 74 72 5C 70 61  \core\src\str\pa
00000001400BD760  74 74 65 72 6E 2E 72 73  19 D7 0B 40 01 00 00 00  ttern.rs...@....
00000001400BD770  4F 00 00 00 00 00 00 00  D9 06 00 00 65 00 00 00  O...........e...
00000001400BD780  19 D7 0B 40 01 00 00 00  4F 00 00 00 00 00 00 00  ...@....O.......
```
In this type, there are 2 (or 1) strings. The reference to the first string is found in the .text section, while the reference to the second is immediately below it along with its string length.

2.
```
[...]

PR #836, opened 2023-07-17
Extract Rust specific strings from binaries #791

Arker123:
This pull request introduces a new feature to extract Rust specific strings from binaries.
Solves: #791


Arker123:
There are still several tasks pending, such as updating references from source code, comments, etc. Therefore, I'm marking this pull request as a draft for now.

mr-tz:
On first glance this appears to be identical or very close to the Go extract script. Can we just reuse that code or functions from it, so we don't repeat the same code twice?

Arker123:
> On first glance this appears to be identical or very close to the Go extract script. Can we just reuse that code or functions from it, so we don't repeat the same code twice?

Agreed! Let's consolidate the code and reuse the functions from the Go extract script in a separate PR after merging this one. We can create a file something like `language/utils.py` to house the shared functions. Thanks for the suggestion!

Arker123:
The current algorithm is based on pull request https://github.com/mandiant/flare-floss/issues/791#issuecomment-1654070181, and it's progressing well.
TODO: Extract references from the .text segment as well and improve code style and comments.

On a positive note, the algorithm appears to be splitting the strings quite effectively, as shown in the attached image.
image
image

Arker123:
TODO: Refactor comments and type hints

Arker123:
The current outcome appears promising.
image

mr-tz (changes requested):

floss/language/rust/extract.py
```diff
@@ -0,0 +1,156 @@
...
+
+    image_base = pe.OPTIONAL_HEADER.ImageBase
+
+    start_rdata, end_rdata, virtual_address, pointer_to_raw_data = get_rdata_section_info(pe)
+
+    # extract utf-8 strings
+    strings = list(b2s.extract_all_strings(buf[start_rdata:end_rdata], min_length))
+
+    # Filtering out strings that are not in .rdata
+    ref_data = filter_strings(strings, start_rdata, end_rdata)
```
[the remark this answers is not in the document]
mr-tz: which is done already, so is the filter needed?

floss/language/rust/extract.py
```diff
@@ -0,0 +1,156 @@
...
+        if not (start_rdata <= address < end_rdata):
+            continue
+
+        split_string(ref_data, address)
+
+    static_strings = []
+
+    for ref in ref_data:
+        try:
+            string = StaticString.from_utf8(ref[0].replace("\n", "").encode("utf-8"), ref[1], min_length)
```
mr-tz, line 127: need to be consistent here with handling the `\n` like we do elsewhere

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,105 @@
...
+            pointer_to_raw_data = section.PointerToRawData
+            section_size = section.SizeOfRawData
+            break
+
+    start_rdata = pointer_to_raw_data
+    end_rdata = pointer_to_raw_data + section_size
+
+    static_strings: Iterable[StaticString] = extract_ascii_unicode_strings(buf[start_rdata:end_rdata], args.min_length)
+
+    rust_strings = extract_utf8_strings(path, args.min_length)
```
mr-tz, line 72: don't we want the extracted rust strings here?

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,105 @@
...
+        large_chunk += string
+
+    total_static_string_length = len(large_chunk)
+
+    for rust_string in rust_strings:
+        # The current strings.py program doesn't take into account \n and \r
+        # so we have to remove them from rust_string to compare
+        string = rust_string.string.replace("\n", "").replace("\r", "")
+        if string not in large_chunk:
+            print(string.encode("utf-8"))
```
mr-tz, line 93: there shouldn't be extra output here, maybe use logger.debug() to show the data if needed

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,97 @@
...
+    end_rdata = pointer_to_raw_data + section_size
+
+    static_strings: Iterable[StaticString] = extract_ascii_unicode_strings(buf[start_rdata:end_rdata], args.min_length)
+
+    rust_strings = extract_utf8_strings(path, args.min_length)
+
+    get_extract_stats(static_strings, rust_strings)
+
+
+def get_extract_stats(static_strings: Iterable[StaticString], rust_strings: List[StaticString]) -> str:
```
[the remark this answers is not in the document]
mr-tz: seems to be resolved now

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,105 @@
...
+    total_static_string_length = 0
+    total_extracted_rust_string_length = 0
+    large_chunk = ""
+
+    for static_string in static_strings:
+        string = static_string.string
+        large_chunk += string
```
mr-tz, line 84: Concatenating all strings here and then checking the rust strings below seems cumbersome and could result in found strings which should actually be split. Can you tweak this to search/compare properly?

floss/language/rust/extract.py
```diff
@@ -0,0 +1,156 @@
...
+def get_rdata_section_info(pe: pefile.PE) -> Tuple[int, int, int, int]:
+    """
+    Retrieve info about .rdata section
+    """
+    for section in pe.sections:
+        if section.Name.startswith(b".rdata\x00"):
+            virtual_address = section.VirtualAddress
+            pointer_to_raw_data = section.PointerToRawData
+            section_size = section.SizeOfRawData
+            break
```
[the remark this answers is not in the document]
mr-tz: No, if we sys.exit here, this cannot be used easily as a library by other programs. Instead, raise a custom exception and catch where this is called.

floss/language/rust/extract.py
```diff
@@ -0,0 +1,158 @@
...
+    Extract UTF-8 strings from the given PE file using binary2strings
+    """
+
+    p = pathlib.Path(sample)
+    buf = p.read_bytes()
+    pe = pefile.PE(data=buf, fast_load=True)
+
+    image_base = pe.OPTIONAL_HEADER.ImageBase
+
+    # start_rdata, end_rdata, virtual_address, pointer_to_raw_data = get_rdata_section_info(pe)
```
mr-tz, line 90:
```suggestion
```

floss/language/rust/extract.py
```diff
@@ -0,0 +1,156 @@
...
+    # Get references from .text segment
+    xrefs = find_lea_xrefs(pe)
+
+    for xref in xrefs:
+        address = xref - image_base - virtual_address + pointer_to_raw_data
+
+        if not (start_rdata <= address < end_rdata):
+            continue
+
+        split_string(ref_data, address)
```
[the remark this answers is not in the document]
mr-tz: TODO

floss/language/rust/extract.py
```diff
@@ -0,0 +1,158 @@
...
+            rdata_structure = section
+            break
+    else:
+        logger.error("No .rdata section found")
+        sys.exit(1)
+
+    return rdata_structure
+
+
+def filter_strings_utf8(strings: List[Tuple[str, str, Tuple[int, int], bool]], start_rdata: int) -> List[Strings]:
```
mr-tz, line 43: this filters and returns modified data (start, end), so better change name or add a separate function?

mr-tz (changes requested):

floss/language/rust/coverage.py
mr-tz: as a reference, we have the Go coverage script, but I think you should think about the algorithm a bit first and suggest something

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,243 @@
...
+    )
+    logging_group = parser.add_argument_group("logging arguments")
+    logging_group.add_argument("-d", "--debug", action="store_true", help="enable debugging output on STDERR")
+    logging_group.add_argument(
+        "-q",
+        "--quiet",
+        action="store_true",
+        help="disable all status output except fatal errors",
+    )
+    arrs = parser.parse_arrs()
```
mr-tz, line 42: `parse_args`?

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,243 @@
...
+    pe = pefile.PE(data=buf, fast_load=True)
+
+    static_strings = list(extract_ascii_unicode_strings(buf, arrs.min_length))
+
+    rust_strings = extract_rust_strings(path, arrs.min_length)
+
+    get_extract_stats(pe, static_strings, rust_strings, arrs.min_length)
+
+
+def get_extract_stats(pe, all_ss_strings: List[StaticString], rust_strings, min_len) -> float:
```
mr-tz, line 70: if this is the same (or almost) as Go, can we reuse that and not duplicate all of the code?

floss/language/rust/extract.py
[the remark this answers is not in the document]
mr-tz: I'd imagine something like below in a block. Does that make sense?
```python
    substring = split_string(ref_data, address)
    string = StaticString.from_utf8(...)
    static_strings.append(string)
```

.github/mypy/mypy.ini
```diff
@@ -38,4 +38,7 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 [mypy-bs4.*]
-ignore_missing_imports = True
\ No newline at end of file
+ignore_missing_imports = True
+
```
mr-tz, line 42:
```suggestion
```

floss/language/rust/extract.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    rdata_structure = pefile.SectionStructure(pe.__IMAGE_SECTION_HEADER_format__)
+
+    for section in pe.sections:
+        if section.Name.startswith(b".rdata\x00"):
+            rdata_structure = section
+            break
+    else:
+        logger.error("No .rdata section found")
+        return rdata_structure
```
mr-tz, line 38: should this be returned here and can you double check the initialization in line 30 is what you want in this function?

floss/language/rust/extract.py
```diff
@@ -0,0 +1,164 @@
...
+
+    return rdata_structure
+
+
+def filter_and_transform_utf8_strings(
+    strings: List[Tuple[str, str, Tuple[int, int], bool]], start_rdata: int
+) -> List[Strings]:
+    """
+    Filter out strings that are not UTF-8
+    """
```
mr-tz, line 48:
```suggestion
```

floss/language/rust/coverage.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # see only .rdata section
+    buf = path.read_bytes()
+    pe = pefile.PE(data=buf, fast_load=True)
+
+    static_strings = list(extract_ascii_unicode_strings(buf, args.min_length))
+
+    rust_strings = extract_rust_strings(path, args.min_length)
+
+    get_extract_stats(pe, static_strings, rust_strings, args.min_length, 0)
```
mr-tz, line 64: does the 0 argument make sense here?

mr-tz:
> > I'd imagine something like below in a block. Does that make sense?
> 
> Sure thing! I get your point. Consider this: If the string is "CatDogHorse" and string pointers are at the end of each word, we'd have "CatDog" and "Horse" as splits. So, multiple splits are possible for a string. So, before adding to the static string, I think it's best to split all possible substrings. This way, we cover all potential splits. What do you think?

I'm not sure I follow. Do you have an example binary and output for this?

floss/language/rust/extract.py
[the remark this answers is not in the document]
mr-tz: Your comment in https://github.com/mandiant/flare-floss/pull/836#issuecomment-1684935367 doesn't clarify things for me really, but regardless, this more a comment on the design. The function can return a single string or if needed a list of strings that have been split. Does that make sense?

floss/language/rust/extract.py
```diff
@@ -56,17 +56,38 @@ def filter_and_transform_utf8_strings(
     return ref_data
 
 
-def split_string(ref_data: List[Strings], address: int) -> None:
+def split_string(ref_data: List[Strings], static_strings: List[StaticString], address: int) -> None:
```
mr-tz, line 59: how about returning `List[StaticString]` here?

floss/language/rust/extract.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # filter out strings that are not UTF-8 and transform them
+    ref_data = filter_and_transform_utf8_strings(strings, start_rdata)
+
+    # append all the ref_data strings to static_strings
+    for ref in ref_data:
+        try:
+            static_strings.append(StaticString.from_utf8(ref[0].encode("utf-8"), ref[1], min_length))
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
mr-tz: we shouldn't affect the coverage, but come up with a clean and easy to follow design

floss/language/rust/extract.py
```diff
@@ -0,0 +1,174 @@
...
+    Retrieve info about .rdata section
+    """
+    rdata_structure = None
+
+    for section in pe.sections:
+        if section.Name.startswith(b".rdata\x00"):
+            rdata_structure = section
+            break
+    else:
+        logger.error("No .rdata section found")
```
mr-tz, line 34: raise exception here and catch where called

floss/language/rust/extract.py
```diff
@@ -0,0 +1,174 @@
...
+    start_rdata = rdata_section.PointerToRawData
+    end_rdata = start_rdata + rdata_section.SizeOfRawData
+    virtual_address = rdata_section.VirtualAddress
+    pointer_to_raw_data = rdata_section.PointerToRawData
+
+    # extract utf-8 strings
+    strings = list(b2s.extract_all_strings(buf[start_rdata:end_rdata], min_length))
+
+    # filter out strings that are not UTF-8 and transform them
+    filter_and_transform_utf8_strings(strings, static_strings, start_rdata, min_length)
```
mr-tz, line 126: return value here

floss/results.py
```diff
@@ -130,18 +130,23 @@ class StaticString:
 
     string: str
     offset: int
-    encoding: StringEncoding
+    offset_end: int = 0
+    encoding: StringEncoding = StringEncoding.ASCII
 
     @classmethod
-    def from_utf8(cls, buf, addr, min_length):
+    def from_utf8(cls, buf, addr, min_length: int, end_addr: Optional[int] = None):
```
mr-tz, line 137: we can calculate this from addr + len(string)

floss/language/rust/extract.py
```diff
@@ -0,0 +1,173 @@
...
+    strings = list(b2s.extract_all_strings(buf[start_rdata:end_rdata], min_length))
+
+    # filter out strings that are not UTF-8 and transform them
+    static_strings = filter_and_transform_utf8_strings(strings, start_rdata, min_length)
+
+    # Get Struct string instances for .rdata section
+    candidates = get_struct_string_candidates(pe)
+
+    for can in candidates:
+        address = can.address - image_base - virtual_address + pointer_to_raw_data
```
mr-tz, line 131: this is the virtual address not the raw offset in the file

mr-tz: nevermind, was looking at the wrong file

mr-tz: Can you double-check the static strings FLOSS output lines up with this?

floss/language/rust/coverage.py
mr-tz: this is still an open question, we should add least add a comment with an explanation here

mr-tz (commented):
sorry, I think I didn't properly format my changes (in the PR to the fork)

floss/language/rust/extract.py
```diff
@@ -104,9 +105,11 @@ def get_string_blob_strings(pe: pefile.PE, min_length: int) -> Iterable[StaticSt
     static_strings = filter_and_transform_utf8_strings(strings, start_rdata)
 
     struct_string_addrs = map(lambda c: c.address, get_struct_string_candidates(pe))
-    xrefs = find_lea_xrefs(pe)
+    xrefs_lea = find_lea_xrefs(pe)
+    xrefs_push = find_push_xrefs(pe)
+    xrefs_mov = find_mov_xrefs(pe)
```
mr-tz, line 110: do we want to do all of these all the time (for both architectures or specific to 32/64 bit?)

floss/language/rust/extract.py
mr-tz, line 1: can you please share some current coverage output for 32 and 64 bit samples?

floss/language/rust/coverage.py
mr-tz: @Arker123 ^^

floss/language/rust/extract.py
mr-tz: can you repeat this for a few random samples and just share the stats here?

floss/language/utils.py
```diff
@@ -0,0 +1,623 @@
...
+    low, high = get_image_range(pe)
+
+    for section in pe.sections:
+        if not section.IMAGE_SCN_MEM_EXECUTE:
+            continue
+
+        code = section.get_data()
+
+        if pe.FILE_HEADER.Machine == pefile.MACHINE_TYPE["IMAGE_FILE_MACHINE_AMD64"]:
+            xrefs: Iterable[VA] = []  # no push instructions on amd64
```
mr-tz, line 201: can we be sure string references would never be pushed? probably, but can you add some context/references around this? the comment is a little misleading currently

floss/language/utils.py
```diff
@@ -0,0 +1,623 @@
...
+    low, high = get_image_range(pe)
+
+    for section in pe.sections:
+        if not section.IMAGE_SCN_MEM_EXECUTE:
+            continue
+
+        code = section.get_data()
+
+        if pe.FILE_HEADER.Machine == pefile.MACHINE_TYPE["IMAGE_FILE_MACHINE_AMD64"]:
+            xrefs: Iterable[VA] = []  # no mov instructions on amd64
```
mr-tz, line 255: same comment as above

floss/language/rust/extract.py
mr-tz: great, that looks pretty promising!

floss/language/utils.py
```diff
@@ -464,6 +464,8 @@ def get_extract_stats(
     pe: pefile, all_ss_strings: List[StaticString], lang_strings: List[StaticString], min_len: int, min_blob_len=0
 ) -> float:
     # min_blob_len: this is the minimum length of a string blob in binary file to be considered for extraction
+    # It is 0 in case of Rust extraction because we want to extract all strings from binary file
+    # while in case of Go extraction, we want to extract only large strings (len > 2800) from binary file
```
mr-tz, line 468: let's only leave the respective comments where this is called, so please remove here

Issue #788, opened 2023-06-05
ValueError: cannot mmap an empty file


mr-tz:
Run floss on empty file -> exception: `ValueError: cannot mmap an empty file`

This should be handled more gracefully.

PR #846, opened 2023-07-28
Handle Empty File Exception in FLOSS Execution

Arker123:
This PR addresses the issue where running FLOSS on an empty file resulted in an exception (ValueError: cannot mmap an empty file). 

Fixes: #788

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -540,7 +540,7 @@ def main(argv=None) -> int:
 
     static_strings = get_static_strings(sample, args.min_length)
     if static_strings == []:
-        return 1
+        return 0
 
     static_runtime = get_runtime_diff(interim)
 
--- a/floss/main.py
+++ b/floss/main.py
@@ -537,7 +537,11 @@ def main(argv=None) -> int:
     # can throw away result later if not desired in output
     time0 = time()
     interim = time0
+
     static_strings = get_static_strings(sample, args.min_length)
+    if static_strings == []:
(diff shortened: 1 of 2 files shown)


floss/main.py
```diff
@@ -537,7 +537,11 @@ def main(argv=None) -> int:
     # can throw away result later if not desired in output
     time0 = time()
     interim = time0
+
     static_strings = get_static_strings(sample, args.min_length)
+    if static_strings == []:
+        return 1
```
williballenthin, line 543: i don't think this should be treated as an error. both of us can trivially construct a non-empty file with 0 ASCII strings and i think FLOSS should be able to handle these. 

printing a logging message might be fine but let's not return an error status code.

floss/utils.py
```diff
@@ -541,6 +542,11 @@ def get_static_strings(sample: Path, min_length: int) -> list:
     """
     Returns list of static strings from the file which are above the minimum length
     """
+
+    if os.stat(sample).st_size == 0:
+        logger.error("File is empty")
```
williballenthin, line 547: i think this should be a warning at most, not an error (something that cannot be recovered).

williballenthin (changes requested):

williballenthin (changes requested):
one small tweak, then ready to merge

williballenthin (approved):
thanks!

Issue #839, opened 2023-07-18
 remove support for python 3.7


mr-tz:
new min: 3.8

e.g. blocker on https://github.com/mandiant/flare-floss/pull/835

PR #850, opened 2023-08-01
Remove Python 3.7 Support

Arker123:
This PR removes support for Python 3.7.

Fixes: #839

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,8 +43,6 @@ jobs:
         include:
           # on Ubuntu run these as well
           - os: ubuntu-20.04
-            python-version: '3.8'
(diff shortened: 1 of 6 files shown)


Arker123:
Is there anything else left to remove or any further tasks I should handle?

.github/workflows/tests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         os: [ubuntu-20.04, windows-2019, macos-11]
         # across all operating systems
         # version numbers are strings, not floats
-        python-version: ['3.7', '3.11']
+        python-version: ['3.8', '3.11']
```
williballenthin, line 42: would you remove the 3.8 reference in the matrix below?

Issue #854, opened 2023-08-05
CI: Add testcases for known binary files (Rust)


Arker123:
Add testcases to CI for known expected strings for all tracked Rust versions

PR #855, opened 2023-08-05
Test the detection of known Rust binaries

Arker123:
This PR includes tests to verify the detection of known strings in various versions of Rust binaries, which are extracted from zip archives.

Solves: #854

### DIFF
--- a/tests/test_language_rust_known_binary.py
+++ b/tests/test_language_rust_known_binary.py
@@ -1,0 +1,166 @@
+import pathlib
+import zipfile
+
+import pytest
+
+from floss.language.rust.extract import extract_utf8_strings
+
+
+@pytest.fixture(scope="module")
+def extract_files_64():
+    zip_file_path = (
+        pathlib.Path(__file__).parent
+        / "data"
+        / "language"
+        / "rust"
+        / "rust-binaries-all-versions"
+        / "bin"
+        / "versions_64.zip"
+    )
+
+    CD = pathlib.Path(__file__).resolve().parent
+
+    abs_zip_path = (CD / zip_file_path).resolve()
+    # check if the zip file exists
+    assert abs_zip_path.exists(), f"Zip file {zip_file_path} does not exist"
+
+    extracted_files = []
+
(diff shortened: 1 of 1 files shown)


tests/test_language_rust_known_binary.py
```diff
@@ -0,0 +1,167 @@
...
+        / "rust"
+        / "rust-binaries-all-versions"
+        / "bin"
+        / "versions_64.zip"
+    )
+
+    CD = pathlib.Path(__file__).resolve().parent
+
+    abs_zip_path = (CD / zip_file_path).resolve()
+    # check if the zip file exists
```
mr-tz, line 24:
```suggestion
```

tests/test_language_rust_known_binary.py
```diff
@@ -0,0 +1,167 @@
...
+    yield
+
+    # Clean up - remove the extracted files after the test finishes
+    for extracted_file in extracted_files:
+        pathlib.Path(extracted_file).unlink()
+    pathlib.Path(extracted_file_path.parent).rmdir()
+
+
+@pytest.fixture(scope="module")
+def extract_files_32():
```
mr-tz, line 53: can you refactor to use one function for unzipping vs. repeating the code

tests/test_language_rust_known_binary.py
```diff
@@ -0,0 +1,167 @@
...
+                / "rust"
+                / "rust-binaries-all-versions"
+                / "bin"
+                / "extracted_64"
+                / zip_info.filename
+            ).resolve()
+            extracted_file = zip_ref.extract(zip_info, path=extracted_file_path.parent)
+            extracted_files.append(extracted_file)
+
+    yield
```
mr-tz, line 44: what's the yield for here?

tests/test_language_rust_known_binary.py
mr-tz: I thought we even had this before 😕

mr-tz (approved):
LGTM, thanks!

Issue #826, opened 2023-07-12
Go strings integration tasks


mr-tz:
```[tasklist]
### Tasks
- [x] render `\t` vs. tab (strings.exe does the latter)
- [x] Go build ID string split into multiple lines, see image1?
- [x] strings split, see image 2?
```

image 1
<img width="626" alt="2023-07-12_12-24-33_WindowsTerminal" src="https://github.com/mandiant/flare-floss/assets/17606537/43711a7d-07ee-463f-ac8e-61bf0bdb19fe">

image 2
<img width="1546" alt="2023-07-12_11-29-15_WinMergeU" src="https://github.com/mandiant/flare-floss/assets/17606537/c61df6ef-f3b2-4363-8306-1d0e71e92840">

Arker123:
> For task 2 and 3 I think the `console` setting (width) results in the added new lines...

So, to address this, should I increase the width setting?

williballenthin:
if the case is what @mr-tz mentioned, then its `rich` deciding to wrap the string across multiple lines, based on the current width of the terminal. we should decide what the behavior should be. i think two reasonable options include:

  - don't wrap, and truncate strings at the end of the terminal, though this potentially makes it difficult to review the complete string contents
  - don't wrap, and let the terminal wrap long lines, as strings.exe does. i dont think blindly changing the console width is the right fix here, since i can trivially construct a string that's longer than whatever width you pick. instead, we should read more about `rich` and how to change the behavior.

mr-tz:
> > render `\t` vs. tab (strings.exe does the latter)
> 
> Is this feature already in place?
> 
> ![image](https://user-images.githubusercontent.com/94680887/261846893-e22b4b40-46ff-4551-acbc-78c3446508e2.png)

we need to decide which behavior we want and make sure it's consistent across the output functions

is that the case currently?

mr-tz:
Ok, then we only have the wrapping issue left to fix.

mr-tz:
closed via #861

PR #861, opened 2023-08-20
Utilize rich "soft_wrap" feature for Improved Display

Arker123:
We can leverage the `rich` library's `soft_wrap` feature, which neatly handles line breaks while maintaining the string's integrity.

Reference:- https://github.com/pycontribs/enrich

Fixes:- #826

### DIFF
--- a/floss/render/default.py
+++ b/floss/render/default.py
@@ -9,7 +9,7 @@ from typing import Dict, List, Tuple, Union
 from rich import box
 from rich.table import Table
 from rich.markup import escape
-from enrich.console import Console
+from rich.console import Console
 
 import floss.utils as util
 import floss.logging_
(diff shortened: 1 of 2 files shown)


mr-tz:
Cool, can you share a before and after output? Also black is not happy here, so please address.

floss/render/default.py
```diff
@@ -301,7 +301,7 @@ def get_color(color):
 
 def render(results: floss.results.ResultDocument, verbose, disable_headers, color):
     sys.__stdout__.reconfigure(encoding="utf-8")
-    console = Console(file=io.StringIO(), color_system=get_color(color), highlight=False)
+    console = Console(file=io.StringIO(), color_system=get_color(color), highlight=False, redirect=False)
```
williballenthin, line 304: should this be using `soft_wrap=True`? i don't see how this change affects the wrapping.

mr-tz: So we could also achieve this with `rich` and setting `soft_wrap=True`?

Issue #860, opened 2023-08-17
Integrate Rust string extraction into FLOSS


Arker123:
This issue proposes the integration of Rust based string extraction into the FLOSS.

PR #863, opened 2023-08-23
Integrate Rust specific extraction into FLOSS

Arker123:
This PR integrates the Rust based string extraction into the FLOSS.

Fixes: #860

### DIFF
--- a/floss/language/identify.py
+++ b/floss/language/identify.py
@@ -26,6 +26,7 @@ def identify_language(sample: Path, static_strings: Iterable[StaticString]) -> L
     Identify the language of the binary given
     """
     if is_rust_bin(static_strings):
+        logger.warning("Presently, the decoding analysis is fine-tuned for static strings.")
         return Language.RUST
 
     # Open the file as PE for further checks
--- a/floss/language/identify.py
+++ b/floss/language/identify.py
@@ -26,7 +26,7 @@ def identify_language(sample: Path, static_strings: Iterable[StaticString]) -> L
     Identify the language of the binary given
     """
     if is_rust_bin(static_strings):
-        logger.warning("Presently, the decoding analysis is fine-tuned for static strings.")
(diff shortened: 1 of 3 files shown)


floss/language/identify.py
```diff
@@ -26,8 +26,6 @@ def identify_language(sample: Path, static_strings: Iterable[StaticString]) -> L
     Identify the language of the binary given
     """
     if is_rust_bin(static_strings):
-        logger.warning("Rust binary detected, Rust binaries are not supported yet; results may be inaccurate")
-        logger.warning("Rust: proceeding with analysis may take a long time")
```
mr-tz, line 30: Should we keep these messages around for the string decoding part? We only handle static strings better now.

floss/render/default.py
```diff
@@ -143,9 +143,9 @@ def strtime(seconds):
     return f"{m:02.0f}:{s:02.0f}"
 
 
-def render_gostrings(language_strings, language_strings_missed, console, verbose, disable_headers):
+def render_language_strings(language, language_strings, language_strings_missed, console, verbose, disable_headers):
```
mr-tz, line 146: cool

mr-tz (approved):
looks good, only one question on keeping the warning around (although there's one already), so curious to hear your thoughts

floss/language/identify.py
```diff
@@ -26,8 +26,7 @@ def identify_language(sample: Path, static_strings: Iterable[StaticString]) -> L
...
+        logger.warning("Presently, the decoding analysis is fine-tuned for static strings.")
```
mr-tz, line 29:
```suggestion
        logger.warning("Rust binary detected, FLOSS handles Rust static strings, but string deobfuscation may be inaccurate and take a long time")
```

mr-tz: should we add the same for Go?

PR #871, opened 2023-08-29
handle failures on Go 1.12 samples

mr-tz:


floss/language/utils.py
```diff
@@ -403,6 +403,7 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
         if not section.IMAGE_SCN_MEM_READ:
             continue
 
+        # TODO add .text here for Go version 1.12?
```
williballenthin, line 406: would you add an issue to the tracker so we can be sure not to forget about this and/or solicit help from external contributors?

williballenthin (approved):

floss/language/utils.py
mr-tz: done #872

Issue #867, opened 2023-08-28
Improve b2s or replace with custom code (unneded wide strings support)


mr-tz:
binary2strings reads utf8 and wide strings (we don't expect or require the latter for Rust strings extraction).

In `rust-hello64.exe` we see
<img width="366" alt="2023-08-28_15-09-13_ida64" src="https://github.com/mandiant/flare-floss/assets/17606537/268829fe-fd3a-49a0-b607-cd5516859dd2">

Which parsed as wide string starting at offset 0x1400BD04F, b2s sees as:
<img width="173" alt="2023-08-28_15-11-07_pycharm64" src="https://github.com/mandiant/flare-floss/assets/17606537/a4e0aa87-39b6-4b6c-914f-d49daa7bed9d">

We then filter out the wide string and end up with ` args` in the final Rust strings output.

mr-tz:
Yes, currently we only require a function to find UTF-8 strings though.

vaijavi:
Hi maintainers,
I would like to work on this issue. Please let me know if I can take it.
Thanks!

mr-tz:
Go ahead

svm1048:
hello @mr-tz! I do realise that vaijavi wants to work on this bug, i would like to offer my help as well so that it dosent come off as "sniping". Is it okay if i start working on this ?

mr-tz:
I think you can go ahead.

svm1048:
PR #1223
take a look, also I haven't been covered under CLA, I have applied will lyk how that goes!

kunalsz:
@mr-tz not snipping this issue, just some research I did. The regex based UTF8 strings extraction like in #1223 :
- itself encodes UTF8 validity rules 
- allows ASCII printable + \t\r\n unconditionally (as part of the regex)
- finds any match anywher; no explicit start filtering needed
- no explicit maximum match size; in worst-case it can match and allocate/decode a very large string in one go

It is simple and fast but I made a decoder using python’s incremental UTF‑8 decoder as the authority on validity which addresses these shortcomings (and doesn't use hardcoded regex)

Here are the results(after|before):
<img width="857" height="313" alt="Image" src="https://github.com/user-attachments/assets/6c0653aa-e2f2-4f2c-a5c9-bd1ac09f1089" />
There's a slight slowdown cause b2s used c++, while this is a pure python implementation but it does the job pretty well
(before|after)
[...]

PR #885, opened 2023-09-26
Implement Custom UTF-8 Decoder

Arker123:
This PR introduces a custom UTF-8 string decoder to replace the existing binary2strings conversion.
Reference: https://en.wikipedia.org/wiki/UTF-8
Fixes: #867

### DIFF
--- a/floss/language/rust/decode_utf8.py
+++ b/floss/language/rust/decode_utf8.py
@@ -1,0 +1,105 @@
+# Copyright (C) 2023 Mandiant, Inc. All Rights Reserved.
+import pefile
+import logging
+import argparse
+import pathlib
+import sys
+
+MIN_STR_LEN = 4
+
+logger = logging.getLogger(__name__)
+
+def get_rdata_section(pe: pefile.PE) -> pefile.SectionStructure:
+    for section in pe.sections:
+        if section.Name.startswith(b".rdata\x00"):
+            return section
+
+    raise ValueError("no .rdata section found")
+
+def extract_utf8_strings(pe, min_length=MIN_STR_LEN):
+    try:
+        rdata_section = get_rdata_section(pe)
+    except ValueError as e:
+        print("cannot extract rust strings: %s", e)
+        return []
+
+    strings = rdata_section.get_data()
+
+    character_and_index = []
+
+    # Reference: https://en.wikipedia.org/wiki/UTF-8 
+
+    for i in range(0, len(strings)):
+        # for 1 byte
+        if strings[i] & 0x80 == 0x00:
+            character = strings[i].to_bytes(1, "big").decode("utf-8", "ignore")
+            character_and_index.append([character, i, 1])
+
+        # for 2 bytes
+        elif strings[i] & 0xE0 == 0xC0:
+            temp = strings[i] << 8 | strings[i+1]
+            character = temp.to_bytes(2, "big").decode("utf-8", "ignore")
+            i += 1
+            character_and_index.append([character, i, 2])
+
+        # for 3 bytes
+        elif strings[i] & 0xF0 == 0xE0:
+            temp = strings[i] << 16 | strings[i+1] << 8 | strings[i+2]
+            character = temp.to_bytes(3, "big").decode("utf-8", "ignore")
+            i += 2
+            character_and_index.append([character, i, 3])
+
+        # for 4 bytes
+        elif strings[i] & 0xF8 == 0xF0:
+            temp = strings[i] << 24 | strings[i+1] << 16 | strings[i+2] << 8 | strings[i+3]
+            character = temp.to_bytes(4, "big").decode("utf-8", "ignore")
+            i += 3
+            character_and_index.append([character, i, 4])
+
+
+    strings = [] # string, start index, end index
+
+    # check for consecutive characters and convert to string
+    for i in range(0, len(character_and_index)):
+        if i == 0:
+            strings.append([character_and_index[i][0], character_and_index[i][1], character_and_index[i][1]])
+        else:
+            if character_and_index[i-1][1] + character_and_index[i-1][2] == character_and_index[i][1] and character_and_index[i][0].isprintable() == True:
+                strings[-1][0] += character_and_index[i][0]
+                strings[-1][2] = character_and_index[i][1]
+            else:
+                if character_and_index[i][0].isprintable() == True:
+                    strings.append([character_and_index[i][0], character_and_index[i][1], character_and_index[i][1]])
+
+    # filter strings less than min length
+    strings = [string for string in strings if len(string[0]) >= min_length]
+
+    return strings
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser(description="Get Rust strings")
+    parser.add_argument("path", help="file or path to analyze")
+    parser.add_argument(
+        "-n",
+        "--minimum-length",
+        dest="min_length",
+        type=int,
+        default=MIN_STR_LEN,
+        help="minimum string length",
+    )
+    args = parser.parse_args(args=argv)
+
+    logging.basicConfig(level=logging.DEBUG)
+
+    pe = pathlib.Path(args.path)
+    buf = pe.read_bytes()
+    pe = pefile.PE(data=buf, fast_load=True)
+
+    strings = extract_utf8_strings(pe, args.min_length)
+    for string in strings:
+        print(string[0])
+
+
+if __name__ == "__main__":
+    sys.exit(main())


williballenthin:
@Arker123 would you include some performance benchmarks as you propose these changes? eg. does FLOSS performance noticeably change when using this implementation?

don't have to address this immediately, just before we consider it for merging

Arker123:
For rust-hello64.exe, the current coverage has dropped from 96% to 71%, which I'm investigating. That's the reason this PR is marked as a draft. On the plus side, it correctly extracts the string 'invalid args,' as shown in issue #867.

Arker123:
Additionally, we still need to address the consideration of 'characters' in this implementation.
image
Source: https://en.wikipedia.org/wiki/UTF-8

Arker123:
Presently, we have achieved coverage levels of >= 96% in 5 out of 6 test files. However, one file, specifically https://github.com/mandiant/flare-floss-testfiles/blob/master/language/rust/rust-unknown-binaries/bin/1.69.0/i386/200c308a793630e4f3686dd846f0d55b6368834a859875970b4135f3ca487f46, is currently only at 86% coverage. :(

Arker123:
After conducting further research, I have discovered additional xrefs within the .rdata section of the i386 architecture. The current coverage for the aforementioned file stands at 91%.

You can view the findings here:
Annotation 2023-10-02 051352

Annotation 2023-10-02 051425

mr-tz:
Let's not mix updates to the extraction algorithm and the custom UTF-8 string finder in this PR.

For performance here, let's focus on the time it takes to find just all the UTF-8 strings from files.

Arker123:
Here are some performance statistics:
-------------------------------------------------------

File name:  1.59.0_amd64_4d41c02164ee4aeacd5e3fd67ecdc097233b8c6a725fab588d1cdde0461fd8cc
File size:  114.84571838378906 MB
perf:  8.53998875617981 s

-------------------------------------------------------

File name:  1.49.0_i386_13af976c769e583e63665cc29ddeb6dc1a81c447ce7ea87bc10e01941576318b
File size:  112.57718658447266 MB
perf:  11.280088901519775 s

-------------------------------------------------------

File name:  1.67.1_amd64_286164c8db3976443ca605209c182e93d2fee5854504d5e9fc32a0b701b78aa8
File size:  125.58065795898438 MB
perf:  20.27586317062378 s

-------------------------------------------------------

File name:  1.53.0_amd64_6b4292500c9603d0f9647ae53adaa4cbf2840ae243d9efa3f14048f079e2d32a
File size:  65.56964874267578 MB
perf:  4.034202814102173 s

-------------------------------------------------------

File name:  1.53.0_amd64_63fa08f342be3c2a9c13f3d4c02f5a808913b291d0dc2691424fbb6e50fadb0d
File size:  65.56991577148438 MB
perf:  4.46514892578125 s

-------------------------------------------------------

File name:  1.65.0_amd64_6f0a952ebd0fd544bf27a13686a0f3494e9be102654a58996bedfe3bcc6f61ab
File size:  83.82809448242188 MB
perf:  9.418390274047852 s

-------------------------------------------------------

File name:  1.53.0_amd64_bcca0fd4d94764ef5a1ade5e81e749b11d87ab1c6c490d9065040e20da9dcf73
File size:  65.56964874267578 MB
perf:  4.185543060302734 s

-------------------------------------------------------

File name:  1.59.0_amd64_a09acc46cc64b945ba7e5b1fcd449b6cad435b31d45c7cdc9a4b8dc3c5d4aa04
File size:  115.15755462646484 MB
perf:  8.811774015426636 s

-------------------------------------------------------

File name:  1.53.0_amd64_15d603b12906f96d4891164febef7b587667363b7c91011ee100c81cad3a0472
File size:  65.56964874267578 MB
perf:  4.438174724578857 s

-------------------------------------------------------

File name:  1.64.0_amd64_8b4a259da875e73e61b72c87191b8e7ba35c29da6db4fc1f7805c5f5ef93d42e
File size:  74.84070587158203 MB
perf:  5.529693603515625 s

-------------------------------------------------------

File name:  1.69.0_amd64_c5f8c0a07123fc97c7b673b22b5f54cf415d66454022827f849a76dbe7ab2853
File size:  17.7099609375 MB
perf:  2.7266366481781006 s

-------------------------------------------------------

File name:  1.69.0_amd64_0eba113dde8ba85afd96116d39444fc4fcc7a280b45d0d15544642f290a5e72a
File size:  31.06982421875 MB
perf:  0.1291041374206543 s

-------------------------------------------------------

File name:  1.67.1_amd64_a585d992715c9d482fea1d047ea938d884f2888fa5d037aeca8c14ee6f7636d1
File size:  125.58039093017578 MB
perf:  19.494038581848145 s

-------------------------------------------------------

mr-tz:
Are these the performance stats just for running the UTF-8 extraction code?

Can you clean up the PR to only include those changes, please?

Arker123:
> Are these the performance stats just for running the UTF-8 extraction code?

Yes

> Can you clean up the PR to only include those changes, please?

Okay😄

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,112 @@
...
+
+def get_rdata_section(pe: pefile.PE) -> pefile.SectionStructure:
+    for section in pe.sections:
+        if section.Name.startswith(b".rdata\x00"):
+            return section
+
+    raise ValueError("no .rdata section found")
+
+
+def extract_utf8_strings(pe: pefile.PE, min_length=MIN_STR_LEN) -> List[Tuple[str, int, int]]:
```
mr-tz, line 23: can you add a function that's signature is analog to `extract_ascii_strings`, i.e. operate on buffers?
this can then be wrapped by this or a similar function.

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,112 @@
...
+
+    strings = rdata_section.get_data()
+
+    character_and_index = []
+
+    # Reference: https://en.wikipedia.org/wiki/UTF-8
+
+    for i in range(0, len(strings)):
+        # for 1 byte
+        if strings[i] & 0x80 == 0x00:
```
mr-tz, line 41: please add tests for various encoded strings to show they're properly decoded

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def extract_utf8_strings(pe: pefile.PE, min_length=MIN_STR_LEN) -> List[List[Any]]:
+    """
+    Extracts UTF-8 strings from the .rdata section of a PE file.
+    """
+    try:
+        rdata_section = get_rdata_section(pe)
+    except ValueError as e:
+        print("cannot extract rust strings: %s", e)
```
VascoSch92, line 86: you can log that instead of printing it

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+            prev = False
+
+    # filter strings less than min length
+    strings = [string for string in strings if len(string[0]) >= min_length]
+
+    return strings
+
+
+def extract_utf8_strings(pe: pefile.PE, min_length=MIN_STR_LEN) -> List[List[Any]]:
```
VascoSch92, line 79:
```suggestion
def extract_utf8_strings(pe: pefile.PE, min_length=MIN_STR_LEN) -> List[Optional[List[Any]]]:
```
the return type is wrong

williballenthin (commented):
how does this impact performance?

floss/language/rust/decode_utf8.py
Arker123: It doesn't pass in mypy like this. Is there another approach?

Arker123:
> how does this impact performance?

There is an increase in extraction percentage from 88% to 91%.

here is a table for time taken to run tests for rust

| Test_name | original_time | current_time |
|--------|--------|--------|
| test_language_rust_coverage.py | 11.146s | 11.637s |
| test_language_rust_known_binary.py  | 36.969s | 34.953s |
| test_language_extract_rust.py | 4.321s | 4.472s |
| test_utf8_decoder.py | 12.366s | 13.194s | 
</table>

*Tests are ran on Core i5 12Gen

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def get_rdata_section(pe: pefile.PE) -> pefile.SectionStructure:
+    for section in pe.sections:
+        if section.Name.startswith(b".rdata\x00"):
+            return section
+
+    raise ValueError("no .rdata section found")
```
williballenthin, line 20: duplicate of function in `utils`?

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def extract_utf8_strings_from_buffer(buf, min_length=MIN_STR_LEN) -> List[List[Any]]:
```
williballenthin, line 23: Please use a return type hint that is more specific than `Any`. I'd expect this routine to return `List[str]` or `List[Tuple[int, str]]`. Please fix the hint and add documentation explaining the return type.

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+    # Reference: https://en.wikipedia.org/wiki/UTF-8
+
+    character_and_index = []
+    strings = []
+
+    for i in range(0, len(buf)):
+        # for 1 byte
+        if buf[i] & 0x80 == 0x00:
+            character = buf[i].to_bytes(1, "big").decode("utf-8", "ignore")
+            character_and_index.append([character, i, 1])
```
williballenthin, line 37: Recommend using a namedtuple or dataclass to capture the character/index/length information in a structured way. Using a list with implicit indices is difficult to follow.

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+            character = temp.to_bytes(3, "big").decode("utf-8", "ignore")
+            i += 2
+            character_and_index.append([character, i, 3])
+
+        # for 4 bytes
+        elif buf[i] & 0xF8 == 0xF0:
+            temp = buf[i] << 24 | buf[i + 1] << 16 | buf[i + 2] << 8 | buf[i + 3]
+            character = temp.to_bytes(4, "big").decode("utf-8", "ignore")
+            i += 3
+            character_and_index.append([character, i, 4])
```
williballenthin, line 58: what about the other cases? are there any?

either way, please add an `else` branch to handle them, either logging or doing an assertion.

floss/language/rust/decode_utf8.py
```diff
@@ -0,0 +1,121 @@
...
+    for i in range(0, len(buf)):
+        # for 1 byte
+        if buf[i] & 0x80 == 0x00:
+            character = buf[i].to_bytes(1, "big").decode("utf-8", "ignore")
+            character_and_index.append([character, i, 1])
+
+        # for 2 bytes
+        elif buf[i] & 0xE0 == 0xC0:
+            temp = buf[i] << 8 | buf[i + 1]
+            character = temp.to_bytes(2, "big").decode("utf-8", "ignore")
```
williballenthin, line 42: why do you use `ignore` here? wouldn't we want to handle the case that invalid UTF-8 data is encountered (and not extract a string there)?

I assume that your algorithm works pretty well, since you've opened the PR, but I can't quite follow how it works. Would you please add some comments explaining the design, and definitely a few test cases that exercise each of the branch arms?

floss/language/rust/decode_utf8.py
williballenthin, line 79:
```suggestion
def extract_rdata_utf8_strings(pe: pefile.PE, min_length=MIN_STR_LEN) -> List[List[Any]]:
```

tests/test_language_rust_coverage.py
```diff
@@ -53,5 +53,5 @@ def test_language_detection_64(binary_file):
     with contextlib.redirect_stdout(None):
         out = get_extract_stats(pe, all_ss_strings, rust_strings, n)
 
-    # check that the output percentage is greater than 88%
-    assert float(out) > 88
+    # check that the output percentage is greater than 86%
+    assert float(out) > 86  # increase to 91 after merging PR #899
```
williballenthin, line 57:
```suggestion
    assert float(out) > 86  # TODO(Arker123): increase to 91 after merging PR #899
```

williballenthin (changes requested):
Please see inline comments.

I think the core algorithm probably works well, but some enhancements to the structure and documentation will ensure it remains maintainable. I like that we'll be able to remove the external dependency!

floss/language/rust/decode_utf8.py
Arker123: Hi, the tests for each branch are in `tests/test_utf8_decoder.py`. Let me know if anything else is required.

mr-tz:
closing stale PR

PR #886, opened 2023-09-27
Feature added: file output arg

aniko33:
I redirected the rendering output to a `open()` statement in order to write the rendered output to a file. I introduced a new command-line argument to the program, as shown below:

```py
parser.add_argument(
    "-o",
    "--output",
    type=argparse.FileType("w"),
    help="specify an output file",
)
```

Furthermore, I implemented logging to track the process of saving the file:

```py
logger.info("saving to '{}'".format(args.output.name))
```

I have conducted testing, and the functionality works flawlessly.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -2,6 +2,7 @@
 # Copyright (C) 2017 Mandiant, Inc. All Rights Reserved.
 import os
 import sys
+import json
 import codecs
 import logging
 import argparse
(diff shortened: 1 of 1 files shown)


aniko33:
> would you please explain the pros/cons of this vs redirecting stdout to a file? we'd prefer not to maintain two ways to do the same thing unless there's a good reason.

If you intend to redirect stdout like this: `floss malware.exe > output.txt`, the pros and cons to consider:

**Pro**:

1. **Clean output**: Provides clean output, excluding Floss logs. It is especially useful if you intend to save the output in JSON format using the `-j` argument.

2. **Code customization**: You can easily customize the `--output` argument in your code, making it a more flexible option than saving stdout.

**Cons**:

1. **Speed**: The main disadvantage is the speed of saving output. It is considerably slower to save output using this redirection method in Python than to use stdout directly.

Implementation of this function should not pose a significant difficulty. It is only a few lines of code and its modification is simple, since it simply redirects the rendered output to a file.

mr-tz:
We had this option before, but opted to let users redirect the output to a file in the terminal instead. What's your use-case here?
I could see this being useful when using FLOSS as a library potentially, but don't think we would want to add it for this as of now.

mr-tz:
Thanks for the suggestion and feedback. Please let us know if you have other ideas to make the tool more useful and easier to use. Happy to discuss :)

PR #898, opened 2023-11-09
Simplify language identification using dictionary mapping

Arker123:
This PR suggests using a dictionary mapping approach for language identification.

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -548,16 +548,14 @@ def main(argv=None) -> int:
 
     # set language configurations
     lang_id: Language
-    if args.language == Language.GO.value:
-        lang_id = Language.GO
-    elif args.language == Language.RUST.value:
-        lang_id = Language.RUST
-    elif args.language == Language.DOTNET.value:
-        lang_id = Language.DOTNET
-    elif args.language == "none":
-        lang_id = Language.UNKNOWN
-    else:
-        lang_id = identify_language(sample, static_strings)
+    lang_id_mapping = {
+        Language.GO.value: Language.GO,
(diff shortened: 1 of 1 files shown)


williballenthin (commented):
can't we just do `Language(args.language)`?

(need a try/except to handle missing cases)

williballenthin:
while we're at it, we may want to derive the argparse choices from the Language enum.

floss/main.py
```diff
@@ -198,8 +198,8 @@ def make_parser(argv):
     advanced_group.add_argument(
         "--language",
         type=str,
-        choices=[l.value for l in Language if l != Language.UNKNOWN] + ["none"],
-        default="",
+        choices=[lang.value for lang in Language],
+        default="none",
```
mr-tz, line 202: does this include the unknown language in the usage output and allow to use "none"?

Arker123: Yes, this change includes the unknown. I've updated it to allow "none" and not "Unknown" in latest commit 😄

floss/main.py
```diff
@@ -548,15 +548,10 @@ def main(argv=None) -> int:
...
-    elif args.language == Language.RUST.value:
-        lang_id = Language.RUST
-    elif args.language == Language.DOTNET.value:
-        lang_id = Language.DOTNET
-    elif args.language == "none":
-        lang_id = Language.UNKNOWN
-    else:
+
+    try:
+        lang_id = Language(args.language)
```
mr-tz, line 553: how can users disable language detection now?

mr-tz:
now part of #901

PR #899, opened 2023-11-09
More i386 xrefs

Arker123:
Referring to https://github.com/mandiant/flare-floss/pull/885#issuecomment-1742411052, this PR supplements additional xrefs discovered separately. For #885, the focus is solely on the UTF-decoder segment. 😄

### DIFF
--- a/floss/language/utils.py
+++ b/floss/language/utils.py
@@ -465,6 +465,34 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
             # dozens of seconds or more (suspect many minutes).
 
 
+def get_raw_xrefs_rdata_i386(pe: pefile.PE, buf: bytes) -> Iterable[VA]:
+    """
+    scan for raw xrefs in .rdata section
+    """
+    format = "I"
+
+    if not buf:
+        return
+
+    low, high = get_image_range(pe)
+
+    # using array module as a high-performance way to access the data as fixed-sized words.
+    words = iter(array.array(format, buf))
+
+    last = next(words)
+    for current in words:
+        address = last
+        last = current
+
+        if address == 0x0:
+            continue
+
+        if not (low <= address < high):
+            continue
+
+        yield address
+
+
 def get_extract_stats(
     pe: pefile, all_ss_strings: List[StaticString], lang_strings: List[StaticString], min_len: int, min_blob_len=0
 ) -> float:
--- a/floss/language/rust/extract.py
+++ b/floss/language/rust/extract.py
@@ -10,7 +10,13 @@ import pefile
 import binary2strings as b2s
 
 from floss.results import StaticString, StringEncoding
-from floss.language.utils import find_lea_xrefs, find_mov_xrefs, find_push_xrefs, get_struct_string_candidates
+from floss.language.utils import (
+    find_lea_xrefs,
+    find_mov_xrefs,
+    find_push_xrefs,
+    get_raw_xrefs_rdata_i386,
+    get_struct_string_candidates,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -151,7 +157,8 @@ def get_string_blob_strings(pe: pefile.PE, min_length: int) -> Iterable[StaticSt
         xrefs_lea = find_lea_xrefs(pe)
         xrefs_push = find_push_xrefs(pe)
         xrefs_mov = find_mov_xrefs(pe)
-        xrefs = itertools.chain(struct_string_addrs, xrefs_lea, xrefs_push, xrefs_mov)
+        xrefs_raw_rdata = get_raw_xrefs_rdata_i386(pe, rdata_section.get_data())
+        xrefs = itertools.chain(struct_string_addrs, xrefs_lea, xrefs_push, xrefs_mov, xrefs_raw_rdata)
 
     elif pe.FILE_HEADER.Machine == pefile.MACHINE_TYPE["IMAGE_FILE_MACHINE_AMD64"]:
         xrefs_lea = find_lea_xrefs(pe)


floss/language/utils.py
```diff
@@ -465,6 +465,34 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
             # dozens of seconds or more (suspect many minutes).
 
 
+def get_raw_xrefs_rdata_i386(pe: pefile.PE, buf: bytes) -> Iterable[VA]:
+    """
+    scan for raw xrefs in .rdata section
```
mr-tz, line 470: what are raw xrefs? can you add an example disassembly listing and add some comments, please?

mr-tz: from the screenshots in https://github.com/mandiant/flare-floss/pull/885#issuecomment-1742411052 I don't see if these are strings and it would help to have some comments explaining what the algorithm looks for

Arker123: Raw xrefs refer to unprocessed xrefs in the binary file, indicating points where strings can be divided. I'll add an example with comments.

mr-tz: Thanks, can you share a few example binary hashes?

Arker123: https://github.com/mandiant/flare-floss-testfiles/blob/master/language/rust/rust-hello/bin/rust-hello.exe

floss/language/utils.py
```diff
@@ -465,6 +465,48 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
             # dozens of seconds or more (suspect many minutes).
 
 
+def get_raw_xrefs_rdata_i386(pe: pefile.PE, buf: bytes) -> Iterable[VA]:
```
mr-tz, line 468: please add a few tests for these strings

floss/language/utils.py
```diff
@@ -473,6 +473,48 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
...
+    They are not encoded as struct String instances.
+
+    example:
+    .rdata:004D6234                 dd offset unk_4C85C9
+    .rdata:004D6238                 dd offset unk_4C85C3
+    .rdata:004D623C                 dd offset unk_4C85BB
+    .rdata:004D6240                 dd offset unk_4C85B3
+
+    From the disassembly, they are called as follows:
+    .text:00498E56                 push    ds:off_4D61E0[ecx*4]
```
williballenthin, line 489: where does the length get stored?

williballenthin (approved):
lgtm but let's also wait for @mr-tz

floss/language/utils.py
Arker123: It's not; the raw xrefs are stored without explicit length information. Lengths are not included in this context. Do we need them, or is there a specific reason for considering length storage?

williballenthin: no, I don't think we need them. but I wondered how Go is able to use the string data without an associated length.

Arker123: Currently, I couldn't find any specific information for Go, but I did come across a similar approach in Rust. I've kept it in the utils.py file, considering the possibility that we might encounter a similar scenario in the future when exploring other languages.

Arker123:
Hey @mr-tz, can we merge this?

floss/language/utils.py
```diff
@@ -473,6 +473,48 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
...
+        address = last
+        last = current
+
+        if address == 0x0:
+            continue
+
+        if not (low <= address < high):
+            continue
+
+        yield address
```
VascoSch92, line 515:
```suggestion
        if address != 0x0 and low <= address < high:
             yield address
```
much simpler

floss/language/utils.py
```diff
@@ -473,6 +473,43 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
             # dozens of seconds or more (suspect many minutes).
 
 
+def get_raw_xrefs_rdata_i386(pe: pefile.PE, buf: bytes) -> Iterable[VA]:
+    """
+    scan for raw xrefs in .rdata section.
+    raw xrefs are 32-bit absolute addresses to strings in .rdata section (i386).
```
williballenthin, line 479: > raw xrefs are 32-bit absolute addresses to strings in .rdata section (i386).

This routine doesn't  validate that the destination is string-like data. Should it? If not, lets remove this part of the documentation.

floss/language/utils.py
williballenthin: I am still curious how this data can be used as a string without the length being stored somewhere.

floss/language/utils.py
```diff
@@ -473,6 +473,43 @@ def get_struct_string_candidates(pe: pefile.PE) -> Iterable[StructString]:
...
+
+    The above are not struct String instances, but are references to strings in .rdata section.
+    They can be used to divide the string blobs into smaller chunks.
+    """
+    format = "I"
+
+    if not buf:
+        return
+
+    low, high = get_image_range(pe)
```
williballenthin, line 499: this appears to be the range of the entire file, not the `.rdata` section. please update the logic or documentation to make things consistent.

tests/test_language_extract_rust.py
```diff
@@ -80,3 +80,21 @@ def test_push(request, string, offset, encoding, rust_strings):
...
+        pytest.param("Invalid branch target in DWARF expression", 0xB813A, StringEncoding.UTF8, "rust_strings32"),
+        pytest.param(
+            "Expected to find an FDE pointer, but found a CIE pointer instead.",
+            0xB8163,
+            StringEncoding.UTF8,
+            "rust_strings32",
+        ),
+    ],
+)
+def test_raw_xrefs(request, string, offset, encoding, rust_strings):
```
williballenthin, line 99: i think its good that we have integration tests that show the whole system working together to find the strings. I think we should also have some tests for the specific routines that you added, so we can verify their behavior directly. something like `test_get_raw_xrefs`.

williballenthin (changes requested):
The concept here seems reasonable, though the documentation is inconsistent with the logic, so please fix that and then we can merge.

tests/test_language_extract_rust.py
Arker123: Hi, what should be the best approach to that? As there are not just `test_get_raw_xrefs`, but also others such as `find_i386_push_xrefs`, `find_lea_xrefs`, etc., should we test them too in a separate file or another PR? What are your thoughts?

mr-tz:
closing stale PR

PR #901, opened 2023-11-10
fix extractions and improve language_strings_missed contents

### LINKED ISSUE #875: Improve language-strings storage (JSON) and show value to users
Currently, the result document JSON contains:
1. `static_strings`: all static strings
2. `language_strings`: language-specific strings
3. `language_strings_missed`: all static strings minus what was found in 2.

This results in:
- a bunch of duplicate data
- false/confusing entries in 3., e.g., it currently includes all strings found outside the targeted language-specific sections like `.data` or `.rdata`

We also don't show the real value clearly yet. Like with FLOSS we see x more strings now easily digestible by humans.

Solution ideas:
1. remove `language_strings` and `language_strings_missed` and just update `static_strings` - easiest
2. don't include static strings for language samples
3. ...

### LINKED ISSUE #900: verify language and include identified version in results JSON
1. Potentially add check to verify that selected language makes sense - warn user just in case.
2. Include the identified language version in the results JSON.

mr-tz:
Various fixes including to update `language_strings_missed` to only include strings from the range (e.g. `.rdata` section for Rust or the strings blob in Go) that may have been missed during extraction.

This renders language strings separately from static strings. Language strings are part of static strings and it's currently not possible to display them individually. This can be added later, if desired.

closes #875
closes #900

### DIFF
--- a/floss/results.py
+++ b/floss/results.py
@@ -1,5 +1,6 @@
 # Copyright (C) 2021 Mandiant, Inc. All Rights Reserved.
 
+import re
 import json
 import datetime
 from enum import Enum
@@ -138,6 +139,9 @@ class StaticString:
             decoded_string = buf.decode("utf-8")
         except UnicodeDecodeError:
             raise ValueError("not utf-8")
+
+        if not re.sub(r"[\r\n\t]", "", decoded_string).isprintable():
+            raise ValueError("not printable")
 
         if len(decoded_string) < min_length:
             raise ValueError("too short")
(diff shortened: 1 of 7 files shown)


floss/results.py
```diff
@@ -139,6 +140,9 @@ def from_utf8(cls, buf, addr, min_length):
         except UnicodeDecodeError:
             raise ValueError("not utf-8")
 
+        if not re.sub(r"[\r\n\t]", "", decoded_string).isprintable():
+            raise ValueError("not printable")
```
mr-tz, line 144: we've had something similar before and removed it, during my testing this was helpful though

Arker123 (approved):

PR #908, opened 2023-11-14
update identify.py

sara-rn:
identify Go samples with stomped PCNLTAB magic bytes
reused function get_rdata_section from language.rust

### DIFF
--- a/floss/language/identify.py
+++ b/floss/language/identify.py
@@ -10,6 +10,7 @@ import pefile
 import floss.logging_
 from floss.results import StaticString
 from floss.rust_version_database import rust_commit_hash
+from floss.language.utils import get_rdata_section
 
 logger = floss.logging_.getLogger(__name__)
 
@@ -89,7 +90,16 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
         b"\xfa\xff\xff\xff\x00\x00",
         b"\xf1\xff\xff\xff\x00\x00",
     ]
-
+    go_functions = [
+        b"runtime.main",
+        b"main.main",
+        b"runtime.gcWork",
+        b"runtime.morestack",
+        b"runtime.morestack_noctxt",
+        b"runtime.newproc",
+        b"runtime.gcWriteBarrier",
+        b"runtime.Gosched",
+    ]
     # look for the .rdata section first
     for section in pe.sections:
         try:
@@ -116,6 +126,32 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
                 pclntab_va = section_data.index(magic) + section_va
                 if verify_pclntab(section, pclntab_va):
                     return True, get_go_version(magic)
+
+    # if not found, the magic bytes may have been patched, search for common Go functions present in all Go samples including obfuscated files
+    # look for the .rdata section first
+    try:
+        section = get_rdata_section(pe)
+        section_va = section.VirtualAddress
+        section_size = section.SizeOfRawData
+        section_data = section.get_data(section_va, section_size)
+        for go_function in go_functions:
+            if go_function in section_data:
+                logger.info("Go binary found, function name %s", go_function)
+                return True
+    except ValueError:
+        logger.debug(".rdata section not found")
+
+    # if not found, search commin all the available sections
+    for section in pe.sections:
+        section_va = section.VirtualAddress
+        section_size = section.SizeOfRawData
+        section_data = section.get_data(section_va, section_size)
+        for go_function in go_functions:
+            if go_function in section_data:
+                logger.info("Go binary found, function name %s", go_function)
+                return True
+
+    return False
     return False, VERSION_UNKNOWN_OR_NA
 
 
--- a/floss/language/identify.py
+++ b/floss/language/identify.py
@@ -135,7 +135,7 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
     except ValueError:
         logger.debug(".rdata section not found")
 
-<<<<<<< HEAD
+
     section_va = section.VirtualAddress
     section_size = section.SizeOfRawData
     section_data = section.get_data(section_va, section_size)
@@ -144,10 +144,7 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
             logger.info("Go binary found, function name %s", go_function)
             return True, VERSION_UNKNOWN_OR_NA
 
-    # if not found, search commin all the available sections
-=======
     # if not found, search in all the available sections
->>>>>>> 55a835f20616d085a0b8bcf5ef20de5083690246
     for section in pe.sections:
         section_va = section.VirtualAddress
         section_size = section.SizeOfRawData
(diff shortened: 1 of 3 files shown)


floss/language/utils.py
```diff
@@ -59,7 +59,13 @@ class StructString:
     address: VA
     length: int
 
+def get_rdata_section(pe: pefile.PE) -> pefile.SectionStructure:
```
mr-tz, line 62: please also remove from `rust/extract.py`

floss/language/identify.py
```diff
@@ -89,7 +90,16 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
...
+        b"runtime.gcWork",
+        b"runtime.morestack",
+        b"runtime.morestack_noctxt",
+        b"runtime.newproc",
+        b"runtime.gcWriteBarrier",
+        b"runtime.Gosched",
+    ]
     # look for the .rdata section first
     for section in pe.sections:
         try:
```
mr-tz, line 105: please use `get_rdata_section` here as well

floss/language/identify.py
```diff
@@ -116,6 +126,31 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
                 pclntab_va = section_data.index(magic) + section_va
                 if verify_pclntab(section, pclntab_va):
                     return True, get_go_version(magic)
+
+    # if not found, the magic bytes may have been patched, search for common Go functions present in all Go samples including obfuscated files
+    # look for the .rdata section first
```
mr-tz, line 131: should we do this as part of the above loops? or is it better to keep them sequential?

floss/language/identify.py
```diff
@@ -10,6 +10,7 @@
 import floss.logging_
 from floss.results import StaticString
 from floss.rust_version_database import rust_commit_hash
+from floss.language.utils import get_rdata_section
```
mr-tz, line 13:
```suggestion
from floss.language.utils import get_rdata_section
from floss.rust_version_database import rust_commit_hash
```

mr-tz: see isort fail (I think)

floss/language/identify.py
sara-rn: done

floss/language/identify.py
sara-rn: I would say I would keep them sequential because it is very likely the version can be extracted in the first loop..

mr-tz: ok, sounds good then

floss/language/identify.py
```diff
@@ -89,22 +90,30 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
...
-            section_va = section.VirtualAddress
-            section_size = section.SizeOfRawData
-            section_data = section.get_data(section_va, section_size)
-            for magic in go_magic:
-                if magic in section_data:
-                    pclntab_va = section_data.index(magic) + section_va
-                    if verify_pclntab(section, pclntab_va):
-                        return True, get_go_version(magic)
+    try:
+        section = get_rdata_section(pe)
```
mr-tz, line 105: let's only do this call in the try/except and the rest afterwards

mr-tz: same below

floss/language/identify.py
```diff
@@ -116,6 +125,31 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
...
+        section_size = section.SizeOfRawData
+        section_data = section.get_data(section_va, section_size)
+        for go_function in go_functions:
+            if go_function in section_data:
+                logger.info("Go binary found, function name %s", go_function)
+                return True, VERSION_UNKNOWN_OR_NA
+    except ValueError:
+        logger.debug(".rdata section not found")
+
+    # if not found, search commin all the available sections
```
mr-tz, line 143:
```suggestion
    # if not found, search in all the available sections
```

floss/language/identify.py
```diff
@@ -89,22 +90,32 @@ def get_if_go_and_version(pe: pefile.PE) -> Tuple[bool, str]:
...
-                if magic in section_data:
-                    pclntab_va = section_data.index(magic) + section_va
-                    if verify_pclntab(section, pclntab_va):
-                        return True, get_go_version(magic)
+    try:
+        section = get_rdata_section(pe)
+    except ValueError:
+        logger.debug(".rdata section not found")
+
+    section_va = section.VirtualAddress
```
mr-tz, line 109: this now needs to be in an else clause so we don't end up here even if there's no .rdata section

stevemk14ebr:
The concept of how this should work seems solid to me. I like that .rdata is checked first, but that all sections are scanned after, this is important for obfuscated binaries with garbled section names.

If we ever discover actors begin to zero out the strings we search for, we can migrate to signature scanning assembly per architecture as I do in GoReSym https://github.com/mandiant/GoReSym/blob/master/objfile/scanner.go but I don't believe this is neccessary at all for now. And I quite like the simpler string based scanning approach to start with.

floss/language/rust/extract.py
```diff
@@ -17,12 +17,6 @@
 MIN_STR_LEN = 4
 
 
-def get_rdata_section(pe: pefile.PE) -> pefile.SectionStructure:
```
mr-tz, line 20: references to this in rust/extract.py need to be updated

mr-tz (changes requested):
please apply the black code style formatting `black -l 120` on all files

can you please share a sample hash? I'd like to add a test as well (or feel free to do that yourself).

floss/language/rust/extract.py
sara-rn: added the function name in the import

floss/language/rust/extract.py
```diff
@@ -10,19 +10,13 @@
 import binary2strings as b2s
 
 from floss.results import StaticString, StringEncoding
-from floss.language.utils import find_lea_xrefs, find_mov_xrefs, find_push_xrefs, get_struct_string_candidates
+from floss.language.utils import find_lea_xrefs, find_mov_xrefs, find_push_xrefs, get_struct_string_candidates, get_rdata_section
```
mr-tz, line 13:
```suggestion
from floss.language.utils import find_lea_xrefs, find_mov_xrefs, find_push_xrefs, get_rdata_section, get_struct_string_candidates
```

floss/language/rust/extract.py
```diff
@@ -10,19 +10,13 @@
 import binary2strings as b2s
 
 from floss.results import StaticString, StringEncoding
-from floss.language.utils import find_lea_xrefs, find_mov_xrefs, find_push_xrefs, get_struct_string_candidates
+from floss.language.utils import find_lea_xrefs, find_mov_xrefs, find_push_xrefs, get_rdata_section, get_struct_string_candidates
```
mr-tz, line 13:
```suggestion
from floss.language.utils import (
    find_lea_xrefs,
    find_mov_xrefs,
    find_push_xrefs,
    get_rdata_section,
    get_struct_string_candidates,
)
```

floss/language/rust/extract.py
```diff
@@ -10,19 +10,19 @@
...
 
 MIN_STR_LEN = 4
 
 
-def get_rdata_section(pe: pefile.PE) -> pefile.SectionStructure:
-    for section in pe.sections:
-        if section.Name.startswith(b".rdata\x00"):
-            return section
-
-    raise ValueError("no .rdata section found")
```
mr-tz, line 27:
```suggestion
```

mr-tz:
Thanks, @sara-rn!

PR #909, opened 2023-11-16
add test of Go sample with stomped PCNLTAB magic bytes

mr-tz:


mr-tz:
@sara-rn for awareness

PR #912, opened 2023-11-22
improve Rust identification

mr-tz:


Arker123 (commented):
Considering the reliability, wouldn't prioritizing hash searches over strings be a more robust approach?

williballenthin (approved):

mr-tz:
What do you mean by reliability?

Arker123:
I meant that when considering reliability, I thought prioritizing hash searches over strings might be more robust. My intuition is that the string 'rust<version>' could appear in many files, whereas hashes are likely to be more unique. What are your thoughts?

mr-tz:
I think it's fairly robust. If we encounter issues we can come back to it.

PR #938, opened 2024-01-21
Fix: Handle default prompt when stdout is redirected

### LINKED ISSUE #930: when stdout is redirected, use default prompt values
when using `floss foo.exe | less`, the prompt for "should disable deobfuscating strings in rust binary?" is hidden because less has the foreground. in this scenario, we should use the default prompt answer ("N").

Arker123:
Fixes #930 - Use default prompt answer ("N"), when using ```floss foo.exe | less```.


mr-tz (approved):
lgtm, thanks!

PR #953, opened 2024-02-29
Indicate duplicate string

### LINKED ISSUE #911: qs: indicate duplicate strings
for example:
```
string 1
string 2
string 3
string 1  << duplicate
string 4
````

this way unique strings can quickly be identified but no data is hidden by default

### PR DESCRIPTION
**Pull request by ooprathamm**
Closes #911
 
#duplicate tag + Mute
* Sample output
image

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -355,18 +355,6 @@ def render_string(width: int, s: TaggedString, tag_rules: TagRules) -> Text:
     return line
 
 
-def get_duplicate_strings(strings: Sequence[ExtractedString]) -> Set[str]:
-    seen = set()
-    duplicates = set()
-
-    for s in strings:
-        if s.string in seen:
-            duplicates.add(s.string)
-        seen.add(s.string)
-
-    return duplicates
-
-
 def get_reloc_offsets(slice: Slice, pe: pefile.PE) -> Set[int]:
     ret: Set[int] = set()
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by williballenthin** [comment]
This is a great start. I wonder if we should not tag the first instance of the string, only the second+ as duplicate. Otherwise important, but duplicated, strings would all get muted every time. 

Do you think this is feasible @ooprathamm?

File: floss/qs/main.py [hunk 1]
```diff
@@ -651,13 +651,21 @@ def tag_strings(self, taggers: Sequence[Tagger]):
         this can be overridden, if a subclass has more ways of tagging strings,
         such as a PE file and code/reloc regions.
         """
+        string_counts = {}
```
mr-tz, line 654: can use defaultdict(int) here

mr-tz (commented):
I think like this is good. How would you otherwise show the << data? In a new column?

ooprathamm:
I was considering rendering count on the console directly ahead of strings, adding a new column for duplicate strings would be excessive. However, appending the count to the strings disrupts other functionalities. Hence, I suggest proceeding with the "#duplicate" tag.

floss/qs/main.py
```diff
@@ -651,13 +652,21 @@ def tag_strings(self, taggers: Sequence[Tagger]):
...
+        
         tagged_strings: List[TaggedString] = []
+
         for string in self.strings:
             # at this moment, the list of strings contains only ExtractedStrings.
             # this routine will transform them into TaggedStrings.
             assert isinstance(string, ExtractedString)
             tags: Set[Tag] = set()
 
+            string_counts[string.string] = string_counts.get(string.string, 0) + 1
```
mr-tz, line 665:
```suggestion
            string_counts[string.string] += 1
```

mr-tz:
thanks!

floss/qs/main.py
williballenthin: or `collections.Counter`

PR #955, opened 2024-03-04
Identify XOR Encoded PE files

### LINKED ISSUE #757: identify embedded XOR encoded PE files and render those
extension of #756

ooprathamm:
Closes #757 

Taking reference from how capa identifies xor encodeded [[features/extractors/helper.py](https://github.com/mandiant/capa/blob/master/capa/features/extractors/helpers.py)]

image

image

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -11,7 +11,7 @@ import argparse
 import functools
 import itertools
 import contextlib
-from typing import Set, Dict, List, Tuple, Literal, Callable, Iterable, Optional, Sequence
+from typing import Set, Dict, List, Union, Tuple, Literal, Callable, Iterable, Optional, Sequence
 from dataclasses import field, dataclass
 
 import pefile
@@ -378,6 +378,13 @@ def get_reloc_offsets(slice: Slice, pe: pefile.PE) -> Set[int]:
     return ret
 
 
+def check_is_xor(xor_key: Union[int, None]):
+    if isinstance(xor_key, int):
+        return ("#decoded",)
+
+    return ()
+
+
 def check_is_reloc(reloc_offsets: Set[int], string: ExtractedString):
     for addr in string.slice.range:
         if addr in reloc_offsets:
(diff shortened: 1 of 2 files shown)


williballenthin (commented):
really neat! can we somehow name the region to indicate the data was decoded and the key used?

ooprathamm:
* Rendering PE is XOR decoded and key in the header
image

williballenthin:
Implementation looks great. 

Question for the room ( @mr-tz ): this mode now prints strings that aren't visible verbatim from strings.exe. Is that ok?

williballenthin (approved):

mr-tz:
Cool! Should we add a new tag like `#floss` or `#decoded`?

mr-tz:
How worried are we about performance here? Can you provide some stats about how long this runs for a small and a huge binary?

williballenthin:
> #decoded

This sounds good to me!

ooprathamm:
@mr-tz I did some pretty basic profiling on the compute_layout() 
* Metric: Execution time
* Each call was performed 15 times taking avg value for calculation of total_time of compute_layout for each file
image

* The results were pretty inconsistent across multiple runs, showing only lowermost and uppermost values

ooprathamm:
* #decoded tag when strings are xor_decoded

image

mr-tz:
Nice, that looks good to me. We could discuss `#decoded` vs. a better term further, but this works for now.

Weird that the install fails, can we revert to lancelot 0.8.7 to see CI succeed?

ooprathamm:
@williballenthin can I make a commit downgrading lancelot ?

williballenthin:
sounds good to me. i won't be able to fix lancelot for maybe a week or so.

mr-tz (approved):
lgtm, thanks!

floss/qs/main.py
```diff
@@ -1051,6 +1089,8 @@ def render_strings(
     BORDER_STYLE = MUTED_STYLE
 
     name = layout.name
+    if isinstance(layout, PELayout) and layout.xor_key:  # Check if the layout is PELayout and is xored
+        name += f" (XOR decoded with key: {layout.xor_key})"
```
mr-tz, line 1093:
```suggestion
        name += f" (XOR decoded with key: 0x{layout.xor_key:x})"
```

mr-tz: last little tweak

PR #959, opened 2024-03-07
changed deprecated functionality

Sylan-Padmakumar:
changed deprecated fspath mentions to self.path as per the python 3 documentation

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -152,7 +152,7 @@ class FLOSSTest(pytest.Item):
         self._test_strings(str(test_path))
 
     def reportinfo(self):
-        self.path, 0, "usecase: %s" % self.name
+        return self.path, 0, "usecase: %s" % self.name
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -152,7 +152,7 @@ def runtest(self):
         self._test_strings(str(test_path))
 
     def reportinfo(self):
-        return self.fspath, 0, "usecase: %s" % self.name
+        self.path, 0, "usecase: %s" % self.name
```
williballenthin, line 155: i think the `return` was lost here

williballenthin (changes requested):

tests/conftest.py
Sylan-Padmakumar: changes made

williballenthin (approved):

mr-tz:
Thanks!

PR #964, opened 2024-03-11
QS: Fuzzy search strings

### LINKED ISSUE #705: ui: fuzzy search the strings

ooprathamm:
Closes #705 

* Experimented with Levenshtein distance and jaro similarity.
* Tokenizing Levenshtein distance for sentence-like strings yielded the best results
*  For Levenshtein distance calculations, the RapidFuzz module was utilized. If a custom function for TaggedStrings is preferred, please let me know :)

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -11,6 +11,7 @@ import argparse
 import functools
 import itertools
 import contextlib
+from rapidfuzz import fuzz
 from collections import defaultdict
 from typing import Set, Dict, List, Union, Tuple, Literal, Callable, Iterable, Optional, Sequence
 from dataclasses import field, dataclass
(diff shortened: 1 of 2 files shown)


williballenthin:
hey @ooprathamm!

Sorry that you're missing a bit of context here: the QS issues tagged "ui" are ideas for an interactive TUI program (that doesn't exist yet). We had in mind to extend QS to have a bunch of additional functionality for exploring strings, like searching, inspecting, cross referencing, etc. using a console interface similar to ranger, gitui, htop, etc. But, we haven't finalized the design or built it yet, so these issues track our ideas.

Anyways, thank you for the PR here - I think we'll be able to use the implementation if/when we add searching to the TUI.

Issue #745, opened 2023-05-19
Windows - right click menu - open with FLOSS


mr-tz:
Provide an option to install right click menu option for Windows (if easily possible). Otherwise, we could do it in FLARE VM.

mr-tz:
closed via #970

PR #970, opened 2024-03-14
Provide an option to install right click menu option for Windows

lyc8503:
This implements #745.

Implemented two options for installing and uninstalling the Windows context menu and tested them on Windows.

By the way, I've also added a hint to upgrade pip and setuptools in the documentation (PEP660 has only been implemented in newer versions of setuptools)

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -77,6 +77,9 @@ This means that Python will load the FLOSS module from this local
 This is good, because it is easy for us to modify files and see the
  effects reflected immediately.
 But be careful not to remove this directory unless uninstalling FLOSS!
+If you encounter the error `ERROR: Project has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook.`,
+ please ensure that you have upgraded to the latest versions of pip and setuptools.
+
 
 - Install FLOSS:
 
--- a/floss/main.py
+++ b/floss/main.py
@@ -257,13 +257,21 @@ def make_parser(argv):
         advanced_group.add_argument(
             "--install-right-click-menu",
             action=floss.utils.InstallContextMenu,
-            help="install FLOSS to the right-click context menu for Windows Explorer and exit",
+            help=(
+                "install FLOSS to the right-click context menu for Windows Explorer and exit"
+                if show_all_options
+                else argparse.SUPPRESS
+            ),
         )
 
         advanced_group.add_argument(
             "--uninstall-right-click-menu",
             action=floss.utils.UninstallContextMenu,
-            help="uninstall FLOSS from the right-click context menu for Windows Explorer and exit",
+            help=(
+                "uninstall FLOSS from the right-click context menu for Windows Explorer and exit"
+                if show_all_options
+                else argparse.SUPPRESS
+            ),
         )
 
     output_group = parser.add_argument_group("rendering arguments")
--- a/floss/main.py
+++ b/floss/main.py
@@ -10,7 +10,6 @@ from enum import Enum
 from time import time
 from typing import Set, List, Optional
 from pathlib import Path
-import platform
 
 import halo
 import viv_utils
@@ -145,20 +144,6 @@ def make_parser(argv):
     )
     parser.register("action", "extend", floss.utils.ExtendAction)
     parser.add_argument("-H", action="help", help="show advanced options and exit")
-    
-    if platform.system() == "Windows":
-        parser.add_argument(
-            "--install-right-click-menu",
-            action=floss.utils.InstallContextMenu,
-            help="install FLOSS to the right-click context menu for Windows Explorer and exit"
-        )
-
-        parser.add_argument(
-            "--uninstall-right-click-menu",
-            action=floss.utils.UninstallContextMenu,
-            help="uninstall FLOSS from the right-click context menu for Windows Explorer and exit"
-        )
-
     parser.add_argument(
         "-n",
         "--minimum-length",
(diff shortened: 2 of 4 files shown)


floss/main.py
```diff
@@ -144,6 +145,19 @@ def make_parser(argv):
     )
     parser.register("action", "extend", floss.utils.ExtendAction)
     parser.add_argument("-H", action="help", help="show advanced options and exit")
+    
+    if platform.system() == "Windows":
+        parser.add_argument(
+            "--install-right-click-menu",
```
mr-tz, line 151: cool, let's add this as an advanced option (see `show_all_options` usage below)

doc/installation.md
```diff
@@ -77,6 +77,9 @@ This means that Python will load the FLOSS module from this local
 This is good, because it is easy for us to modify files and see the
  effects reflected immediately.
 But be careful not to remove this directory unless uninstalling FLOSS!
+If you encounter the error `ERROR: Project has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook.`,
```
mr-tz, line 80: 👍

floss/utils.py
```diff
@@ -10,6 +10,7 @@
 from typing import Set, Tuple, Iterable, Optional
 from pathlib import Path
 from collections import OrderedDict
+import sys
```
mr-tz, line 13: please run this through isort, black, etc. so the code styler linter passes

lyc8503: Sorry, I didn't notice the code linter.😂 Should have fixed it in the latest commit.

mr-tz:
This looks great. I will test it next week and let you know if there's any other changes needed. Thank you very much!

floss/main.py
```diff
@@ -254,6 +253,18 @@ def make_parser(argv):
         version="%(prog)s {:s}".format(__version__),
         help="show program's version number and exit" if show_all_options else argparse.SUPPRESS,
     )
+    if sys.platform == "win32":
+        advanced_group.add_argument(
+            "--install-right-click-menu",
+            action=floss.utils.InstallContextMenu,
+            help="install FLOSS to the right-click context menu for Windows Explorer and exit",
```
mr-tz, line 260: we need to add `if show_all_options else argparse.SUPPRESS` to these

floss/utils.py
```diff
@@ -42,6 +43,48 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+    def __call__(self, parser, namespace, values, option_string=None):
+        # Theoretically, we don't need to check the platform again here,
+        # because non-Windows platforms will not accept the --install-right-click-menu parameter at all.
+        # This judgment is just to make the mypy type check pass.
+        # The same logic applies to `UninstallContextMenu` below.
+        if sys.platform == "win32":
+            import winreg as reg
+
+            menu_name = "Open with FLOSS"
+            menu_command = f'C:\\windows\\system32\\cmd.exe /K "python -m floss ^"%1^""'
```
mr-tz, line 59: this won't work if we run the standalone binary
can you think of an implementation that supports both usage/installation methods?

floss/utils.py
```diff
@@ -42,6 +43,48 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+
+            shell_key = reg.OpenKey(reg.HKEY_CURRENT_USER, r"Software\\Classes\\*\\shell", 0, reg.KEY_SET_VALUE)
```
mr-tz, line 61: on my system, only `Computer\HKEY_CURRENT_USER\SOFTWARE\Classes\*\shellex\` exists, can we create the key or do we need to support multiple registry paths?

mr-tz: `Computer\HKEY_CLASSES_ROOT\*\shell\` is another potential path

floss/utils.py
```diff
@@ -42,6 +43,48 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+            reg.SetValue(shell_key, menu_name, reg.REG_SZ, menu_name)
+
+            menu_key = reg.OpenKey(shell_key, menu_name, 0, reg.KEY_SET_VALUE)
+            reg.SetValueEx(menu_key, "AppliesTo", 0, reg.REG_SZ, "System.ItemName:exe")
```
mr-tz, line 65: we can run FLOSS on any file, so we don't need this

mr-tz: but ideally, we can set the icon to the FLOSS icon via the `Icon` value name

floss/utils.py
lyc8503: I considered using the `Computer\HKEY_CLASSES_ROOT\*\shell\` path, but it requires administrator privileges to write, and affects other users using the same computer (not all of them have Python and floss installed, and may not have access to the current user's standalone floss.exe). I'll try to create `shell` when it doesn't exist.

floss/utils.py
lyc8503: It seems that we didn't include the icon for floss in the pip installation, so I only added the icon when the user uses a standalone exe.

mr-tz (commented):
Awesome, thank you very much. I like this solution and if people have issues given the various installation methods, we can fix this.

Only thing left to do is a small entry in the documentation at https://github.com/mandiant/flare-floss/blob/master/doc/usage.md. Can you add this, please?

lyc8503:
Hi, I have updated the document, and thanks for your code review and feedback too. 😃

mr-tz (approved):
great work, thanks a lot!

Issue #696, opened 2023-04-30
Import Detection Heuristics from `mrphrazer/obfuscation_detection` Repository


r0ny123:
This issue note aims to suggest the integration of detection heuristics from the [mrphrazer/obfuscation_detection](https://github.com/mrphrazer/obfuscation_detection) by @mrphrazer into the `FLOSS` tool, to enhance its capabilities in detecting and decoding obfuscated strings in malware samples. The [mrphrazer/obfuscation_detection](https://github.com/mrphrazer/obfuscation_detection) repository contains a comprehensive set of detection heuristics for various obfuscation techniques mentioned [here](https://github.com/mrphrazer/obfuscation_detection#detection-heuristics).

Integrating these heuristics into `FLOSS` would allow the tool to recognize a wide range of obfuscation methods and ultimately improve its decoding performance.

mr-tz:
Very cool! Thanks for sharing this here. Could be useful for FLOSS and capa.

c-urly:
Hi @mr-tz  , I was thinking of trying this out.

Do you have any heuristics that you want to implement from the repo? or we need to put some scoping effort on this?

So if i take xor decryption loops heuristics as an example, We need to find the function with the xor loops and print its offset right? I am trying to figure out once we get the heuristics how we will print it out.

mr-tz:
Hey, I don't have specific heuristics in mind so feel free to propose some here. XOR decryption loops sounds like a good start.

Once we have the heuristics, we'd want to plug them into the decoding function identification code (e.g. [`find_decoding_function_features`](https://github.com/mandiant/flare-floss/blob/ff9de8ed1157a86f764ad681f1a30a9ba2df9130/floss/identify.py#L115)). These functions are then emulated to obtain potential obfuscated strings.

c-urly:
Hi @mr-tz ,
In binaryViewType class in binaryNinja, we have a block.dominance_frontier to detect loop. 
Is there such a feature in viv_utils which can detect if the basic block is in a loop or not?

c-urly:
I have made some changes should I raise a PR on master branch?
What i did is extracted instruction with a immediate value as operand and did an intersection with xor + loop. So xor decryption logic will be xor + loop + imm heuristic.

mr-tz:
Did you check https://github.com/mandiant/flare-floss/blob/master/floss/features/extract.py? We have loop and XOR detection, for example. If you have more heuristics, feel free to open a PR :)

[...]

PR #981, opened 2024-03-21
Added XOR decryption heuristic from mrphrazer/obfuscation_detection

c-urly:
PR for #696 
Added heuristic to check instruction with immediate value. 
Added xor decryption loop heuristic from mrphrazer/obfuscation_detection

### DIFF
--- a/floss/features/extract.py
+++ b/floss/features/extract.py
@@ -12,6 +12,7 @@ from envi.archs.i386.opconst import INS_MOV, INS_ROL, INS_ROR, INS_SHL, INS_SHR,
 import floss.logging_
 from floss.const import TS_TIGHT_FUNCTION_MAX_BLOCKS
 from floss.features.features import (
+    Imm,
     Mov,
     Loop,
     Nzxor,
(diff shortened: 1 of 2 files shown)


floss/features/extract.py
```diff
@@ -34,6 +36,16 @@
 logger = floss.logging_.getLogger(__name__)
 
 
+def extract_insn_const(f, bb, insn):
+    """
+    parse const instruction operand: OP dst, src, imm
+    """
+    for op in insn.opers:
+        if op.isImmed():
+            print("Imm:", insn)
```
c-urly, line 45: Oh  I missed this print statement, please ignore it for now.

floss/features/features.py
```diff
@@ -127,6 +127,10 @@ def score(self):
         return 1.0
 
 
+class Imm(Mnem):
+    weight = HIGH
```
c-urly, line 131: Please let me know, how to assign weights to the class?

mr-tz:
Where do you see a benefit versus the generic `nzxor`? It seems we capture all cases using that heuristic already.

mr-tz:
@c-urly friendly bump on my question above.

c-urly:
Yes i also think that we dont need these changes here. I was thinking to resuse this PR for other heuristics but if you want i can delete the pr and raise new one for other heuristics.

mr-tz:
Ok, then let's use a new PR for future heuristics. Thank you!

PR #985, opened 2024-04-06
QS: Migrate to pkgutil for databases 

### LINKED ISSUE #759: use pkgutil to fetch database contents
via: https://pyinstaller.org/en/stable/spec-files.html#using-data-files-from-a-module

```python
import pkgutil
help_bin = pkgutil.get_data( 'helpmod', 'help_data.txt' )
```

ooprathamm:
Closes #759

### DIFF
--- a/floss/qs/db/expert.py
+++ b/floss/qs/db/expert.py
@@ -1,11 +1,9 @@
 import re
-import pathlib
+import pkgutil
 from typing import Set, Dict, List, Tuple, Literal, Sequence
 from dataclasses import dataclass
 
 import msgspec
-
-import floss.qs.db
 
 
 class ExpertRule(msgspec.Struct):
@@ -51,13 +49,13 @@ class ExpertStringDatabase:
         return ret
 
     @classmethod
-    def from_file(cls, path: pathlib.Path) -> "ExpertStringDatabase":
+    def from_file(cls, package: str, resource: str) -> "ExpertStringDatabase":
         string_rules: Dict[str, ExpertRule] = {}
         substring_rules: List[ExpertRule] = []
         regex_rules: List[Tuple[ExpertRule, re.Pattern]] = []
 
         decoder = msgspec.json.Decoder(type=ExpertRule)
-        buf = path.read_bytes()
+        buf = pkgutil.get_data(package, resource)
         for line in buf.split(b"\n"):
             if not line:
                 continue
@@ -81,9 +79,10 @@ class ExpertStringDatabase:
             regex_rules=regex_rules,
         )
 
-
-DEFAULT_PATHS = (pathlib.Path(floss.qs.db.__file__).parent / "data" / "expert" / "capa.jsonl",)
+DEFAULT_FILENAMES = (
+    "capa.jsonl",
+)
 
 
 def get_default_databases() -> Sequence[ExpertStringDatabase]:
-    return [ExpertStringDatabase.from_file(path) for path in DEFAULT_PATHS]
+    return [ExpertStringDatabase.from_file('floss.qs.db', 'data/expert/' + f) for f in DEFAULT_FILENAMES]
--- a/floss/qs/db/expert.py
+++ b/floss/qs/db/expert.py
@@ -1,5 +1,6 @@
 import re
 import pathlib
+import pkgutil
 from typing import Set, Dict, List, Tuple, Literal, Sequence
 from dataclasses import dataclass
 
@@ -51,13 +52,13 @@ class ExpertStringDatabase:
         return ret
 
     @classmethod
-    def from_file(cls, path: pathlib.Path) -> "ExpertStringDatabase":
+    def load_database(cls, buf: bytes) -> "ExpertStringDatabase":
         string_rules: Dict[str, ExpertRule] = {}
         substring_rules: List[ExpertRule] = []
         regex_rules: List[Tuple[ExpertRule, re.Pattern]] = []
 
         decoder = msgspec.json.Decoder(type=ExpertRule)
-        buf = path.read_bytes()
+
         for line in buf.split(b"\n"):
             if not line:
                 continue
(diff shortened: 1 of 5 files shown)


mr-tz:
overall this appears to add complexity vs. reducing it. can you walk me through the benefits of this approach?

ooprathamm:
`pkgutil.get_data `  is supposed to be better at getting a resource from a package.

Profiling load_databases:
image
image
(Similar perf across multiple runs though)

* As QS is distributed as a bundled `.exe` pkgutil references directly to these embedded files, rather than relative to `floss.qs.db` path.
* Directly reading bytes, rather than 2 step approach to find path and then `read_bytes()`

Also pyinstaller .spec guide mentions both the approaches without specifing any added benefits [[using .spec]](https://pyinstaller.org/en/stable/spec-files.html#using-data-files-from-a-module:~:text=When%20your%20script%20executes%2C%20you%20could%20find%20help_data.txt%20by%20using%20its%20base%20folder%20path%2C%20as%20described%20in%20the%20previous%20section.%20However%2C%20this%20data%20file%20is%20part%20of%20a%20module%2C%20so%20you%20can%20also%20retrieve%20its%20contents%20using%20the%20standard%20library%20function%20pkgutil.get_data()) 

@williballenthin could you provide us with your concerns for opening the issue .

williballenthin:
I understand `pkgutil` to expose an interface that Python packagers will implement when building single-file executable, like PyInstaller. The `get_data` routine lets you access supporting data (non-.py files) found within a module. 

This doesn't really matter for packagers that extract a copy of the module to the file system during loading, line PyInstaller. For these packagers, we can use pathlib to access file system paths reletive to Python module contents. 

However, there are packagers that don't write module content to disk. They might keep Python byte code in memory, having read it from a compressed resource or even from the network. In this case, we can't rely on using pathlib to access data files. We should use pkgutil instead. 

But! We might decide that we don't want to support those packagers (which seem to be uncommon) and consider this out of scope. Or, we could follow the spec and be prepared to switch to something other than PyInstaller. 

Does that make sense? @mr-tz @ooprathamm

floss/qs/db/expert.py
```diff
@@ -51,13 +49,13 @@ def query(self, s: str) -> Set[str]:
         return ret
 
     @classmethod
-    def from_file(cls, path: pathlib.Path) -> "ExpertStringDatabase":
+    def from_file(cls, package: str, resource: str) -> "ExpertStringDatabase":
```
williballenthin, line 52: let's keep `from_file` around since it's really useful for dev and testing. but let's add a `from_pkgutil` to address the feature request. and please factor out the common code into a sub routine.

williballenthin (changes requested):
we should also try to find one of these packagers and see if it works with the proposed changes. I suspect Nuitka might work for this.

floss/qs/db/expert.py
ooprathamm: Which routine will the `get_default_databases` use for now ?

williballenthin: please prefer pkgutil

ooprathamm:
For simplicity, the code for using `from_file` is commented out.
Or we could leave something like `BASE_PATH = pathlib.Path(floss.qs.db.__file__).parent ` constant to infer on own.

williballenthin (approved):
these changes look great!

I still think we should try to find a package that uses the pkgutil interface. Do you have some time to find one @ooprathamm?

ooprathamm:
Current Scenario - 
Pyinstaller - Uses a bootloader that [uncompresses](https://pyinstaller.org/en/stable/operating-mode.html#how-the-one-file-program-works:~:text=The%20bootloader%20uncompresses%20the%20support%20files%20and%20writes%20copies%20into%20the%20the%20temporary%20folder.) data files to a temp folder. [[Limitation](https://pyinstaller.org/en/stable/operating-mode.html#:~:text=In%20order%20to%20locate%20included%20files%20at%20run%20time%2C%20your%20program%20needs%20to%20be%20able%20to%20learn%20its%20path%20at%20run%20time%20in%20a%20way%20that%20works%20regardless%20of%20whether%20or%20not%20it%20is%20running%20from%20a%20bundle.)]

Other Executable Builder Packages-

- py2exe -  Windows Only, data files bundled alongside the .exe file in the directory structure
- cx_Freeze - Doesn't embed data files to executables. Keeps data files in a [directory](https://cx-freeze.readthedocs.io/en/stable/faq.html#data-files:~:text=They%E2%80%99ll%20be%20copied%20to%20the%20build%20directory%20alongside%20the%20executable) structure
- Pyoxidizer [unmaintained now] - [pkgutil compatability](https://pyoxidizer.readthedocs.io/en/stable/oxidized_importer_behavior_and_compliance.html?#:~:text=pkgutil%20Compatibility) mentioned for importing modules from memory, for[ loading data files](https://pyoxidizer.readthedocs.io/en/latest/oxidized_importer_resource_files.html#:~:text=Loading%20Resource%20Files,-%C2%B6) encourages the use of  importlib.resources, doesn't mention pkgutil

Viable Option - 

- Nuitka - Python -> C++ -> executable. [Supports pkgutil for loading data files](https://nuitka.net/user-documentation/user-manual.html#data-files:~:text=Standalone%3A%20Finding%20files,the%20standalone%20binary)

williballenthin:
first off, great research @ooprathamm!

next, the pyoxidizer documentation is outstanding. it does nice job of summarizing the options. and since it doesn't touch on pkgutil, I wonder if we're using that wrong? or is it the new interface alluded to at the end?

based on that article, what do you think the best strategy is, @ooprathamm? and, would you consider trying out pyoxidizer on FLOSS to see if it works with our code?

ooprathamm:
On further looking into PyOxidizer.
pkgutil.get_data() enhancement still pending [github](https://github.com/indygreg/PyOxidizer/issues/457#:~:text=pkgutil.get_data()%20to%20read%20their%20bundled%20resources%2C%20currently%20fail%20when%20the%20sources%20are%20in%20memory%2C%20as%20they%20lack%20a%20__file__%20attribute.)

The incompatibility is further confirmed here.  ([comparing different utilities ](https://github.com/indygreg/PyOxidizer/issues/649#:~:text=I%27ve%20tested%20several,pkgutil%20never%20works)).

For now, It might not be good to consider PyOxidizer

williballenthin:
ok sounds good

vee1e:
stale, `quantumstrand` is now `master` and bundling is fixed via https://github.com/mandiant/flare-floss/commit/266d7c6c so now pyinstaller works with QS-inherited databases

the pkgutil idea from #759 is still good but needs a fresh PR against `floss/tags/*` on master.

i recommend we close, but thanks for the PR @ooprathamm 

cc: @mr-tz

mr-tz:
Agreed, thanks all.

PR #1012, opened 2024-06-11
relax pyproject dependency versions and introduce requirements.txt

### LINKED ISSUE #1011: relax pyproject dependency versions and introduce requirements.txt
We should do the same things to floss, viv utils, dncil, etc. and perhaps make it standard practice (if it works).

_Originally posted by @williballenthin in https://github.com/mandiant/capa/issues/2132#issuecomment-2154623431_

williballenthin:
closes #1011


mr-tz (approved):
thanks

Issue #925, opened 2023-12-22
[BUG]The offset of the RUST STRINGS is misaligned.


pinksawtooth:
## Description

The offset of the "FLOSS RUST STRINGS" item in "FLOSS" is different from the actual value.

The following is an excerpt of the results of running floss on a Rust program I created.

```
 ──────────────────────────
  FLOSS RUST STRINGS (529)
 ──────────────────────────

0x01d590 invalid args
0x01d5b0 /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1\library\core\src\fmt\mod.rs
0x01d618 /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1\library\core\src\alloc\layout.rs
0x01d680 attempt to divide by zero
0x01d699 February
0x01d6a1 March
0x01d6a6 April
0x01d6ae June
0x01d6b2 July
0x01d6b6 August
0x01d6bc September
0x01d6c5 October
0x01d6cc November
0x01d6d4 December
0x01d6dc January
0x01d6e3 initial contents
0x01d6f3 to string:
0x01d720 Hello,
0x01d727 world!
0x01d72d add string:
0x01d760 global static:
0x01d790 global const:
0x01d7c0 This is const variableA
0x01d7e8 This is const variableB
0x01d810 This is const variableC
0x01d838 month:
0x01d860 This is static variableA
0x01d888 This is static variableB
0x01d8b0 This is static variableC
(snip)
```

This time I will focus on this string. (The problem is the same for all strings.)
`0x01d720 Hello,`

But the actual offset is `0x1e520` as shown in the image below.
![image](https://github.com/mandiant/flare-floss/assets/18203311/96bd5607-6e17-48db-a4da-35550484cc08)

```
>>> hex(0x1e520-0x01d720)
'0xe00'
```

The other strings are offset by the same amount. Verify with the following string

```
0x01d760 global static:`
```

Add 0xe00 to this offset.

```
>>> hex(0x1d760+0xe00)
'0x1e560'
```

Then the actual offset value will be displayed as follows.
`14001e560`

![image](https://github.com/mandiant/flare-floss/assets/18203311/027dcbde-042e-4d31-9abf-84d34bc2dae5)

I believe it is a bug in Rust Strings because the offset in static strings is normal.

```
+------------------------------------+
| FLOSS STATIC STRINGS: ASCII (2336) |
+------------------------------------+

0x00004d !This program cannot be run in DOS mode.
```

[...]

Issue #937, opened 2024-01-16
UI: show file offsets and virtual address in verbose output


mr-tz:
Display raw and virtual offsets in verbose mode.

This could help to match things up more quickly between FLOSS output, the file, and disassembler view.

Om-Thorat:
Hey, I would like to work on this. Though can you specify more? as to what out is needed? Should there be more columns or should i just log them as info?

mr-tz:
Hi! I imagined extra columns for this, correct.

Om-Thorat:
![Screenshot from 2024-02-27 18-59-44](https://github.com/mandiant/flare-floss/assets/76207818/4a3d6760-2552-4cc5-a355-a81989a2d665)


Will something like this be ideal?

mr-tz:
Yes, however, this doesn't apply so much to decoded strings vs. the language strings (Go and Rust).
If we had something for the language strings that would be amazing!

mr-tz:
Would also be interesting to show (and order by) `referenced from va` or `referenced from fva`.

Chee-Tzu:
Hello, is this still valid?

mr-tz:
Yes ☺️

divyansh360:
Hi @mr-tz, I’m applying for GSoC 2025 and would be happy to make my small contributions here, as I feel this could be a good start for the project relating to language-specific strings. Can I take it?

mr-tz:
Absolutely, thanks.

Vineet817:
Hi @mr-tz, I’m applying for GSoC 2026 and I am looking for small contributions here. 

I noticed that this issue appears to have been addressed by PR #1105, which adds offset and virtual address support for language-specific strings in verbose mode.

However, since the issue is still open, I wanted to ask whether there is any remaining work or follow-up expected for this task.

I would be happy to look into contributing.

PR #1019, opened 2024-06-23
Add verbose output of file offsets and virtual addresses for language-specific strings

Arker123:
This PR enhances the verbose output to display raw file offsets and virtual addresses for Rust strings.

Fixes: #925 #937

*The PR is in draft status because it currently only supports Rust strings.

image

### DIFF
--- a/floss/language/rust/extract.py
+++ b/floss/language/rust/extract.py
@@ -135,6 +135,23 @@ def get_static_strings_from_rdata(sample, static_strings) -> List[StaticString]:
     return list(filter(lambda s: start_rdata <= s.offset < end_rdata, static_strings))
 
 
+def get_file_offset_in_rdata(sample: pathlib.Path) -> int:
+    pe = pefile.PE(data=pathlib.Path(sample).read_bytes(), fast_load=True)
+
+    try:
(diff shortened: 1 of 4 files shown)


williballenthin:
do you think we should add column headers?

floss/language/rust/extract.py
```diff
@@ -135,6 +135,23 @@ def get_static_strings_from_rdata(sample, static_strings) -> List[StaticString]:
...
+    try:
+        rdata_section = get_rdata_section(pe)
+    except ValueError:
+        return -1
+
+    image_base = pe.OPTIONAL_HEADER.ImageBase
+    virtual_address = rdata_section.VirtualAddress
+    pointer_to_raw_data = rdata_section.PointerToRawData
+
+    print("DD:", image_base + virtual_address - pointer_to_raw_data)
```
mr-tz, line 150: remove please

mr-tz (commented):
how does this handle strings outside of .rdata?

alternative, we could consider adding a virtual offset field (and section info) to the StaticString class

Arker123:
Currently looks like this in verbose:-
image

mr-tz:
closing stale PR draft

PR #1105, opened 2025-03-26
Fixes issue #937: Show Offset and VirtAddr for language-specific strings (Go and Rust)

### LINKED ISSUE #937: UI: show file offsets and virtual address in verbose output

divyansh360:
This PR enhances the verbose output to display raw file offsets and virtual addresses for language-specific strings (Go and Rust)
File Offsets tested to match using Hex Viewer.
Virtual Addresses tested to match using r2.
Fixes: #937

Go Strings
Sample: 28f18dd863198ca06ee20bdd6893296ffd487099c8c3868550724b0484bc6057
Screenshot 2025-03-27 170805

Rust Strings (inspired and enhanced from #1019)
Sample: acc31048e00d1a0f4cd5569d5d4db539da8f506cc7a6a171942d015ecc817d43
Screenshot 2025-03-26 200143

Considering the string "readme.txt"
Screenshot 2025-03-27 160636

Looking forward to your feedback and guidance towards more contribution opportunities.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ lib/
 # Pyenv file
 .python-version
 .venv
+floss-venv/
 
 # Test executables
 bin/
@@ -33,3 +34,4 @@ flare_floss.egg-info
 .direnv/
 .env/
 .envrc
+*.code-workspace
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -424,6 +424,25 @@ def get_static_strings_from_blob_range(sample: pathlib.Path, static_strings: Lis
     return list(filter(lambda s: string_blob_start <= s.offset < string_blob_end, static_strings))
 
 
+def get_file_offset_in_blob(sample: pathlib.Path) -> int:
+    pe = pefile.PE(data=pathlib.Path(sample).read_bytes(), fast_load=True)
+
+    struct_strings = list(sorted(set(get_struct_string_candidates(pe)), key=lambda s: s.address))
+    if not struct_strings:
+        return -1
+
+    try:
+        string_blob_start, _ = find_string_blob_range(pe, struct_strings)
+    except ValueError:
+        return -1
+    
+    image_base = pe.OPTIONAL_HEADER.ImageBase
+    virtual_address = string_blob_start - image_base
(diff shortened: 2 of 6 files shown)


mr-tz:
Cool, thanks! Can you please add some tests showing the offsets are calculated as expected?

mr-tz:
Additionally let's check first if the data can be added to the string classes/datatypes during extraction.

divyansh360:
> Cool, thanks! Can you please add some tests showing the offsets are calculated as expected?

I edited the description to include these.
Also, an interesting thing I noticed is that, for Go strings, the initial few strings (untill "runtime method:" string) can be correctly matched with their offsets in hex viewer, but for virtual address a difference of `0xe00` needs to be made to reach to correct virtual address. I observed this same pattern in multiple Go samples, and also all of them had the same initial few strings requiring the adjustment.
For instance consider the string "l512"
Screenshot 2025-03-26 200044
Screenshot 2025-03-27 170027


I feel this is because of the initial Go runtime metadata block. I tried to dynamically make the adjustment by running a while loop to check for and use the first non-runtime string as a reference point, to stop making the adjustment. 

But is this a proper fix? Would appreciate your guidance about the possible cause and how to proceed. Thanks!

divyansh360:
> Additionally let's check first if the data can be added to the string classes/datatypes during extraction.

Sure, I'll check if I can add it

williballenthin:
> But is this a proper fix? Would appreciate your guidance about the possible cause and how to proceed.

I'm not sure why this is so, but I think its worth a bit more investigation. I don't really like the strategy of brute forcing the offset delta unless there's no other way. I wonder if some of the data is stored in a separate section?

vee1e:
thanks for this, but we can close this as duplicate of #1220. both address #937, but #1220 keeps the fix per-string instead of a global `Metadata.file_offset`. That fits the current `StaticString` model on master and keeps the VA in the results file. I left review notes on #1220 to remove the reformatting and fix the Go last-pointer VA.

mr-tz:
Thanks all!

PR #1130, opened 2025-06-07
Migrate to modern logger interface

emmanuel-ferdman:
## Description
The `logger.warn()` method is deprecated since Python2.7 and replaced with `logger.warning()`. It leads to those warnings:
```python
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```


williballenthin:
nice!

williballenthin (approved):

mr-tz (approved):
Thanks!

PR #1139, opened 2025-07-07
QS enhancements

mr-tz:
This PR introduces a web-based visualizer for FLOSS QuantumStrand analysis results, refactors the core data structures for clarity, and adds a new test suite for the QS analysis pipeline.

   - Web-based Visualizer: A new React-based web application (floss/qs/qs-web) has been added to provide an interactive way to explore QS analysis results. It allows users to upload a results.json file and view
     the file layout, strings, and tags in a user-friendly interface.
   - Data Structure Refactoring:
       - The Layout class in floss/qs/main.py has been refactored to simplify its structure and remove redundant fields (offset_from_dict, end_from_dict). Deserialization from JSON is now handled more cleanly.
       - The TaggedString class has been updated to make the structure field optional, and the to_dict method now omits the structure key if it's not present.
   - New Test Suite: A new test file, tests/test_qs.py, has been added to validate the QS analysis pipeline, including serialization/deserialization, string extraction, tagging, and structure marking.
   - XOR-decoding logic removed: The XOR-decoding logic has been removed from floss/qs/main.py.

---

The layout changes/serialization needs further work - ideally there's a generic layout for serialization/rendering so this is easily extendable for more file formats. But maybe that's not even necessary because we will only support a limited set (for now, i.e., PE and ELF).


floss/qs/main.py
```diff
@@ -79,12 +79,14 @@ class Slice:
     a bit like a memoryview.
     """
 
-    buf: bytes
+    buf: Optional[bytes]
```
williballenthin: when would this be None?

floss/qs/main.py
```diff
@@ -116,7 +123,36 @@ class ExtractedString:
...
+
+        return TaggedString(
+            string=ExtractedString(
+                string=string,
+                slice=Slice(None, Range(offset, len(string))),
+                encoding=encoding,
+            ),
+            tags=tags,
+            structure=structure,
+        )
```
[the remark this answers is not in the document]
williballenthin: agree with Gemini that length should be computed from the encoded bytes, not from the string character length.

floss/qs/main.py
```diff
@@ -591,6 +617,29 @@ class Layout(abc.ABC):
...
     strings: List[TaggedString] = field(init=False, default_factory=list)
 
+    def to_dict(self) -> Dict:
+        d = {
+            "name": self.name,
+            "offset": self.offset,
+            "size": self.slice.range.length,
+            "children": [child.to_dict() for child in self.children],
+        }
+        return d
```
williballenthin: we might want to rely on Pydantic to do this (de)serialization for us.

I'm not sure I like the JSON not matching the Python dataclasses. I'd suggest the nesting and structure be the same so its easier to map JS to/from Python.

floss/qs/main.py
```diff
@@ -696,66 +746,49 @@ def mark_structures(self, structures: Optional[Tuple[Dict[int, Structure], ...]]
...
-
-
 @dataclass
 class PELayout(Layout):
-    # xor key if the file was xor decoded
-    xor_key: None
-
     # file offsets of bytes that are part of the relocation table
-    reloc_offsets: Set[int]
+    reloc_offsets: Optional[Set[int]] = field(init=False, default=None)
```
williballenthin: why not empty set?

williballenthin: its ok to provide a default initializer to the empty set.

floss/qs/main.py
mr-tz: when deserializing from the JSON

floss/qs/main.py
mr-tz: so should we move from dataclass to pydantic dataclass/BaseModel?

williballenthin: that seems reasonable to me. should reduce the amount of code we have to write and maintain.

mr-tz: on it, I'm also reconsidering using your originally proposed structure where each layout stores its strings nested under itself

pydanti by default results in a quite verbose serialization (with all the the current classes, e.g., slice + range) so maybe we want to fine tune that - or not :)

floss/qs/qs-web/README.md
```diff
@@ -0,0 +1,12 @@
...
+This template provides a minimal setup to get React working in Vite with HMR and some ESLint rules.
+
+Currently, two official plugins are available:
+
+- [@vitejs/plugin-react](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react) uses [Babel](https://babeljs.io/) for Fast Refresh
+- [@vitejs/plugin-react-swc](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react-swc) uses [SWC](https://swc.rs/) for Fast Refresh
+
+## Expanding the ESLint configuration
+
+If you are developing a production application, we recommend using TypeScript with type-aware lint rules enabled. Check out the [TS template](https://github.com/vitejs/vite/tree/main/packages/create-vite/template-react-ts) for information on how to integrate TypeScript and [`typescript-eslint`](https://typescript-eslint.io) in your project.
```
williballenthin:
````suggestion
## dev

```console
$ npm install --include=dev
$ npm run dev
```

Then visit: http://localhost:5173
````

mr-tz:
see #1146

PR #1146, opened 2025-07-08
Switch to Pydantic and render based on Result* data

mr-tz:
Supersedes #1139 (part in that PR are still outstanding here so it stays a good reference for now).

Moving to pydantic and rendering based on a simplified serialized JSON in the form of
```json
{
  "name": "pe",
  "offset": 0,
  "length": 16384,
  "strings": [],
  "parent": null,
  "children": [
    {
      "name": "header",
      "offset": 0,
      "length": 4096,
      "strings": [
        {
          "string": "!This program cannot be run in DOS mode.",
          "offset": 77,
          "size": 40,
          "encoding": "ascii",
          "tags": [
            "#common"
          ],
          "structure": ""
        },
        {
          "string": "`.rdata",
          "offset": 519,
          "size": 7,
          "encoding": "ascii",
          "tags": [
            "#common"
          ],
          "structure": "section header"
        },
        {
          "string": "@.data",
          "offset": 559,
          "size": 6,
          "encoding": "ascii",
          "tags": [
            "#common"
          ],
          "structure": "section header"
        }
      ],
      "parent": "pe",
      "children": [],
      "has_visible_predecessors": false,
      "has_visible_successors": true
    },
    {
      "name": ".text",
      "offset": 4096,
      "length": 4096,
      "strings": [
        {
          "string": "ugh 0@",
          "offset": 4460,
          "size": 6,
          "encoding": "ascii",
          "tags": [
            "#code"
          ],
          "structure": ""
        }
      ],
      "parent": "pe",
      "children": [],
      "has_visible_predecessors": true,
      "has_visible_successors": true
    },
...
```

Strictly speaking the dataclasses could remain for the internal representations

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -12,13 +12,13 @@ import functools
 import itertools
 import contextlib
 from collections import defaultdict
-from typing import Set, Dict, List, Union, Tuple, Literal, Callable, Iterable, Optional, Sequence
+from typing import Set, Dict, List, Union, Tuple, Literal, Callable, Iterable, Optional, Sequence, Annotated
 
 import pefile
 import colorama
 import lancelot
 import rich.traceback
-from pydantic import Field, BaseModel
+from pydantic import Field, BaseModel, model_serializer, model_validator
 from rich.text import Text
 from rich.style import Style
 from rich.console import Console
@@ -49,9 +49,6 @@ class Range(BaseModel):
     offset: int
     length: int
 
-    # class Config:
-    #     arbitrary_types_allowed = True
-
     @property
     def end(self) -> int:
         return self.offset + self.length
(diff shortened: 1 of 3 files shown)


floss/qs/main.py
```diff
@@ -248,10 +251,24 @@ class ResultLayout(BaseModel):
     offset: int
     length: int
     strings: List[ResultString]
-    parent: Optional[str] = ""
+    parent: Optional[str] = ""  # TODO remove?
     children: List["ResultLayout"]
-    has_visible_predecessors: bool
-    has_visible_successors: bool
+    has_visible_predecessors: bool  # TODO used for rendering, remove?
+    has_visible_successors: bool  # TODO used for rendering, remove?
```
williballenthin: agree that maybe this shouldn't be in the model itself, but something that can be computed as needed.

williballenthin: reason: it would be easy to end up in an inconsistent state if someone modifies the layout, such as adding/removing a sibling.

mr-tz: changed

PR #1153, opened 2025-07-17
Remove transitive dependencies

foivospro:
As part of our ongoing research on Python dependency management, we noticed a potential improvement in your project’s declared dependencies.

Specifically, several dependencies appear to be  transitive, meaning they are already required by top-level packages and do not need to be listed explicitly.

The following dependencies were removed from the configuration files after verifying they are not directly used:

`msgpack` `annotated-types` `termcolor` `markdown-it-py` `dncil` `spinners` `python-flirt` `pyasn1-modules` `six` `cxxfilt` `pycparser` `funcy` `log-symbols` `colorama` `mdurl`

Removing these helps keep the dependency list clean and maintainable. It also allows pip’s resolver to handle transitive requirements more efficiently.

The [pip documentation](https://pip.pypa.io/en/stable/user_guide/#requirements-files) also encourages auditing your top-level requirements and removing unused or transitive ones to simplify the dependency graph and minimize the risk of version conflicts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,6 @@ dependencies = [
     # latest bug fixes we need here.
     "viv-utils[flirt]>=0.7.9",
     "vivisect>=1.1.1",
-    "dncil>=1.0.2",
 
     # ---------------------------------------
     # Dependencies with version caps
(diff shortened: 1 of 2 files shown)


williballenthin:
we use requirements.txt as a lock file that intentionally includes transitive dependencies.

mr-tz:
Thank you for the suggestion but see the comment above.

PR #1165, opened 2025-08-31
Allowing users to add tags to strings

FredCoast:
Web App updates
1. Allowed individual strings to have new tags added
2. Added a selection button so that users can add a tag to multiple strings 
3. Added an export JSON button to allow users to export their new tags
4. Sorted out some scaling issues affecting zooming

Python Updates
1. Added a new arg to allow users to update their personal expert database by importing their updated tagged strings


vee1e:
Stale paths, merge conflicts plus viewer scaling changes. The viewer was renamed and rebuilt on master, so diff conflicts heavily.

I recommend we close and if needed, refile viewer and tag-db parts against master.

cc: @mr-tz

mr-tz:
Agreed, thanks all!

PR #1210, opened 2026-01-27
feat: add snprintf/sprintf hook

### LINKED ISSUE #318: apicall hook `snprintf` function and variants
Some decoding routines rely on string formatting routines such as `snprintf`. Example: `9dab...106c`

codesensei-tushar:
### Add SnprintfHook implementation
Fixes #318

### Changes
- Implement `snprintf`/`sprintf` hook using Python's native `%` formatting
- Handle pointer dereferencing for `%s` format specifiers by reading strings from emulator memory
- Support signed/unsigned integer conversion for `%d`, `%i`, `%u`, `%x`, `%X`, `%o`
- Support width and precision modifiers (e.g., `%10s`, `%.5s`, `%08x`)
- Handle both `sprintf(buf, fmt, ...)` and `snprintf(buf, size, fmt, ...)` variants

### Testing
Tested with this sample C file: [test_snprintf_comprehensive.c](https://github.com/user-attachments/files/24886181/test_snprintf_comprehensive.c)


floss/api_hooks.py
```diff
@@ -329,6 +329,94 @@ def __call__(self, emu, api, argv):
             fu.call_return(emu, api, argv, ptr)
             return True
 
+class SnprintfHook:
```
williballenthin, line 333: please add some tests that demonstrate this routine works the way we think it does

codesensei-tushar: Thanks for the review! I'd like to clarify what kind of tests you're looking for:

- **Unit tests** - pytest tests that mock the emulator and verify SnprintfHook._prepare_args() and .call() work correctly with various format specifiers.

- **Integration tests** - Tests that run FLOSS on an actual binary (like the C test file I shared) and verify decoded strings are extracted.

codesensei-tushar:
Added `tests/test_snprintf_hooks.py` in the latest commit! 
All 11 tests passing. Let me know if you'd like any additional test cases!

williballenthin (approved):
thank you @codesensei-tushar!

the unit tests look good

codesensei-tushar:
I did run black previously, but it might not have run properly. In the latest commit, I've verified that all pre-commit checks:

The mypy check was failing due to a pre-existing issue in `floss/utils.py `(unrelated to the `SnprintfHook` changes). The error was a false positive — the code is correct, but mypy couldn't verify it statically due to the dynamic `**kwargs` pattern used for platform-specific `mmap` arguments:
Without a type annotation, `mypy` sees `**kwargs` and can't determine whether it contains valid arguments for `mmap`, so it flags it as an error. I added `# type: ignore[arg-type]` to suppress this false positive.

vee1e:
lgtm @codesensei-tushar could you resolve the merge conflicts?

vee1e:
thanks tushar! @mr-tz if you'd give the final verdict

mr-tz:
thanks!

PR #1215, opened 2026-02-15
fix: update comment handling in render-binja-import-script.py

schecthellraiser606:
1. ret of get_comment_at is str

> https://api.binary.ninja/binaryninja.function-module.html#binaryninja.function.Function.get_comment_at

2. set_comment is unkown in ver5.2

> https://api.binary.ninja/binaryninja.function-module.html#binaryninja.function.Function.set_comment_at


mr-tz (approved):
thanks!

PR #1223, opened 2026-03-03
Fix b2s wide string garbage extraction in Rust binaries issue #867

svm1048:


vee1e:
@mr-tz i think this can also be closed in action to willi's comment here https://github.com/mandiant/flare-floss/issues/867#issuecomment-4172873804

mr-tz:
Thanks, closing this now.

PR #1232, opened 2026-03-15
Add Go decoded stackstring extraction via runtime.slicebytetostring callsites

### LINKED ISSUE #828: Update Stackstrings Algorithm in Go Extraction Code
The current stackstrings algorithm used in the Go extraction code needs to be updated to improve efficiency and accuracy. The existing implementation might not be optimal for handling certain scenarios and can lead to incorrect results.

kunalsz:
Resolves #828 

This PR aims to implement :
-  Go-specific decoded stackstring pass that locates runtime.slicebytetostring (and common aliases) in the vivisect workspace and collects its callsites
- Follows the idea : `Note that generally decoding strings such as this involves converting a byte array back into a string, so stackstrings are typically followed by a call to runtime_slicebytetostring` as discussed in #828 

Till now a module has been added(`pclntab.py`) to parse Go metadata (pclntab/moduledata) for function names and addresses as the repo doesnt have Go runtime symbol resolution helpers.
This is used to locate `runtime.slicebytetostring` ,  function VA of `runtime.slicebytetostring`  and list of call sites.
pclntab_floss

Next steps would be to , recover the decoded bytes at each identified call site and hence the runtime stackstrings.

References/Inspirations for the work : 
- https://github.com/[dipusone/golang_pclntab_parser](https://github.com/dipusone/golang_pclntab_parser)
- https://go.dev/src/debug/gosym/pclntab.go

Note : For `test_language_go_pclntab.py` help of copilot was taken to build version specific minimal pclntab blob

### DIFF
--- a/floss/language/go/extract.py
+++ b/floss/language/go/extract.py
@@ -20,17 +20,19 @@ import struct
 import logging
 import pathlib
 import argparse
-from typing import List, Tuple, Iterable, Optional
+from typing import Dict, List, Tuple, Iterable, Optional
 from pathlib import Path
 from itertools import chain
 from dataclasses import dataclass
 
 import pefile
+import vivisect.const
 from typing_extensions import TypeAlias
 
 import floss.utils
 from floss.results import StaticString, StringEncoding
 from floss.language.utils import StructString, find_lea_xrefs, get_struct_string_candidates
+from floss.language.go.pclntab import find_runtime_functions
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


kunalsz:
@mr-tz A small surface level review for the direction of the PR would be helpful! 🙃

floss/language/go/extract.py
```diff
@@ -424,6 +426,92 @@ def get_static_strings_from_blob_range(sample: pathlib.Path, static_strings: Lis
...
+       to build a name -> VA map and then look up xrefs in the vivisect workspace
+
+    The two approaches are complementary: step 1 is O(names) + fast; step 2
+    handles the obfuscated case that is the whole motivation for this work
+    """
+    _SLICEBYTETOSTRING_TOKENS = (
+        "runtime.slicebytetostring",
+        "runtime_slicebytetostring",
+        "slicebytetostring",
+    )
```
vee1e, line 449: The token `"slicebytetostring"` is a substring match, so `runtime.slicebytetostringtmp` also hits. That helper still exists in Go's `runtime/string.go` (often intrinsified, still present in pclntab on some builds). The longer tokens are redundant once the short token is in the tuple. Extra callees would send later recovery at the wrong sites.

**Suggestion:** Match the exact names `runtime.slicebytetostring` and `runtime_slicebytetostring` with an explicit `tmp` exclusion).

floss/language/go/extract.py
```diff
@@ -424,6 +426,92 @@ def get_static_strings_from_blob_range(sample: pathlib.Path, static_strings: Lis
...
+        for va, name in items:
+            lname = str(name).lower()
+            if any(tok in lname for tok in _SLICEBYTETOSTRING_TOKENS):
+                callers = _xrefs_to(va)
+                if callers:
+                    out[va] = callers
+    except Exception as exc:
+        logger.debug("vw.getNames() failed: %s", exc)
+
+    if out:
```
vee1e, line 473: any vivisect name-table hit with xrefs and skips pclntab. Name-table VAs can be thunks or stubs. pclntab should be the source of truth for the real function VA.

Both paths still depend on vivisect xrefs for callers, so a wrong callee VA yields wrong or empty call sites and never gets a second chance.

always resolve the callee VA from pclntab when it is available. Use the name table only as a fallback when pclntab is missing.

tests/test_language_go_pclntab.py
```diff
@@ -0,0 +1,515 @@
...
+
+    try:
+        pe = pefile.PE(str(fixture_path), fast_load=True)
+    except Exception:
+        pytest.skip(f"pefile could not open {fixture_path}")
+
+    result = parse_pclntab(pe)
+    assert isinstance(result, dict)
+    # For real Go binaries we expect at least some functions
+    if result:
```
vee1e, line 512: - the "real binary" test passes when `parse_pclntab` returns `{}`. `8f62cf...exe_` has no pclntab magic and returns empty, so that case cannot fail.
- the synthetic tests only round-trip blobs that `_build_v12` / `_build_v116` / `_build_v118` emit in the parser's own layout (Copilot wrote those blobs)
- There are no tests for `find_slicebytetostring_callsites`.

suggestion: make known-good fixtures such as `go-hello64.exe` and `amd64_go1.{12,16,18,20}`, assert a non-empty map and that `runtime.slicebytetostring` is present. add a mocked-vw test for call-site collection. also keep one sample that is allowed to be empty only if it is documented as non-Go or packed.

floss/language/go/extract.py
```diff
@@ -424,6 +426,92 @@ def get_static_strings_from_blob_range(sample: pathlib.Path, static_strings: Lis
...
 
 
+def find_slicebytetostring_callsites(vw, pe: Optional[pefile.PE] = None) -> Dict[int, List[int]]:
+    """
+    Locate all call sites to runtime.slicebytetostring within the vivisect workspace vw
+    Returns {callee_va: [caller_va, ...]}
+
+    Strategy
+    1. vivisect name table fast path that works for non-stripped binaries
+       where the Go toolchain embeds symbol names into the ELF/PE symbol table
```
vee1e, line 436: the docstring says the vivisect name table covers ELF/PE symbols, but the pclntab fallback only opens `pefile.PE`. That matches today's PE-only `extract_go_strings`. `vw.getFiles()[0]` plus `pefile.PE` fails closed on ELF and the exception is logged at debug. The pclntab module docstring also claims stripped ELF.

floss/language/go/extract.py
```diff
@@ -424,6 +426,92 @@ def get_static_strings_from_blob_range(sample: pathlib.Path, static_strings: Lis
     return list(filter(lambda s: string_blob_start <= s.offset < string_blob_end, static_strings))
 
 
+def find_slicebytetostring_callsites(vw, pe: Optional[pefile.PE] = None) -> Dict[int, List[int]]:
+    """
+    Locate all call sites to runtime.slicebytetostring within the vivisect workspace vw
+    Returns {callee_va: [caller_va, ...]}
+
+    Strategy
```
vee1e, line 434: the new function docstring restates the two-step strategy and the "motivation for this work". Several new comments only narrate control flow (`# empty return` in `pclntab.py`, `## step 1 vivisect name table`, the test-builder "NB: sentinel only has a va" note).

basically existing Go extract comments explain non-obvious constraints. These do not.

i recommend to delete comments that restate the next line and point at Go's pclntab.go for layout instead of retelling it in tests.

scripts/debug_find_slicebytetostring.py
```diff
@@ -0,0 +1,117 @@
+"""
+Debug script to tesst/locate runtime.slicebytetostring call sites in a Go binary
```
vee1e, line 2: fix the typo (`tesst`) and add the standard license header on new `floss/` and `scripts/` files. also either document the helper under `scripts/` or keep it out of the merge until the feature is wired in.

PR #1233, opened 2026-03-16
Fix RICH_HEADER attribute error

vee1e:
@williballenthin @mr-tz sorry for not catching this earlier, but there was a bug in #1231 which expected a RICH_HEADER.key if the attribute was already existing but `None`.

While running qs against the test binaries just now, in `dotnet-hello.exe` it resulted in an AttributeError.

This is fixed here by adding an additional check to ensure the rich header object isn't `None` to ensure it doesn't throw an error but ignore the code and continue.

image

This is probably a sign for me to work on better tests for qs that cover the wider range of properties being processed now 😅

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -890,7 +890,7 @@ def collect_pe_structures(slice: Slice, pe: pefile.PE) -> Sequence[Structure]:
                         )
                     )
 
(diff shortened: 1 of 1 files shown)


williballenthin:
thank you!

PR #1240, opened 2026-03-18
Refactor language coverage module

### LINKED ISSUE #1213: Refactor proposal: deduplicate language “coverage” CLI boilerplate (Go/Rust)
### Issue / current state 
The Go and Rust language coverage utilities (`floss/language/go/coverage.py` and `floss/language/rust/coverage.py`) contain near-identical CLI boilerplate: argument parsing , logging setup, and PE validation via pefile. This duplication increases maintenance cost and creates risk of subtle drift

### Why this refactor is needed
- Single source for shared CLI behavior reduces bugs and review overhead
- Makes future language additions easier and more consistent.

### How it will be done
Introduce a small helper module (`floss/language/cli_common.py`) that provides:
- a function to add common argparse option
- a function to configure logging consistently
- a helper to open/validate PE input 

@mr-tz Looking forward to your insights!

kunalsz:
Resolves #1213 

Introduces `language/cli_common.py`, deduplicating the argument parsing , logging setup, and PE validation logic into one file.
Makes it easier for setting up new languages

### DIFF
--- a/floss/language/cli_common.py
+++ b/floss/language/cli_common.py
@@ -1,0 +1,39 @@
+import logging
+import argparse
+
+import pefile
+
+
+def add_common_args(parser: argparse.ArgumentParser, default_min_length: int):
+    parser.add_argument("path", help="file or path to analyze")
+    parser.add_argument(
+        "-n",
+        "--minimum-length",
+        dest="min_length",
+        type=int,
+        default=default_min_length,
+        help="minimum string length",
+    )
+
+    logging_group = parser.add_argument_group("logging arguments")
+    logging_group.add_argument("-d", "--debug", action="store_true", help="enable debugging output on STDERR")
+    logging_group.add_argument(
+        "-q",
+        "--quiet",
+        action="store_true",
+        help="disable all status output except fatal errors",
+    )
+
+
+def configure_logging(debug):
+    level = logging.DEBUG if debug else logging.INFO
+    logging.basicConfig(level=level)
+    logging.getLogger().setLevel(level)
+
+
+def open_pe_or_none(path: str, logger: logging.Logger):
+    try:
+        return pefile.PE(path)
+    except pefile.PEFormatError as err:
+        logger.debug(f"NOT a valid PE file: {err}")
+        return None
--- a/floss/language/cli_common.py
+++ b/floss/language/cli_common.py
@@ -1,7 +1,23 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import logging
 import argparse
 
 import pefile
+
+logger = logging.getLogger(__name__)
 
 
 def add_common_args(parser: argparse.ArgumentParser, default_min_length: int):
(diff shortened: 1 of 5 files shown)


kunalsz:
Hi @mr-tz , if you are available can you review this 🙃

williballenthin:
@kunalsz  this code style is not consistent with the rest of the codebase. the code and message read like AI slop. please go back and demonstrate more effort working with the project or we'll close out this PR.

kunalsz:
Hi @williballenthin, sorry if it came across that way. I made a few improvements, but the logger messages were taken directly from the original code before the refactor

floss/language/cli_common.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2023 Google LLC
```
williballenthin: why does this say 2023?

floss/language/cli_common.py
```diff
@@ -0,0 +1,58 @@
...
+def configure_logging(args: argparse.Namespace):
+    if args.debug:
+        logging.basicConfig(level=logging.DEBUG)
+        logging.getLogger().setLevel(logging.DEBUG)
+    else:
+        logging.basicConfig(level=logging.INFO)
+        logging.getLogger().setLevel(logging.INFO)
+
+
+def open_pe_or_none(path: str):
```
williballenthin: i don't think this function is worth it. let's remove.

williballenthin (changes requested):

floss/language/cli_common.py
kunalsz: I used that from the other file, I'll update that to 2026

mr-tz:
@vee1e what do you think on this one?

vee1e:
> @vee1e what do you think on this one?

code looks fine to me. would be very useful if/when new language support is introduced. the above review comments are addressed so this seems good to go

floss/language/cli_common.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+
+def configure_logging(args: argparse.Namespace):
+    if args.debug:
+        logging.basicConfig(level=logging.DEBUG)
+        logging.getLogger().setLevel(logging.DEBUG)
+    else:
+        logging.basicConfig(level=logging.INFO)
+        logging.getLogger().setLevel(logging.INFO)
```
mr-tz, line 48:
```suggestion
        logging.getLogger().setLevel(logging.INFO)

```

mr-tz:
@kunalsz can you please address the failing CI checks?

floss/language/cli_common.py
```diff
@@ -0,0 +1,39 @@
...
+        "--quiet",
+        action="store_true",
+        help="disable all status output except fatal errors",
+    )
+
+
+def configure_logging(debug):
+    level = logging.DEBUG if debug else logging.INFO
+    logging.basicConfig(level=level)
+    logging.getLogger().setLevel(level)
```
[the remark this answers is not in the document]
vee1e: this is also worth taking a look into imo

vee1e:
@kunalsz lints are still failing

kunalsz:
@vee1e This shouldn't be happening, there were linting issues with files not in the scope of this PR (main.py and test_load.py). I still pushed their updated versions and it's passing on my machine
image

vee1e:
this is weird. @mr-tz do you have any idea what could be going wrong here?

mr-tz:
Looks like black changed(?) and fails due to some auto detection. 

Warning: Python 3.10 cannot parse code formatted for Python 3.15. To fix this: run Black with Python 3.15, set --target-version to py310, or use --fast to skip the safety check. Black's safety check verifies equivalence by parsing the AST, which fails when the running Python is older than the target version.

We should update the config or run settings in CI if that's the case.

vee1e:
@mr-tz this refactor could be useful as the general clean-up of main.py takes place in the qs/floss merge. considering the CI issues should we close the PR (keeping the issue open) and address it in the merger PR?

ref.: https://gist.github.com/vee1e/828e8ff65b06fc44691ec5bf28ec5d4f#215-how-flossmainpy-splits

vee1e:
@mr-tz this is done by #1337, safe to close

mr-tz:
thanks for the initial work @kunalsz

PR #1245, opened 2026-03-24
- pyinstaller include path fix

Dont-Copy-That-Floppy:
pyinstaller failed to create a self contained binary file on build. fixed by setting the base dir, and spec dir to run from build process from base dir.

### DIFF
--- a/.github/pyinstaller/floss.spec
+++ b/.github/pyinstaller/floss.spec
@@ -33,12 +33,14 @@ with open("./floss/version.py", "wb") as f:
     )
     f.write(("__version__ = '%s'" % version).encode("utf-8"))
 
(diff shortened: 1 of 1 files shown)


vee1e:
@Dont-Copy-That-Floppy can you please sign the CLA, the PR seems fine to me

Dont-Copy-That-Floppy:
@vee1e  Sorry man, I'm not entering into an agreement with Google then handing over my personal info just to update an Apache Licensed software which is what I already agreed to by submitting the PR. 

CLA enables:
dual licensing
proprietary forks
enterprise-only versions
internal-only derivatives of future versions

not including mandating I have a google account, which I don't, to sign the CLA.

PR #1250, opened 2026-03-31
build(deps): comment out pydantic-core to fix update conflicts

mr-tz:
Revert explicit `pydantic-core` listing and grouping to avoid version mismatch conflicts during updates.

As discussed, `pydantic` and `pydantic-core` are tightly coupled, and keeping them as separate explicit dependencies in `requirements.txt` leads to broken PRs when Dependabot tries to update them to incompatible versions.

### Changes:
- Commented out `pydantic-core` in `requirements.txt`.
- Removed the `pydantic` group from `.github/dependabot.yml`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,10 +11,6 @@ updates:
           - "pyasn1"
           - "pyasn1-modules"
           - "msgpack"
-      pydantic:
-        patterns:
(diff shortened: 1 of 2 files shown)


mr-tz:
this was merged but the UI fails here?!

PR #1253, opened 2026-04-02
Get list of bb start/end eas for loops in `extract.py`

kunalsz:
In reference to the TODO in `extract.py`:
```python
        if len(comp) >= 2:
            # TODO get list of bb start/end eas
            yield Loop(comp)
 ```

- Added loop BB range extraction for each SCC loop `(len(comp) >= 2)`, it now builds sorted `(start_ea, end_ea)` pairs from function basic blocks and passes them to Loop.
- Extended `Loop` in `features.py` to keep `bb_ranges` while preserving existing comp behavior for compatibility

PS. `main.py` and `tests/test_load.py` got added due to black formatting, they have nothing to do with this PR

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -126,7 +126,8 @@ def make_parser(argv):
         " 1. Go:   strings from binaries written in Go\n"
         " 2. Rust: strings from binaries written in Rust\n"
     )
-    epilog = textwrap.dedent("""
+    epilog = textwrap.dedent(
+        """
         only displaying core arguments, run `floss -H` to see all supported options
 
         examples:
@@ -138,8 +139,10 @@ def make_parser(argv):
 
           only extract stack and tight strings
             floss --only stack tight -- suspicious.exe
-        """)
-    epilog_advanced = textwrap.dedent("""
+        """
+    )
+    epilog_advanced = textwrap.dedent(
+        """
         examples:
           extract all strings from 32-bit shellcode
             floss -f sc32 shellcode.bin
@@ -149,7 +152,8 @@ def make_parser(argv):
         
           extract strings from a binary written in Go (if automatic language identification fails)
             floss --language go program.exe
-        """)
+        """
+    )
 
     show_all_options = "-H" in argv
 
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -3,7 +3,8 @@ import textwrap
 import floss.main
 
 # floss --no static -j tests/data/src/decode-in-place/bin/test-decode-in-place.exe
-RESULTS = textwrap.dedent("""
+RESULTS = textwrap.dedent(
+    """
 {
     "analysis": {
         "enable_decoded_strings": true,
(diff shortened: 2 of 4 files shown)


floss/main.py
williballenthin, line 1: revert this please

tests/test_load.py
williballenthin, line 1: revert this

williballenthin:
> Extended Loop in features.py to keep bb_ranges while preserving existing comp behavior for compatibility

This is not the right way to go about this. Features describe what people are looking for, so putting the addresses of found loops there doesn't make any sense.

When features are extracted, they're associated with a list of addresses where the feature was found. This is probably the right place to yield this information. Although to be honest, I'm not sure if it's worth the overhead of tracking this information because I can't really imagine many scenarios where people will care about the loop locations. That's why we haven't yet addressed this comment in the source code

williballenthin (changes requested):

williballenthin:
I consider addressing this to-do issue as fairly low value. So I am not willing to go back and forth many times refining a solution. I'm willing to review perhaps one or maybe two more revisions of this PR. Otherwise we'll close it out and address the TODO sometime in the future if it becomes important.

kunalsz:
> > Extended Loop in features.py to keep bb_ranges while preserving existing comp behavior for compatibility
> 
> This is not the right way to go about this. Features describe what people are looking for, so putting the addresses of found loops there doesn't make any sense.
> 
> When features are extracted, they're associated with a list of addresses where the feature was found. This is probably the right place to yield this information. Although to be honest, I'm not sure if it's worth the overhead of tracking this information because I can't really imagine many scenarios where people will care about the loop locations. That's why we haven't yet addressed this comment in the source code

@williballenthin So should I close the PR ? As it wont be a meaningful contribution if we revert the `Loop` feature change and keep the code in just `extract.py`

williballenthin:
> So should I close the PR ? As it wont be a meaningful contribution if we

This isn't how I think about the project. I'm not looking for "meaningful contributions" but whether or not the project and its code is improved.

kunalsz:
> I consider addressing this to-do issue as fairly low value. So I am not willing to go back and forth many times refining a solution. I'm willing to review perhaps one or maybe two more revisions of this PR. Otherwise we'll close it out and address the TODO sometime in the future if it becomes important.

I noticed quite a few other TODOs in the codebase. Since this one was of a low value, do you have any higher value TODOs you’d prefer I work on instead? I’d be happy to pick one that is more useful for the project

kunalsz:
> > So should I close the PR ? As it wont be a meaningful contribution if we
> 
> This isn't how I think about the project. I'm not looking for "meaningful contributions" but whether or not the project and its code is improved.

Understood ! What I meant was that I only added that change to support the loop location feature, so if that feature is not worth keeping, the changes in `extract.py` will also not improve codebase (atleast I can't think of any rn)😅

vee1e:
I agree with Willi, and I'd go even as far to say the TODO is misguided. loop BBs don't need to be reported to the application or the analyst, i can't see where they would be useful as they're purely for scoring. @kunalsz should we close this?

mr-tz:
Closing, thanks.

PR #1264, opened 2026-04-13
scripts: add project root to sys.path

### LINKED ISSUE #1185: No module named 'floss.results'
Title

mr-tz:
Fixes #1185.

This PR adds the project root to `sys.path` in various helper scripts. This ensures they can find the `floss` module regardless of how they are invoked (e.g. from within the `scripts/` directory) or if `floss` is not installed in the current environment.

### DIFF
--- a/scripts/idaplugin.py
+++ b/scripts/idaplugin.py
@@ -23,10 +23,15 @@ email: willi.ballenthin@gmail.com
 """
 
 import os
+import sys
 import time
 import logging
 from typing import List, Union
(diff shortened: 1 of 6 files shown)


williballenthin:
> or if floss is not installed in the current environment.

this doesn't seem very safe to me: dependencies might not be installed either.

mr-tz:
ups, sorry for the noise

williballenthin:
i'm happy to discuss :-) i was just surfing github before bed

PR #1273, opened 2026-04-20
QS: robustness fixes and unit tests

mr-tz:
This PR adds unit tests for Range and Slice boundary conditions and improves PE parsing robustness in quantumstrand.

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -90,6 +90,10 @@ class Slice(BaseModel):
     range: Range
 
     @property
+    def offset(self) -> int:
+        return self.range.offset
+
+    @property
     def data(self) -> bytes:
         "get the bytes in this slice, copying the data out"
         return self.buf[self.range.offset : self.range.end]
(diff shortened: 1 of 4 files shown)


mr-tz:
@vee1e please review when you have a moment

floss/qs/main.py
```diff
@@ -1811,7 +1832,10 @@ def compute_layout(slice: Slice) -> Layout:
     # If XOR key is found, apply XOR decoding
     if xor_key is not None:
         decoded_data = xor_static(slice.data, xor_key)
-        decoded_slice = Slice(buf=decoded_data, range=Range(offset=0, length=len(decoded_data)))
+        # Pad the decoded data to match the absolute offset, 
+        # so that Slice/Range logic based on absolute offsets still works.
+        decoded_buf = (b"\x00" * slice.offset) + decoded_data
+        decoded_slice = Slice(buf=decoded_buf, range=Range(offset=slice.offset, length=len(decoded_data)))
```
williballenthin, line 1838: hm good catch here. it seems inefficient to create this buffer with a bunch of NULLs as prefix, but maybe handling this another way would require a lot more code and book keeping

qs-viewer/src/App.tsx
```diff
@@ -218,7 +221,7 @@ const App: React.FC = () => {
   };
 
   const handleFocusView = () => {
-    const noisyTags = ['#code', '#code-junk', '#duplicate', '#reloc'];
+    const noisyTags = ['#code', '#code-junk', '#duplicate', '#reloc', '#common'];
```
williballenthin, line 224: maybe move this to a global const?

vee1e: agreed, there's some duplication going on here so a global value would be better

mr-tz:
sorry merged this accidentally, see #1274

PR #1294, opened 2026-05-18
perf: optimize is_all_zeros using fast bytes comparison

mr-tz:
Replaces `all([b == 0 for b in buffer])` in is_all_zeros with `buffer == b"\x00" * len(buffer)`. This eliminates intermediate list allocations and leverages C-level byte string equality checking for performance. Similar optimizations were validated and incorporated upstream in capa (PR 3078).

### DIFF
--- a/floss/utils.py
+++ b/floss/utils.py
@@ -392,7 +392,8 @@ def get_runtime_diff(time0):
 
 
 def is_all_zeros(buffer: bytes):
-    return all([b == 0 for b in buffer])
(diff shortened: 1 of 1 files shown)


mr-tz:
cc: @mike-hunhoff @williballenthin wrt https://github.com/mandiant/capa/pull/3078/

williballenthin (approved):

vee1e (approved):

PR #1301, opened 2026-05-23
Add AGENTS.md

### LINKED ISSUE #1288: add AGENTS.md to avoid slop
similar to https://github.com/mandiant/capa/issues/3056

> to say things like: "use pre-commit ... to lint"

should also have links to the algorithms for string extraction and documentation for the agent to go through

vee1e:
closes #1288 

Adds both an AGENTS.md for AI agents and a symlink to CLAUDE.md for Claude Code for them to go through the repository with the required prerequisite knowledge instead of spamming slop everywhere.

Some areas this could be improved might be reordering of sections to give agents the best context and covering specific cases where previous contributors submitted low quality code and why.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,115 @@
+# AGENTS.md
+
+## Project Overview
+
+FLOSS (FLARE Obfuscated String Solver) is a static analysis tool that automatically extracts and deobfuscates strings from malware binaries. It extends traditional string-extraction utilities (e.g., `strings.exe`) by using emulation-based techniques to recover strings that are never present in plaintext on disk.
+
+Maintained by Google/Mandiant FLARE team. Licensed under Apache 2.0.
+
+### String Categories
+
+| Type | Description |
+|------|-------------|
+| **Static** | Regular ASCII / UTF-16LE strings found directly in the binary |
+| **Stack** | Strings constructed byte-by-byte on the stack at runtime |
+| **Tight** | Stack strings built inside tight loops, refined further |
+| **Decoded** | Strings recovered by emulating decoding/decryption routines |
+| **Language-specific** | Go and Rust runtime strings extracted from metadata (no emulation required) |
+
+## Theory & Algorithm
+
+FLOSS performs static analysis combined with selective emulation. The pipeline is:
+
+1. **Workspace loading**: vivisect parses the binary, identifies functions, basic blocks, and cross-references.
+2. **Static string extraction**: regex scans for ASCII (≥4 chars) and UTF-16LE sequences via `binary2strings`.
+3. **Feature extraction** (`floss/identify.py`, `floss/features/`): each function is scored on instruction features: MOV counts, XOR (non-zeroing), shifts/rotates, tight loops, call counts. This produces a ranked list of decoder candidates.
+4. **Stack string extraction** (`floss/stackstrings.py`): emulates functions with ≥5 MOV-to-stack instructions; monitors stack memory as strings are assembled. A context monitor captures strings passed to callees.
+5. **Tight string extraction** (`floss/tightstrings.py`): specialized emulation for tight single-block loops (≤`TS_MAX_INSN_COUNT`=10 000 instructions).
+6. **Decoded string extraction** (`floss/string_decoder.py`): emulates top-ranked decoder functions; takes memory snapshots before and after; diffs them with a binary-search algorithm (`memdiff_search`) to find newly appeared strings.
+7. **Language-specific extraction** (`floss/language/`): detects Go via `pclntab` and Rust via `.rodata` patterns; parses runtime string tables directly without emulation.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+Make sure this is configured to run before **every single commit**
+
+```
+pre-commit run --all-files
+```
+
+Pre-commit is configured in `.pre-commit-config.yaml` and runs on `pre-commit` and `pre-push` stages. It enforces:
+- `isort`: import sorting (profile: black, length-sort)
+- `black`: formatting (line length 120)
+- `mypy`: type checking (`--check-untyped-defs`)
```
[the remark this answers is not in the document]
vee1e: @gemini-code-assist no its explicitly given as `check` not `allow` in the pre commit config

AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+
+> CI runs tests on Ubuntu 22.04, Windows 2022, and macOS 14 across Python 3.10, 3.11, and 3.12. Test locally on your platform before pushing, and ensure the full matrix passes in CI before merging.
+
+## Coding Conventions
+
+### Style
+
+- **Black** with line length 120. Do not exceed this limit.
+- **isort** with `profile = black` and `length_sort = true`.
+- **Type hints are required** on all functions. mypy enforces this with `--check-untyped-defs`.
```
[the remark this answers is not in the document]
vee1e: same as above

vee1e:
@mr-tz just pinging here for a review/update when you're free

mr-tz:
ack, will take a look soon hopefully, thanks for the ping!

AGENTS.md
```diff
@@ -0,0 +1,115 @@
...
+3. **Feature extraction** (`floss/identify.py`, `floss/features/`): each function is scored on instruction features: MOV counts, XOR (non-zeroing), shifts/rotates, tight loops, call counts. This produces a ranked list of decoder candidates.
+4. **Stack string extraction** (`floss/stackstrings.py`): emulates functions with ≥5 MOV-to-stack instructions; monitors stack memory as strings are assembled. A context monitor captures strings passed to callees.
+5. **Tight string extraction** (`floss/tightstrings.py`): specialized emulation for tight single-block loops (≤`TS_MAX_INSN_COUNT`=10000 instructions).
+6. **Decoded string extraction** (`floss/string_decoder.py`): emulates top-ranked decoder functions; takes memory snapshots before and after; diffs them with a binary-search algorithm (`memdiff_search`) to find newly appeared strings.
+7. **Language-specific extraction** (`floss/language/`): detects Go via `pclntab` and Rust via `.rodata` patterns; parses runtime string tables directly without emulation.
+8. **Rendering** (`floss/render/`): outputs formatted tables (Rich), JSON, or script stubs for IDA/Ghidra/Binary Ninja/Radare2/x64dbg.
+
+## Development Setup
```
williballenthin, line 32: i sorta feel like everything after this line is verbose, and i'm not sure it's a good use of tokens. 

like rather than including the black format inline, we can just say "use black". and then the agent can pull the config if it cares. 

and with the project style... i'm inclined to keep that blank for now and add things as we need to fix them, rather than starting with an LLM-generated summary (which no offense i don't trust right now). 


i also can't prove any of this. so... it's just a vibe? feel free to push back and/or discuss.

williballenthin: even more specifically, "always invoke these pre-commit phases after each commit" and then let the agent poke around in the configs if it cares.

AGENTS.md
```diff
@@ -0,0 +1,115 @@
+# AGENTS.md
+
+## Project Overview
+
+FLOSS (FLARE Obfuscated String Solver) is a static analysis tool that automatically extracts and deobfuscates strings from malware binaries. It extends traditional string-extraction utilities (e.g., `strings.exe`) by using emulation-based techniques to recover strings that are never present in plaintext on disk.
+
+Maintained by Google/Mandiant FLARE team. Licensed under Apache 2.0.
```
mr-tz, line 7:
```suggestion
```

AGENTS.md
mr-tz: I agree and even above this is maybe irrelevant. My intention for this issue was more along the lines of giving project specific guidelines so we can avoid obvious issues in PRs like lint fails, format, adding stuff in the wrong places, etc. We should start with a minimal list here (just short bullets maybe) and extend from there when issues come up.

AGENTS.md
mr-tz: but really see below comment, would you mind creating a new draft (from scratch focusing on issues we've seen or you can think of)?

mr-tz (changes requested):

AGENTS.md
vee1e: hm alright

PR #1303, opened 2026-05-26
ci: add pinact workflow to verify pin hashes and versions

### LINKED ISSUE #1280: CI: Implement verification script for GitHub Action hashes and versions
### Problem
Recent submissions have introduced incorrect or non-existent GitHub Action versions and commit hashes in our workflows. Relying on manual review for 40-character hashes is error-prone and bypasses our security goal of using pinned, immutable references.

### Proposed Solution
Implement a verification script that programmatically validates all GitHub Action pins in `.github/workflows/`.

**Requirements:**
1. **Existence Check:** Verify that the 40-character commit hash actually exists in the remote repository.
2. **Version Consistency:** If a version tag is provided in a comment (e.g., `# v4.1.0`), verify that the hash corresponds to that specific tag.
3. **Automation:** The script should be runnable locally and as a blocking step in the PR pipeline.

### Draft Script Concept (Python)
A script using `git ls-remote` can efficiently check these constraints without cloning the remote repositories.

```python
#!/usr/bin/env python3
import re
import subprocess
import sys
from pathlib import Path

# Matches: uses: owner/repo@hash # version
ACTION_RE = re.compile(r'uses:\s+([\w\-/]+)@([a-f0-9]{40})(?:\s+#\s+(v[\d.]+))?')

def get_remote_refs(repo_path):
    url = f'https://github.com/{repo_path}.git'
    try:
        return subprocess.check_output(['git', 'ls-remote', '--tags', '--heads', url], stderr=subprocess.DEVNULL, text=True)
    except:
        return None

def verify_actions():
    workflow_dir = Path('.github/workflows')
    errors = []
    for workflow_file in workflow_dir.glob('*.yml'):
        content = workflow_file.read_text()
        for line_num, line in enumerate(content.splitlines(), 1):
            match = ACTION_RE.search(line)
            if not match: continue
            repo, commit_hash, version = match.groups()
            refs = get_remote_refs(repo)
            if refs is None or commit_hash not in refs:
                errors.append(f'{workflow_file.name}:{line_num}: Hash {commit_hash} not found in {repo}')
            elif version and not re.search(rf'{commit_hash}\s+refs/tags/{version}', refs):
                errors.append(f'{workflow_file.name}:{line_num}: Version {version} mismatch for {commit_hash}')
    if errors:
        for e in errors: print(f'[!] {e}')
        sys.exit(1)
    print('All actions verified.')

if __name__ == '__main__':
    verify_actions()
```

vee1e:
closes #1280 

The options I went through and finalized with reason @mr-tz 

- zizmor is overkill honestly, too many knobs we don't need just for a sha verifier
- pinact works and has a github action already, the `-verify` flag does exactly what the requirements demand.
- custom script is not needed and adds more tech debt

Bad action test: https://github.com/vee1e/flare-floss/pull/2

image

Good action test: https://github.com/vee1e/flare-floss/pull/3

image

Do let me know if the approach is correct or not and what to change

### DIFF
--- a/.github/workflows/verify-pins.yml
+++ b/.github/workflows/verify-pins.yml
@@ -1,4 +1,4 @@
-name: Verify Action Pins
+name: Verify GitHub Action Pins
 
 on:
   pull_request:
--- a/.github/workflows/verify-pins.yml
+++ b/.github/workflows/verify-pins.yml
@@ -15,6 +15,6 @@ jobs:
       - uses: suzuki-shunsuke/pinact-action@896d595f299e71d65b9d28349d6956abe144390a # v3.0.0
         with:
           verify: "true"
-          fix: "false"
+          fix: "true"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/verify-pins.yml
+++ b/.github/workflows/verify-pins.yml
@@ -1,0 +1,20 @@
+name: Verify Action Pins
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/**"
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: suzuki-shunsuke/pinact-action@896d595f299e71d65b9d28349d6956abe144390a # v3.0.0
+        with:
+          verify: "true"
+          fix: "false"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


vee1e:
already catching an existing issue in the workflows

should address this in a separate PR

image

williballenthin (approved):
🚀

mr-tz:
Thanks!

Should we use the fix option?

Can you address the issue as part of this PR or open a separate PR please?

vee1e:
> Should we use the fix option?

i lean to it yes.

> Can you address the issue as part of this PR or open a separate PR please?

I'll open a separate PR for it. I think this is in a mergeable state but would like a second pass.

.github/workflows/verify-pins.yml
```diff
@@ -0,0 +1,20 @@
...
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: suzuki-shunsuke/pinact-action@896d595f299e71d65b9d28349d6956abe144390a # v3.0.0
+        with:
+          verify: "true"
+          fix: "true"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mr-tz, line 20: should we use the recommended access token etc. here? per https://github.com/suzuki-shunsuke/pinact-action#how-to-use

would fix then actually work? could you maybe test in your fork?

vee1e: yup, we need to make a github app for it but it works, see https://github.com/vee1e/flare-floss/pull/2 and the workflow on https://github.com/vee1e/flare-floss/actions/runs/26532388354/job/78152761610

can't be pinned errors are there as well but will have to be fixed manually if they ever arise

mr-tz: ok, let's not get into extra trouble with the app then, are we good to go here?

mr-tz: or did you set that up? :thinking:

vee1e: > ok, let's not get into extra trouble with the app then, are we good to go here?

yeah i reverted to the normal non-modifying workflow commit

> or did you set that up? 🤔

only on my fork

.github/workflows/verify-pins.yml
```diff
@@ -0,0 +1,20 @@
+name: Verify Action Pins
```
mr-tz:
```suggestion
name: Verify GitHub Action Pins
```

mr-tz (changes requested):
then good to go, thanks

vee1e:
done!

PR #1312, opened 2026-06-02
feat: add QS ELF parser prototype

### LINKED ISSUE #785: QS: parse and render ELF structures

vee1e:
Extends QS to parse and analyze ELF executables natively, the same way it already handles PE and Mach-O files.

Closes #785 

## Summary

- Parses ELF section headers to build a structured layout of the binary
- Skips SHT_NOBITS sections (`.bss`, etc.) that have no file-backed content
- Deduplicates overlapping sections to avoid scanning the same bytes twice
- Tags strings found inside relocation sections as `#reloc`
- Tags strings decoded via XOR with `#decoded`
- Upgrades the lancelot integration from the old `Workspace` API to `BinExport2`-based instruction iteration (affects PE analysis too)

## Screenshots

Running on [`test-decode-base64`](https://github.com/mandiant/flare-floss-testfiles/blob/4c3f4bb64d1132ce98f834220986237d22a52c04/src/decode-base64/bin/test-decode-base64):

image

image

## TODOs 

- [x] ~Code range detection (`#code` tagging) is not implemented, `code_offsets` is always empty, so ELF strings are never tagged as code-region strings even if they fall inside `.text`~ Added `#code` tagging support.
- [x] The lancelot `BinExport2` change is a breaking API migration (`lancelot==0.9.7`), not just an ELF addition. Existing PE analysis behavior changes at the instruction granularity level (was basic-block-length, now per-instruction `raw_bytes`). See willi's comment here https://github.com/mandiant/flare-floss/issues/785#issuecomment-3286535023 for why it was needed.
- [x] ~ELF structure marking is minimal for now, only `.shstrtab` is registered as a named structure; other metadata sections (`.dynsym`, `.dynamic`, etc.) are not tracked.~ Now these are populated into structures as `string table`, `section header`, etc.

Credits to @FredCoast for his work on the lancelot migration and initial ELF support.

### DIFF
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -27,8 +27,9 @@ from rich.text import Text
 from rich.style import Style
 from rich.console import Console
 from elftools.elf.elffile import ELFFile
-from elftools.elf.relocation import RelocationSection
 from elftools.elf.constants import SH_FLAGS
+from elftools.elf.relocation import RelocationSection
+from elftools.common.exceptions import ELFError
 
 import floss.main
 import floss.qs.db.gp
@@ -108,8 +109,8 @@ class Slice(BaseModel):
 
     def contains_range(self, offset: int, size: int) -> bool:
         """
-        checks if this slice's buffer contains the given range,
-        where offset is relative to the start of this slice's buffer.
+        checks if this slice contains the given range,
+        where offset is relative to the start of this slice.
         """
         if not (0 <= offset <= self.range.length):
             return False
@@ -608,7 +609,7 @@ def get_reloc_offsets(slice: Slice, pe: pefile.PE) -> Set[int]:
     return ret
 
 
-def get_relocations_elf(slice: Slice, elf: ELFFile) -> List[Tuple[int, int]]:
+def get_relocations_elf(slice_: Slice, elf: ELFFile) -> List[Tuple[int, int]]:
     ranges: List[Tuple[int, int]] = []
 
     for section in elf.iter_sections():
@@ -616,11 +617,11 @@ def get_relocations_elf(slice: Slice, elf: ELFFile) -> List[Tuple[int, int]]:
             offset = section["sh_offset"]
             size = section["sh_size"]
 
-            if not slice.contains_range(offset, size):
+            if not slice_.contains_range(offset, size):
                 logger.warning("relocation directory points to an invalid location, skipping")
                 continue
 
-            ranges.append((slice.offset + offset, slice.offset + offset + size - 1))
+            ranges.append((slice_.offset + offset, slice_.offset + offset + size - 1))
     return _merge_overlapping_ranges(ranges)
 
 
@@ -813,13 +814,13 @@ class Structure(BaseModel):
     name: str
 
 
-def collect_elf_structures(slice: Slice, elf: ELFFile) -> Sequence[Structure]:
+def collect_elf_structures(slice_: Slice, elf: ELFFile) -> Sequence[Structure]:
     structures: List[Structure] = []
 
     # ELF file header: 52 bytes (32-bit) or 64 bytes (64-bit)
     header_size = 52 if elf.elfclass == 32 else 64
-    if slice.contains_range(0, header_size):
-        structures.append(Structure(slice=slice.slice(0, header_size), name="elf header"))
+    if slice_.contains_range(0, header_size):
+        structures.append(Structure(slice=slice_.slice(0, header_size), name="elf header"))
 
     # Program header table
     phoff = elf.header["e_phoff"]
@@ -827,8 +828,8 @@ def collect_elf_structures(slice: Slice, elf: ELFFile) -> Sequence[Structure]:
     phnum = elf.header["e_phnum"]
     if phnum > 0 and phentsize > 0:
         ph_total = phentsize * phnum
-        if slice.contains_range(phoff, ph_total):
-            structures.append(Structure(slice=slice.slice(phoff, ph_total), name="program header"))
+        if slice_.contains_range(phoff, ph_total):
+            structures.append(Structure(slice=slice_.slice(phoff, ph_total), name="program header"))
 
     # Section header table
     shoff = elf.header["e_shoff"]
@@ -836,8 +837,8 @@ def collect_elf_structures(slice: Slice, elf: ELFFile) -> Sequence[Structure]:
     shnum = elf.header["e_shnum"]
     if shnum > 0 and shentsize > 0:
         sh_total = shentsize * shnum
-        if slice.contains_range(shoff, sh_total):
-            structures.append(Structure(slice=slice.slice(shoff, sh_total), name="section header"))
+        if slice_.contains_range(shoff, sh_total):
+            structures.append(Structure(slice=slice_.slice(shoff, sh_total), name="section header"))
 
     # String tables (.shstrtab, .strtab, .dynstr) and symbol tables (.symtab, .dynsym)
     for section in elf.iter_sections():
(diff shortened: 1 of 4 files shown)


floss/qs/main.py
```diff
@@ -605,6 +607,23 @@ def get_reloc_offsets(slice: Slice, pe: pefile.PE) -> Set[int]:
...
+            offset = section["sh_offset"]
+            size = section["sh_size"]
+
+            if not slice.contains_range(offset, size):
+                logger.warning("relocation directory points to an invalid location, skipping")
+                continue
+
+            for fo in slice.range.slice(offset, size):
+                ret.add(fo)
+    return ret
```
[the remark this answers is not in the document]
mr-tz: along these lines please look into profiling the performance of the ELF parsing, bonus points for also covering PE and Macho

floss/qs/main.py
```diff
@@ -1178,11 +1321,12 @@ def _merge_overlapping_ranges(ranges: List[Tuple[int, int]]) -> List[Tuple[int,
     return merged_ranges
 
 
-def _get_code_ranges(ws: lancelot.Workspace, pe: pefile.PE, slice_: Slice) -> List[Tuple[int, int]]:
+def _get_code_ranges(be2, pe: pefile.PE, slice_: Slice) -> List[Tuple[int, int]]:
```
mr-tz: please use type annotations

floss/qs/main.py
```diff
@@ -1859,6 +2009,12 @@ def compute_layout(slice: Slice) -> Layout:
             # TODO: narrow exception handling once machofile error types are clearer.
             logger.debug("failed to parse as Mach-O file: %s", e)
 
+    if decoded_slice.data.startswith(b"\x7fELF"):
```
mr-tz: this doesnt make sense in the compute_pe_layout function

floss/qs/main.py
vee1e: > along these lines please look into profiling the performance of the ELF parsing, bonus points for also covering PE and Macho

i'll make an issue regarding it and raise a separate PR later before we merge

floss/qs/main.py
vee1e: Its in `compute_layout` not `compute_pe_layout` though? I think maybe github didn't show it properly in the diff

https://github.com/vee1e/flare-floss/blob/b11fc2a2b4388de75bce9dc8a5ccd3114bc0874c/floss/qs/main.py#L1962-L2012

mr-tz: oh yeah, nevermind, the diff view confused me

should these ifs be else ifs though?

vee1e: yeah, actually i'll add a handler for when the input file is none of these too

floss/qs/main.py
```diff
@@ -815,14 +816,46 @@ class Structure(BaseModel):
...
 
-    shstrtab = elf.get_section_by_name(".shstrtab")
-    if shstrtab and slice.contains_range(shstrtab["sh_offset"], shstrtab["sh_size"]):
-        structures.append(
-            Structure(
-                slice=slice.slice(shstrtab["sh_offset"], shstrtab["sh_size"]),
-                name="section header",
-            )
-        )
+    # ELF file header: 52 bytes (32-bit) or 64 bytes (64-bit)
```
mr-tz, line 820: can we reuse https://github.com/mandiant/capa/blob/master/capa/features/extractors/elf.py or parts of it here?

vee1e: i don't think its needed for the more rudimentary parsing we're doing in FLOSS. plus we've already included `machofile` and `pefile` as dependencies so this follows that trend. plus if we port features directly from capa that increases the chance of code drift

vee1e:
> The lancelot BinExport2 change is a breaking API migration (lancelot==0.9.7), not just an ELF addition. Existing PE analysis behavior changes at the instruction granularity level (was basic-block-length, now per-instruction raw_bytes). See willi's comment here https://github.com/mandiant/flare-floss/issues/785#issuecomment-3286535023 for why it was needed.

@mr-tz what are your thoughts on this? Should I refactor all the breaking lancelot code/test in this PR only?

mr-tz:
This breaking change is fine by me if it allows more functionality. We're not on a real release track here yet anyway.

vee1e:
ci failing cus https://github.com/mandiant/flare-floss-testfiles/pull/24 hasn't been merged yet. locally checked out and it works

vee1e:
tests work now had to update submodule locally and push

floss/qs/main.py
```diff
@@ -1194,22 +1338,29 @@ def get_offset_from_rva_cached(rva):
...
-            if not slice_.contains_range(offset, size):
-                logger.warning("lancelot identified code at an invalid location, skipping basic block at 0x%x", rva)
-                continue
+                if not slice_.contains_range(offset, size):
+                    logger.warning("lancelot identified code at an invalid location, skipping instruction at 0x%x", rva)
+                    continue
 
-            code_ranges.append((slice_.offset + offset, slice_.offset + offset + size - 1))
+                code_ranges.append((slice_.offset + offset, slice_.offset + offset + size - 1))
     return code_ranges
```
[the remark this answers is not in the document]
mr-tz: resolved @vee1e ?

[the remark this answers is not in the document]
vee1e: yup

vee1e:
addressed the test feedback

- finds the `.text` child and asserts it sits at the expected offset. a parsing regression in section offsets would fail this immediately.
- replaced the single truthiness check in `test_elf_structures_present` with five pinned assertions.
- `test_elf_code_and_reloc_offsets` asserts the exact merged ranges (0x10a0, 0x1265) for code and (0x568, 0x66f) for the adjacent `.rela.dyn`+`.rela.plt` pair.
- in `_make_be2_mocks`, dropped the `base_address` parameter that was accepted but never used; all three callers updated.

for the mock situation, the protobuf situation would just take the dependency introduction elsewhere, `_get_code_ranges` still calls `lancelot.be2utils.find_be2_base_address` and `lancelot.be2utils.BinExport2Index`. decided to go with dependency injection:

- `_get_code_ranges` now takes `idx` and `base_address` as explicit parameters with no lancelot calls inside. `compute_pe_layout` does those two calls and passes the results in.
- all three `_get_code_ranges` tests drop the `with patch("floss.qs.main.lancelot")` context manager entirely. `idx` from `_make_be2_mocks` and the literal `0x400000` go straight into the function call. A lancelot API change now breaks only the production code (and the test catches it), not the test infrastructure.

~CI is failing I'll fix that~ done

vee1e:
@mr-tz lemme know if there's any other changes required

tests/test_qs_elf.py
```diff
@@ -0,0 +1,70 @@
+from pathlib import Path
+
+from floss.qs.main import Slice, ELFLayout, compute_layout
+
+CD = Path(__file__).resolve().parent
+ELF_DIR = CD / "data" / "elf"
+
+# x86-64 PIE, dynamically linked, not stripped
+X86_64_PIE = "055da8e6ccfe5a9380231ea04b850e18.elf"
```
mr-tz: what does PIE mean here?

tests/test_qs_elf.py
```diff
@@ -0,0 +1,70 @@
...
+# ARM64 shared object, dynamically linked, stripped, Android linker
+ARM64_SO = "687e79cde5b0ced75ac229465835054931f9ec438816f2827a8be5f3bd474929.elf"
+# ARM64 PIE, dynamically linked, stripped
```
mr-tz, line 12: also here PIE?

mr-tz (approved):
just need some clarification on the PIE then good to go, thanks @vee1e!

tests/test_qs_elf.py
vee1e: stands for program independent code (PIE) executable, it loads into a different memory address instead of a constant like 0x400000 as a security mechanism. two testfiles are executable PIE and one is a shared object library hence the variable naming

mr-tz: ah gotcha

tests/test_qs_elf.py
```diff
@@ -0,0 +1,70 @@
+from pathlib import Path
+
+from floss.qs.main import Slice, ELFLayout, compute_layout
+
+CD = Path(__file__).resolve().parent
+ELF_DIR = CD / "data" / "elf"
+
+# x86-64 PIE, dynamically linked, not stripped
```
mr-tz:
```suggestion
# x86-64 Position Independent Executable (PIE), dynamically linked, not stripped
```

vee1e:
nice!

PR #1320, opened 2026-06-12
Catch lancelot panics during PE layout computation

mr-tz:
Lancelot (Rust library) can panic on certain malformed or complex PE files, raising pyo3_runtime.PanicException (which inherits from BaseException). Catch BaseException (excluding KeyboardInterrupt/SystemExit) to prevent crashing the entire run.


vee1e:
lgtm

williballenthin:
can you send me hashes of some of these PE files?

mr-tz:
see https://github.com/williballenthin/lancelot/issues/236

PR #1327, opened 2026-07-04
feat(qs): automate OSS string database builds

### LINKED ISSUE #796: qs: automate the construction of databases
via #761 and @r0ny123 

> For example, to build the expert db, we can use GitHub CI, to automatically add the strings from capa rules whenever a rule with a string is added/updated in the capa rules repo.

### LINKED ISSUE #1310: Create a local database build script and test it locally
Addresses 

> Perform local testing of the library build pipelines 

In Week 2

vee1e:
## Summary

Adds an automated pipeline for rebuilding the FLOSS Quantumstrand OSS string databases. The pipeline is implemented by a new Python script, `floss/qs/db/data/oss/build_oss_db.py`, and a new GitHub Actions workflow, `.github/workflows/build-oss-db.yml`, that runs the script on a bi-weekly cron (the 1st and 15th of each month at 00:00 UTC) and on manual dispatch.

- Closes #796 
- Closes #1310

## What the script does

The script automates the existing vcpkg plus jh technique described in `floss/qs/db/data/oss/readme.md`:

1. Install static libraries via vcpkg for a chosen triplet.
2. Extract string features and function names from each .lib using jh.
3. Convert the jh output to JSONL and compress with gzip.

Per library, it produces one `.jsonl.gz` file in the output directory, matching the schema consumed by `floss.qs.db.oss.OpenSourceStringDatabase`.

Two correctness fixes are baked into the converter:

- Within-library dedup. One metadata object per unique string in a library, matching the loader's `metadata_by_string` semantics.

The script also merges into any `.jsonl.gz` already present in the output directory rather than replacing them. New entries win on string collisions, which is the right behavior when a library version changes, and libraries that are not in the current `--libraries` list are still re-checked against the cross-library dedup and rewritten if the shared string set changes. This makes it safe to run the script locally with `--libraries <one-lib>` without losing the other databases.

## What the workflow does

`.github/workflows/build-oss-db.yml`:

- Runs on a bi-weekly cron (1st and 15th of each month at 00:00 UTC) and on `workflow_dispatch`.
- Checks out flare-floss and lancelot, sets up Python 3.12 and Rust stable.
- Builds jh from lancelot (`cargo build --release -p lancelot-bin`).
- Runs the build script with `--continue-on-error` so a single failed library does not block the others.
- Opens a PR against `quantumstrand` with the regenerated `.jsonl.gz` files if anything changed.

The auto-update PR step is what makes this end-to-end. The script plus workflow is closed-loop: any change in upstream vcpkg package versions surfaces as a database diff, which becomes a small PR with metrics.

### Errors in pipeline

- [x] `ncurses`, `s2n` and `liboqs` are only available on windows, i'm considering removing this one because its doubtful it'll appear in windows samples
- [x] `clBLAS` is hitting a rust [assertion error](https://github.com/vee1e/flare-floss/actions/runs/28712689489/job/85148757414#step:9:109)
- [x] LFS isn't being detected at all, they're [reading the pointers as text lmeow](https://github.com/vee1e/flare-floss/actions/runs/28712689489/job/85148757414#step:9:305)

### Code review

- [x] dependency handling in https://github.com/mandiant/flare-floss/pull/1327#discussion_r3523672572
- [x] caching in https://github.com/mandiant/flare-floss/pull/1327#discussion_r3523674462
- [x] documenting and discussing tradeoffs of chosen triplet/compilers in https://github.com/mandiant/flare-floss/pull/1327#discussion_r3523676296
- [x] move the script to `scripts/` and add copyright


scripts/build_oss_db.py
```diff
@@ -0,0 +1,973 @@
...
+                    {
+                        "string": value,
+                        "library_name": library,
+                        "library_version": version,
+                        "file_path": file_path,
+                        "function_name": value,
+                        "line_number": None,
+                    }
+                )
+                explicit_function_names.add(value)
```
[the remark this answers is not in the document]
vee1e: @gemini-code-assist the PR description was outdated, `jh` emits jsonl now.

floss/qs/db/data/oss/libraries.json
```diff
@@ -0,0 +1,73 @@
...
+    "libimobiledevice",
+    "libarchive",
+    "libmysql",
+    "libhv",
+    "cello",
+    "czmq",
+    "boost-test",
+    "wolfssl",
+    "boost-wave",
+    "curl",
```
williballenthin, line 23: isn't zlib a dep of curl, so does that mean the zlib strings get filter as duplicate, rather than being assigned to zlib?

how do you handle this dependency issue?

.github/workflows/build-oss-db.yml
```diff
@@ -0,0 +1,109 @@
...
+          restore-keys: |
+            ${{ runner.os }}-cargo-jh-
+
+      - name: Cache vcpkg
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: |
+            C:\vcpkg\installed
+            C:\vcpkg\downloads
+          key: ${{ runner.os }}-vcpkg-oss-db-${{ hashFiles('flare-floss/floss/qs/db/data/oss/build_oss_db.py') }}
```
williballenthin: seems like this keys the hash on the generation script, not the contents of the libraries. am i reading that right?

williballenthin: i think what we'd want to use is a vcpkg lock file, if that exists

floss/qs/db/data/oss/libraries.json
```diff
@@ -0,0 +1,73 @@
+{
+  "triplet": "x64-windows-static",
+  "compiler": "msvc143",
```
williballenthin, line 3: how is this chosen and when should this be updated?

williballenthin: what effect, if any, does it have on the compilation?

floss/qs/db/data/oss/readme.md
```diff
@@ -94,3 +94,19 @@ These files are then gzip'd:
 ```console
 $  gzip -c zlib.jsonl > zlib.jsonl.gz
 ```
+
+The `build_oss_db.py` script automates the steps above and additionally
+removes strings that appear in two or more libraries from every database.
+A shared string is not a unique identifier for any specific library, so
```
williballenthin: as noted above: unless one library links or vendors another library.

floss/qs/db/data/oss/readme.md
```diff
@@ -94,3 +94,19 @@ These files are then gzip'd:
 ```console
 $  gzip -c zlib.jsonl > zlib.jsonl.gz
 ```
+
+The `build_oss_db.py` script automates the steps above and additionally
```
williballenthin: i like that the logic lives in this script, and that the GH workflow is minimal. this reduces our dependence on GH, which i believe is prudent.

williballenthin (commented):
i think every 2 weeks is a reasonable starting point. if this is too noisy, we can reduce the frequency. 

ultimately, i like that this demonstrates how to rebuild the databases, rather than having the databases be some magic artifact that can't be reproduced. i'm not convinced they need to be rebuilt so often, but i don't think it hurts.

floss/qs/db/data/oss/libraries.json
vee1e: vcpkg builds curl and zlib as separate static .lib files so `jh` only sees zlib's symbols when processing zlib.lib directly. When processing libcurl.lib, it only extracts strings from curl's own compiled code. it doesn't re-extract zlib's symbols just because curl _links_ against it.

you can also see a very minimal string overlap in my local fork: https://github.com/vee1e/flare-floss/pull/9

> 367: [...zlib.jsonl.gz (149 entries, 8 removed as shared with other libraries)](https://github.com/vee1e/flare-floss/actions/runs/28712689489/job/85148757414#step:9:367)

.github/workflows/build-oss-db.yml
vee1e: > seems like this keys the hash on the generation script, not the contents of the libraries. am i reading that right?

yup, not perfect ofc, i was thinking of using the libraries.json config file as the hash so it updates whenever we add/remove a lib to build

> i think what we'd want to use is a vcpkg lock file, if that exists

the requirement for that is fulfilled by the libraries.json file already imo

I think we can drop `C:\vcpkg\installed` from the cache because vcpkg names files in downloads/ by their content hash (e.g. `openssl-3.4.1.tar.gz` with a sha512 in the filename). When vcpkg installs a package it checks downloads/ first, then if the tarball is already there and the hash matches, it skips the network fetch and uses it directly.

floss/qs/db/data/oss/libraries.json
vee1e: > how is this chosen and when should this be updated?

taken directly from the existing qs documentation

https://github.com/mandiant/flare-floss/blob/0fbe890606316a8726335daf6372e59c512bd795/floss/qs/db/data/oss/readme.md?plain=1#L48-L52

> what effect, if any, does it have on the compilation?

i tried the builds locally on macosx-arm64. no significant differences in string counts was observed

vee1e:
> i think every 2 weeks is a reasonable starting point. if this is too noisy, we can reduce the frequency.

me and @mr-tz had a few back and forths with this. I ran a few tests on vcpkg versions and found out that ~biweekly is the best to capture around 5-8 update events because libraries update on different schedules.

vcpkg update times (in days) that we found:

image

- Mean of per-library averages (78 libs): 305 days (~10 months)
- Median of per-library averages: 181 days (~6 months)
- Pooled across all individual gaps (814 gaps from 87 libs): 157 days (~5.2 months) — this weights by how often each library actually releases

but I still do agree, i think we can get a better estimate after letting this cook for a while, so the frequency is volatile for now. we should also think about how many update events we should/need to capture — weekly would give us 3-4, monthly 10-15 from my tests.

vee1e:
Ran the [workflow on my own fork](https://github.com/vee1e/flare-floss/actions/runs/28712689489/job/85148757414) — 118ish minutes for the whole thing to complete

floss/qs/db/data/oss/build_oss_db.py
```diff
@@ -0,0 +1,977 @@
...
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+)
+logger = logging.getLogger("build_oss_db")
+
+
+# Fallback library list when neither --libraries nor libraries.json is
+# provided. The actual production list lives in libraries.json.
+DEFAULT_LIBRARIES: List[str] = [
```
mr-tz: Are these necessary or helpful here?

mr-tz (commented):
Nice work. Further discussion in the threads Willi opened is needed.

floss/qs/db/data/oss/build_oss_db.py
vee1e: yeah i don't think so. removing it.

floss/qs/db/data/oss/libraries.json
mr-tz: Missing these would be bad

floss/qs/db/data/oss/libraries.json
mr-tz: What builds did you try?

floss/qs/db/data/oss/libraries.json
williballenthin: agree that when the dependency is via linking, then things may work out. but in some cases the dependency may be vendored, ie, source code copied into the project. in this case, i think the strings will be detected as duplicates. 

to be clear, i believe this is a hard problem to handle generically. though maybe it can also be solved by a little research and hacks/configs/shorcuts.

.github/workflows/build-oss-db.yml
williballenthin: ah ok so it's not a cache of the built artifacts, but of the source downloads. 

i still think this cache key isn't right, and probably confusing as is  (maybe just a comment is needed).

floss/qs/db/data/oss/libraries.json
williballenthin: i probably picked 143 at the time because it was probably the latest version or something. i assume there's not a major difference between versions, but we should document our expectations and assumptions. that's why i raise it here.

williballenthin:
> vcpkg update times (in days) that we found

imho a more interesting metric is how frequently does an important string get added to the db. project releases don't necessarily imply meaningful additions to the databases. 

likewise, there's probably some study that could be done about strings getting removed. 

and then, a scenario where strings are added and then removed without that middle release getting captured (such as a zlib version string or something). would a more comprehensive approach therefore be to build every release? what are the pros/cons there?

we don't necessarily need to do all this research, just raising some ideas.

floss/qs/db/data/oss/libraries.json
vee1e: > What builds did you try?

Using raw strings on both I got this:

| Library       |  macOS |  MinGW | In Both | mac Only | Win Only |
| ------------- | -----: | -----: | ------: | -------: | -------: |
| binn          |    212 |    401 |       3 |      209 |      398 |
| brotli        |  2,768 |  3,281 |   1,087 |    1,681 |    2,194 |
| bzip2         |    571 |    632 |      63 |      508 |      569 |
| curl          |  5,996 |  6,869 |   1,972 |    4,024 |    4,897 |
| flatcc        |    751 |  1,085 |      87 |      664 |      998 |
| fmt           |    719 |  1,246 |      59 |      660 |    1,187 |
| freetype      |  4,938 |  5,985 |     954 |    3,984 |    5,031 |
| libarchive    |  5,328 |  9,169 |   1,446 |    3,882 |    7,723 |
| libflac       |  2,349 |  3,775 |     382 |    1,967 |    3,393 |
| libjpeg-turbo |  4,000 |  3,786 |     320 |    3,680 |    3,466 |
| liblzma       |  1,182 |  1,653 |     159 |    1,023 |    1,494 |
| libogg        |    226 |    269 |      81 |      145 |      188 |
| libpng        |  1,909 |  2,111 |     436 |    1,473 |    1,675 |
| libtheora     |  1,171 |  1,690 |      39 |    1,132 |    1,651 |
| libvorbis     |  3,273 |  4,796 |     620 |    2,653 |    4,176 |
| libwebp       |  3,185 |  4,803 |     117 |    3,068 |    4,686 |
| libxml2       |  8,506 |  9,200 |   3,874 |    4,632 |    5,326 |
| lz4           |    540 |  1,290 |      28 |      512 |    1,262 |
| lzo           |    604 |    811 |      13 |      591 |      798 |
| mbedtls       |  6,931 |  7,592 |   2,509 |    4,422 |    5,083 |
| openal-soft   |  7,862 | 17,976 |   2,093 |    5,769 |   15,883 |
| openssl       | 53,464 | 43,900 |  13,148 |   40,316 |   30,752 |
| opus          |  2,729 |  3,224 |     581 |    2,148 |    2,643 |
| opusfile      |    284 |    565 |      14 |      270 |      551 |
| parson        |    245 |    343 |       5 |      240 |      338 |
| physfs        |  1,202 |  1,588 |      76 |    1,126 |    1,512 |
| sqlite3       |  8,335 |  7,518 |   1,221 |    7,114 |    6,297 |
| tre           |    330 |    595 |      34 |      296 |      561 |
| zlib          |    674 |    828 |      52 |      622 |      776 |
| zstd          |  2,157 |  5,250 |     133 |    2,024 |    5,117 |


```
=== binn macOS-only (arm64 machine code) ===
  '\t@9j'           ← random bytes, not text
  ' #[z9'           ← ditto
  '@9+\t@9,'        ← ditto
  'K)@)'            ← ditto
  'R)#@'            ← ditto
  '_AddValue'       ← symbol name (mangled C function)
  '_AdvanceDataPos' ← symbol name
  '__.SYMDEF SORTED'← macOS BSD ar archive metadata
  '#1/12  1782384003  501  20  644  57180  `'
                     ← macOS BSD ar archive metadata

=== binn mingw-only (x86_64 machine code) ===
  '\tw}A'           ← random bytes
  '([^_]A\A]A^A_'   ← fragments of x86_64 instructions
  '8[^_]A\A]A^A_'   ← ditto
  'AWAVAUATUWVSH'   ← x86_64 function prologue bytes
  'AVWVSH'          ← ditto
  '/  1783197601  0  0  0  0  2520  `'  ← GNU ar archive metadata
```

So there is a difference in strings but the total count remains similar. **I didn't look at which strings overlapped before, hence the slightly wrong conclusion that there was no difference.** But most of the lack of overlap can be attributed to:

1. Random printable byte runs in compiled code that look like strings to `strings -n 4`. Different ISAs produce different random byte runs, so the noise never overlaps.
2. Section names differ. Mach-O has __TEXT, __cstring etc. while COFF has .rdata, .pdata, .xdata
3. Symbol name prefixes are different too. Mach-O adds a leading _ (_AddValue), COFF doesn't (AddValue)

It could be worth building another db set for different platforms if we want to cover all possible executables, but it'll increase build times and cost. I think its a fair point to debate imo

.github/workflows/build-oss-db.yml
vee1e: > i still think this cache key isn't right, and probably confusing as is (maybe just a comment is needed).

yeah we can add comments on why the decision was made. maybe we can forgo the cache entirely?

floss/qs/db/data/oss/libraries.json
vee1e: instead of removing duplicates altogether should we be upfront to the user and tell them exactly how commonly a string was found? maybe if its truly unique we can indicate it as a "pure" indicator tag, and with every subsequent instance of the string being found we can somehow show the library tag (#zlib etc) as "less" of a true oracle for its source.

vee1e:
> imho a more interesting metric is how frequently does an important string get added to the db. project releases don't necessarily imply meaningful additions to the databases.

> likewise, there's probably some study that could be done about strings getting removed.

yeah i did some preliminary research on this for a few libs. results are here:

Library | Transition | +Strings | -Strings | +Funcs | -Funcs
-- | -- | -- | -- | -- | --
zlib | 1.3.1 → 1.3.2 | 5 | 3 | 10 | 1
expat | 2.4.8 → 2.6.2 | 2 | 1 | 2 | 0
expat | 2.6.2 → 2.7.3 | 8 | 3 | 13 | 2
expat | 2.7.3 → 2.8.2 | 7 | 3 | 7 | 0
lz4 | 1.9.3 → 1.9.4 | 3 | 1 | 36 | 5
lz4 | 1.9.4 → 1.10.0 | 4 | 2 | 20 | 80
brotli | 1.0.9 → 1.1.0 | 42 | 43 | 29 | 4
brotli | 1.1.0 → 1.2.0 | 8 | 12 | 13 | 0
libpng | 1.6.45 → 1.6.53 | 30 | 32 | 19 | 15
libpng | 1.6.53 → 1.6.58 | 4 | 3 | 0 | 0
sqlite3 | 3.34.1 → 3.42.0 | 248 | 158 | 229 | 106
sqlite3 | 3.42.0 → 3.48.0 | 98 | 39 | 107 | 39
sqlite3 | 3.48.0 → 3.53.2 | 72 | 21 | 51 | 16
mbedtls | 2.24.0 → 2.28.5 | 86 | 67 | 1 | 1
mbedtls | 2.28.5 → 3.6.1 | 571 | 110 | 151 | 52
mbedtls | 3.6.1 → 3.6.5 | 36 | 22 | 8 | 1
libxml2 | 2.11.8 → 2.14.6 | 1062 | 853 | 269 | 198
libxml2 | 2.14.6 → 2.15.3 | 224 | 124 | 36 | 197
curl | 8.11.0 → 8.21.0 | 910 | 581 | 655 | 345
openssl | 1.1.1j → 3.0.8 | 3845 | 825 | 5218 | 544
openssl | 3.0.8 → 3.6.3 | 13023 | 987 | 4404 | 183


the almost equal share of `+` and `-` strings in each case indicate some stuff is probably just being updated with time (like `Copyright 2026` → `Copyright 2027` instead of new features being added. but its not 100% clear and needs more research on what strnigs are being changed instead of just how many.

> and then, a scenario where strings are added and then removed without that middle release getting captured (such as a zlib version string or something). would a more comprehensive approach therefore be to build every release? what are the pros/cons there?

yeah hence my point that we shouldn't wait **too long** to build because of the missing middle, and we shouldn't wait **too little** because of wasted build time for not a lot of new information being gained.

floss/qs/db/data/oss/libraries.json
mr-tz: The comment here was on the msvc version :) interesting additional information though. With our PE focus we should stick to Windows for now.

.github/workflows/build-oss-db.yml
williballenthin: i think that's fine. if we end up spending obviously a lot of computer resources maybe we can optimize some, but let's not do that prematurely.

floss/qs/db/data/oss/libraries.json
mr-tz: Keeping duplicates around may be the safer bet for now until we see more of this in practice or do more research into the overlaps.

vee1e:
- fixed LFS where gzip reads were failing on the pointer text by adding `lfs: true`
- dropped Cargo + vcpkg caching from the workflow since the rebuild cadence is slow enough that the cold start is fine. as discussed in https://github.com/mandiant/flare-floss/pull/1327#discussion_r3525525772
- Stopped removing cross-library duplicates. The vendored-dep case (zlib copied into curl) was silently dropping legitimate zlib indicators. Now every entry is stamped with `count` = number of libraries the string appears in via a new `OpenSourceString.count` field. strings stay in every DB they appear in. `count=1` = unique, >=2 means shared. as discussed https://github.com/mandiant/flare-floss/pull/1327#discussion_r3527809013
- the loader is backward compatible, msgspec defaults `count` to 1, so legacy `.jsonl.gz` files keep decoding. Preserved (non-rebuilt) libraries are only rewritten when their counts actually change, so the workflow doesn't produce spurious diffs.
- renamed `test_oss_db.py` to `test_qs_oss_db.py` so `pytest -k qs` picks it up alongside the other qs tests.

mr-tz:
What benefit does the added count give? Can't this just be deducted from the databases easily? When tagging strings we currently show multiple tags already.

vee1e:
Actually yeah the current tagger is fine as-is, I'll remove the count var. Reverted.

.github/workflows/build-oss-db.yml
```diff
@@ -0,0 +1,88 @@
...
+  pull-requests: write
+
+jobs:
+  build-databases:
+    name: Build OSS string databases
+    runs-on: windows-latest
+
+    env:
+      # Workaround for zydis 3.1.3's CMakeLists.txt requiring an older
+      # cmake_minimum_required syntax. Remove once Lancelot updates zydis.
```
mr-tz, line 27: is there an issue for that in Lancelot? if not can you please add one?

mr-tz: and reference here

floss/qs/db/data/oss/libraries.json
```diff
@@ -0,0 +1,69 @@
...
+  "profile": "release",
+  "libraries": [
+    "openssl",
```
mr-tz: can we sort these A-Z or what's the order meaning here?

floss/qs/db/data/oss/readme.md
```diff
@@ -94,3 +94,19 @@ These files are then gzip'd:
...
+removes strings that appear in two or more libraries from every database.
+A shared string is not a unique identifier for any specific library, so
+attributing it would produce false-positive library matches at query time.
```
mr-tz: outdated

floss/qs/db/data/oss/readme.md
```diff
@@ -94,3 +94,19 @@ These files are then gzip'd:
...
+
+The script also merges into any `.jsonl.gz` already present in the output
+directory rather than replacing them. For each library being rebuilt, the
+newly-extracted entries are combined with the existing entries (new entries
+win on string collisions, which is the right behavior when the library
+version changes). Libraries that are not in the current `--libraries` list
+are also re-checked against the cross-library dedup and rewritten if the
```
mr-tz: outdated?

floss/qs/db/data/oss/libraries.json
vee1e: they were ordered based on number of descending strings found. i'll just sort them alphabetically.

scripts/build_oss_db.py
```diff
@@ -0,0 +1,867 @@
...
+
+  1. install static libraries via vcpkg
+  2. extract string features (and function names) via jh
+  3. convert to JSONL and compress with gzip
+
+It is intentionally modular so the underlying extractor (jh today) can be
+swapped for a more minimal tool later without rewriting the orchestration.
+
+Strings are NOT deduped across libraries: a string observed in both zlib
+and curl (e.g. when zlib is vendored into curl) stays in both databases.
```
mr-tz, line 29: remove or worth keeping?

.github/workflows/build-oss-db.yml
vee1e: yeah there is one, i'll make an issue regarding it

scripts/build_oss_db.py
```diff
@@ -0,0 +1,867 @@
...
+The query tagger already emits one ``#<library>`` tag per matching
+database, so the consumer can see the overlap directly. Within a single
+library, the same string still collapses to one entry (when
+``--no-deduplicate`` is not passed).
```
mr-tz, line 33: is this current?

vee1e: yeah `--no-deduplicate` only changes intra-lib strings.

scripts/build_oss_db.py
```diff
@@ -0,0 +1,867 @@
...
+    def extract(
+        self,
+        lib_path: pathlib.Path,
+        library: str,
+        version: str,
+        triplet: str,
+        compiler: str,
+        profile: str,
+    ) -> str:
+        """Run jh on a single static library and return its CSV output."""
```
mr-tz: is it still csv output?

vee1e: nope, outdated comment, fixing

floss/qs/db/data/oss/readme.md
vee1e: yup

scripts/build_oss_db.py
vee1e: don't think so, removed

vee1e:
Summary of what was added in this round per the code review by @mr-tz

The build script now has:

- The `write_entries()` dead code function completely removed
- One `make_db_entry(...)` helper at module top, used at the 4 sites that previously built the dict inline.
- The converter parse function tallies objects and function names in the same pass that builds entries, no separate `count_jsonl_rows` call.

Also the test script, now updated:

- For the `Converter` parsing:
   - unique object/function counts (incl. duplicate-row, missing-function, `path=None`)
   - dedup on/off behavior
   - `emit_function_names` on/off
   - explicit `function_name` rows not duplicated as synthetic
   - malformed JSONL lines skipped with warning, empty lines skipped.
- For the `load_existing_entries` function:
   - round-trip with `make_db_entry`
   - missing file returns []
   - malformed lines skipped with warning
   - non-dict rows and rows missing `"string"` skipped
   - unknown keys ignored, missing keys filled with `None`
   - empty file handled

floss/qs/db/data/oss/libraries.json
mr-tz: worth documenting these findings anyhwere else?

vee1e: I'll add few lines in the readme for it for future reference

mr-tz:
some more auto feedback

Unneeded / dead code

- Iterable imported from typing (:255) — never used.
- field imported from dataclasses (:256) — never used; none of the dataclasses use field(...).
- run()'s env parameter (:346) — plumbed through but every call site (vcpkg.install, vcpkg.list, jh.extract, cargo build) omits it. Either drop it or use it (e.g. for VCPKG_ROOT/CARGO_* overrides).
- import time inside build_library() (:689) — should be a top-level import like everything else; there's no lazy-loading justification here.
- defaults["libraries"] = config.get("libraries",) (:873-875) — a trailing comma on its own line for a call with a single positional arg. It's just config.get("libraries") dressed up oddly; looks like a leftover from an edit, not intentional.

Correctness risk worth a look

- Vcpkg._find_all_static_libs() (:479) is the fallback used when the .list file can't be located — it scans the entire triplet lib/ directory and returns every .lib/.a file present, not just the ones belonging to the library in question. Since vcpkg installs all dependencies' static libs into the same shared lib/ folder, this fallback would silently attribute another package's strings/functions to the wrong library. Given it's only a "fallback," it may rarely trigger, but it's a silent correctness bug when it does — worth at least logging which files it's including so bad merges are traceable.
- main()'s exit code logic (:1043-1047): when --continue-on-error is passed (which the workflow always does) and every library fails, the script still returns 0. The workflow step will show green, and create-pull-request will just find "no changes" — a total pipeline failure (e.g. vcpkg broken, jh path wrong) could go unnoticed except by someone manually reading build_metrics.jsen successful == 0, or wiring a separate CI check on the metricssummary.

Tests

Good unit coverage for Converter.parse, load_existing_entries, and make_db_entry, but there's no test for merge_entries (cross-vs-within-library merge precedence) or for the main() orchestration (merged/existing-fil), which is exactly where the cross-library dedup question abovewould have been caught by a test.

vee1e:
Thanks for the review. All eight items verified as true against the current code.

### Dead code removed
- `Iterable` (typing) and `field` (dataclasses) imports.
- `run(env=)` parameter, none of the four call sites pass it. the `merged_env` block inside `run()` is gone too.
- `import time` inside `build_library()` moved to the top.
- `config.get("libraries",)` trailing-comma removed.

### Correctness
- `Vcpkg._find_all_static_libs` fallback no longer silently scans the whole `lib/` dir. The `.list` file is the only authoritative source for which libs belong to a package; when it's missing we now log a strong error listing the candidate files we deliberately skipped and return `[]` so the library is treated as "no static libs found" rather than being polluted with another package's strings/functions.
- `main()` exit code: with `--continue-on-error`, return `1` when every library failed so a broken pipeline (vcpkg down, wrong jh path, etc.) doesn't silently show a green workflow step. Partial success still exits 0 as before.

### Tests
- Added 12 new tests
  - 5 for `merge_entries` (new-wins on collision, dedup off, both empty, single-side copies)
  - 7 for `main()` orchestration with stubbed `Vcpkg`/`JHExtractor`/`build_library` (per-library output, merge with pre-existing DB, pre-existing libraries not in the current run are preserved, three exit-code paths, `build_metrics.json` content).

34/34 passing.

The build-parameters notes from the thread above also landed in `floss/qs/db/data/oss/readme.md` under a new "Why these build parameters" section.

mr-tz:
last inconsistencies between code and PR description, please clean this up and then we can merge 🥳 

1. Cross-library dedup — still unresolved (carried over from last review)
Code and docstrings (scripts/build_oss_db.py:245, :702, :890-894) explicitly say strings are not deduped across libraries, while the PR description still claims the opposite ("removed from every database ... correctness fix"). Neither the code nor the PR text was touched here — this needs to be settled one way or the other, since it's the stated fix for issues #796/#1310.

2. New finding — "Caches Cargo and vcpkg" isn't in the workflow
The PR body says under "What the workflow does": "Caches Cargo and vcpkg." I grepped the entire diff for cache (case-insensitive) and got zero hits — there is no actions/cache step, no VCPKG_BINARY_SOURCES setup, nothing. I confirmed via search that windows-latest runners do come with vcpkg preinstalled (VCPKG_ROOT/VCPKG_INSTALLATION_ROOT both point at C:\vcpkg), so the script will find vcpkg fine — that part isn't broken. But without any binary cache, every biweekly run will rebuild ~60 vcpkg ports (including boost-*, openssl, poco, sdl2) and rebuild jh via cargo build --release from scratch, every single time. That's a real cost/duration concern for a "closed-loop automated" job, and it doesn't match what's documented. Either add the caching (vcpkg's GHA binary cache + actions/cache for ~/.cargo and target/), or correct the PR description.

vee1e:
those were just old PR description artifacts. updated the desc removin the outdated information

mr-tz:
Great work here, thanks!

.github/workflows/build-oss-db.yml
vee1e: - https://github.com/williballenthin/lancelot/issues/240

Issue #37, opened 2016-04-08
potential issue: "global stackstrings" won't be detected


williballenthin:
malware that uses a stackstrings-like technique to initialize a global string will not be detected, since we currently inspect only the active stack frame. jay's script supports this type of decoding, though i can't think of an example off the top of my head.

williballenthin:
need to dev a test case, and then can tackle the feature

williballenthin:
yes. example attached. 

a single byte XOR routine is inlined by the compiler into a constant number of DWORD and BYTE-wise XOR operations against global data. this is difficult to detect (no loops) using the "floss" algorithms, but global stackstrings would catch this.

![keylogger-s-new](https://cloud.githubusercontent.com/assets/156560/15525699/523cfd2e-21fa-11e6-96ff-70919914aba2.gif)

williballenthin:
91b08896fbda9edb8b6f93a6bc811ec6

amancslab:
I implemented support for detecting strings constructed character-by-character in global memory.

The implementation uses Vivisect’s existing memory-write log during stack-string emulation to identify written non-stack regions, extract completed strings, and report them as global `DecodedString` results.

Main implementation PR:

mandiant/flare-floss#1335

Companion test-data PR:

mandiant/flare-floss-testfiles#27

Testing results:

* Linux: passed
* Windows 32-bit: passed
* Windows 64-bit: passed
* Additional string, loading, and rendering regression tests: passed

PR #1335, opened 2026-07-13
Detect strings written to global memory

amancslab:
## Summary

Detect strings constructed character-by-character in global memory during stack-string emulation.

The stack-string emulator already enables Vivisect memory-write logging. This change uses the write log to:

* identify non-stack addresses written during function emulation
* group adjacent written addresses into memory regions
* extract strings from the completed global-memory regions
* report them as `DecodedString` results with `AddressType.GLOBAL`
* preserve the existing `extract_stackstrings()` API
* avoid duplicate decoded-string results when the existing decoder pipeline finds the same string

This resolves the global stackstrings case described in #37.

## Testing

```bash
python -m pytest tests/data/src/decode-global-stackstrings/test.yml --runxfail -vv
```

Result:

```text
3 passed
```

The test passed for:

* Linux
* Windows 32-bit
* Windows 64-bit

Additional regression tests:

```bash
python -m pytest tests/test_strings.py tests/test_load.py tests/test_render.py -vv
```

Result:

```text
5 passed
```

Formatting hooks passed:

* isort
* black
* `git diff --check`

The mypy hook still reports the existing Windows-specific `mmap` typing errors in `floss/utils.py:617`; no new typing errors are introduced by this change.

Companion test-data PR: mandiant/flare-floss-testfiles#27

### DIFF
--- a/floss/main.py
+++ b/floss/main.py
@@ -71,7 +71,7 @@ from floss.identify import (
     get_functions_without_tightloops,
 )
 from floss.logging_ import TRACE, DebugLevel
-from floss.stackstrings import extract_stackstrings
+from floss.stackstrings import extract_stack_and_global_strings
 from floss.tightstrings import extract_tightstrings
 from floss.string_decoder import decode_strings
 from floss.language.identify import Language, identify_language_and_version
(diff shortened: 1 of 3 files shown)


vee1e:
master already has this with `AddressType.GLOBAL` in string_decoder and results, so vote to close this

cc: @mr-tz

vee1e:
Additionally consider closing the associated issue as well

Issue #943, opened 2024-02-02
Extend FLOSS to use the rendering techniques pioneered by QUANTUMSTRAND


mr-tz:
[QUANTUMSTRAND](https://github.com/mandiant/flare-floss/tree/quantumstrand/floss/qs) is an experiment that augments traditional strings.exe output with context to aid in malware analysis and reverse engineering. For example, we show the structure of a file alongside its strings and mute/highlight entries based on their global prevalence, library association, expert rules, and more.

FLOSS should use the techniques pioneered by QUANTUMSTRAND to highlight important information while muting common and/or analytically irrelevant noise.

Anu-Ra-g:
@mr-tz Can I setup floss on WSL2? Or flare-vm is preferred?

williballenthin:
no preference, as long as the environment is comfortable for you to develop Python

andy0712dik:
@mr-tz i would like to contribute to this project because it is looking interesting I am final-year student and my interest is in cybersecurity can you please guide me that where should i start to get involve in this project

mr-tz:
A good start would be to familiarize yourself with the tool, documentation, and source. Then you can browse the open issue list and/or ask focused questions.

officialasishkumar:
Is there any slack or element channel?

danishzulfiqar:
Can you assign issue to me, I can start getting understanding of it

williballenthin:
@danishzulfiqar no, this is a large-scoped issue tracking a project candidate for GSoC'24. If you'd like to contribute there, please follow our posted guidance on how to get started.

PR #1337, opened 2026-07-16
refactor: split QUANTUMSTRAND into first-class layout/tags modules

### LINKED ISSUE #1213: Refactor proposal: deduplicate language “coverage” CLI boilerplate (Go/Rust)

### LINKED ISSUE #1313: QS: Lints and formatters do not have full code coverage
More "housekeeping" related. Currently `isort`, `black` and `mypy` all fail on the pre-commit stage for Quantumstrand regardless of the changes pushed due to mostly unrelated files upstream not being formatted in the expected manner. This results in ~35 LOC of pure formatting being modified each time we attempt to push a feature.

This makes contributing especially annoying since we have to pass `--no-verify` every time which could result in false negatives.

Examples:

```diff
diff --git a/floss/qs/main.py b/floss/qs/main.py
index d87f5b9..d2f4d75 100644
--- a/floss/qs/main.py
+++ b/floss/qs/main.py
@@ -1,11 +1,11 @@
 import io
-import struct
 import re
 import abc
 import sys
 import json
 import time
 import bisect
+import struct
 import hashlib
 import logging
 import pathlib
@@ -18,9 +18,9 @@ from typing import Set, Dict, List, Tuple, Literal, Callable, Iterable, Optional
 from collections import defaultdict

 import pefile
-import machofile
 import colorama
 import lancelot
+import machofile
 import rich.traceback
 from pydantic import Field, BaseModel, ConfigDict
 from rich.text import Text
@@ -907,24 +907,19 @@ def collect_pe_structures(slice: Slice, pe: pefile.PE) -> Sequence[Structure]:
                         )
                     )

-    if hasattr(pe, 'RICH_HEADER') and pe.RICH_HEADER:
+    if hasattr(pe, "RICH_HEADER") and pe.RICH_HEADER:
         key_bytes = pe.RICH_HEADER.key

-        rich_sig_offset = pe.__data__.find(b'Rich', 0x40, pe.DOS_HEADER.e_lfanew)
+        rich_sig_offset = pe.__data__.find(b"Rich", 0x40, pe.DOS_HEADER.e_lfanew)
         # The structure end is 'Rich' (4) + key (4) = 8 bytes
         rich_end = rich_sig_offset + 8

         # Find the start of rich header by looking for 'DanS' XORed with the key
-        xor_dans = bytes(a ^ b for a, b in zip(b'DanS', key_bytes))
+        xor_dans = bytes(a ^ b for a, b in zip(b"DanS", key_bytes))
         rich_start = pe.__data__.rfind(xor_dans, 0x40, rich_sig_offset)

         if rich_sig_offset != -1 and rich_start != -1:
-            structures.append(
-                Structure(
-                    slice=slice.slice(rich_start, rich_end - rich_start),
-                    name="rich header"
-                )
-            )
+            structures.append(Structure(slice=slice.slice(rich_start, rich_end - rich_start), name="rich header"))

     return structures

@@ -1456,9 +1451,7 @@ def _parse_fat_arches(data: bytes) -> List[Tuple[str, int, int]]:
         else:
             if offset + 20 > len(data):
                 break
-            cputype, cpusubtype, arch_offset, size, _align = struct.unpack(
-                endian + "IIIII", data[offset : offset + 20]
-            )
+            cputype, cpusubtype, arch_offset, size, _align = struct.unpack(endian + "IIIII", data[offset : offset + 20])
             offset += 20

         arch_name = _format_macho_arch(cputype, cpusubtype)
@@ -1591,9 +1584,7 @@ def _attach_nested_layout(parent: Layout, child: Layout):
         parent.add_child(child)


-def _parse_superblob_blobs(
-    slice_: Slice, cs_offset: int, cs_size: int
-) -> Sequence[Tuple[int, int, int]]:
+def _parse_superblob_blobs(slice_: Slice, cs_offset: int, cs_size: int) -> Sequence[Tuple[int, int, int]]:
     blobs: List[Tuple[int, int, int]] = []
     if cs_size <= 0:
         return blobs
@@ -2090,8 +2081,11 @@ def render_strings(
                         is_group_start = True

             line = render_string(
-                console.width, string, tag_rules,
-                prev_tags=prev_tags, prev_tags_width=prev_tags_width,
+                console.width,
+                string,
+                tag_rules,
+                prev_tags=prev_tags,
+                prev_tags_width=prev_tags_width,
                 is_group_end=is_group_end,
                 is_group_start=is_group_start,
             )
```

vee1e:
## Summary

Remove the floss/qs product package and qs CLI/binary packaging. Move layout analysis into floss/layout, string tagging into floss/tags (with DBs still under floss/qs/db/data), and extract FLOSS argparse into floss/cli. Promote former [qs] dependencies to core install deps.

- Closes #1213 
- Closes https://github.com/mandiant/flare-floss/issues/1313
- Addresses https://github.com/mandiant/flare-floss/issues/943

See for the previous discussion:
* https://github.com/vee1e/flare-floss/pull/18


floss/cli.py
```diff
@@ -0,0 +1,327 @@
...
+        logging.getLogger("viv_utils.emulator_drivers").setLevel(logging.DEBUG)
+    elif debug <= DebugLevel.TRACE:
+        logging.getLogger("viv_utils.emulator_drivers").setLevel(logging.ERROR)
+
+    # install the log message colorizer to the default handler.
+    # because basicConfig is just above this,
+    # handlers[0] is a StreamHandler to STDERR.
+    #
+    # calling this code from outside script main may do something unexpected.
+    logging.getLogger().handlers[0].setFormatter(floss.logging_.ColorFormatter())
```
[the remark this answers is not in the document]
vee1e: resolved in https://github.com/mandiant/flare-floss/pull/1337/commits/e353743e4178b1aa0714f2240ee805acb8e3f4b5

floss/layout/macho.py
```diff
@@ -0,0 +1,497 @@
...
+                    if section_offset + section_size > cmd_end:
+                        break
+                    structures.append(
+                        Structure(slice=slice_.slice(section_offset, section_size), name="section header")
+                    )
+                    section_offset += section_size
+
+        offset += cmdsize
+
+    return structures, segments, code_sig
```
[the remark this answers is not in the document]
vee1e: resolved in https://github.com/mandiant/flare-floss/pull/1337/commits/e353743e4178b1aa0714f2240ee805acb8e3f4b5

floss/layout/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+    ]
+
+    xor_key = None
+    decoded_slice = slice_
+
+    # Try to find the XOR key
+    for mz, key in mz_xor:
+        if slice_.data.startswith(mz):
+            xor_key = key
+            break
```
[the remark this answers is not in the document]
vee1e: resolved in https://github.com/mandiant/flare-floss/pull/1337/commits/e353743e4178b1aa0714f2240ee805acb8e3f4b5

.github/pyinstaller/floss.spec
```diff
@@ -110,8 +141,8 @@ exe = EXE(
     console=True,
 )
 
-# enable the following to debug the contents of the .exe
+# enable the following to debug the contents of the .exe
 # writes to ./dist/floss-dat
 #coll = COLLECT(
 #    exe, a.binaries, a.zipfiles, a.datas, strip=None, upx=True, name="floss-dat"
-#)
+#)
```
mr-tz, line 148: whats the diff here?

.github/mypy/mypy.ini
```diff
@@ -1,5 +1,6 @@
 [mypy]
 plugins = pydantic.mypy
+mypy_path = .
```
mr-tz, line 3: what does this do?

floss/layout/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        logger.debug("unrecognized file format, falling back to binary layout")
+
+    return SegmentLayout(
+        slice=slice_,
+        name="binary",
+    )
+
+
+__all__ = [
```
mr-tz: we don't use this in this project (yet) but could consider it
I'm not sure about the re-exports (e.g. MIN_STR_LEN) here and what the best approach is

floss/layout/base.py
```diff
@@ -0,0 +1,295 @@
...
+    section: Optional[pefile.SectionStructure] = None
+
+
+class SegmentLayout(Layout):
+    """region not covered by any section, such as PE header or overlay"""
+
+    pass
+
+
+class PELayout(Layout):
```
mr-tz, line 196: maybe a nit but should these go into the respective pe/elf/macho files?

floss/layout/types.py
mr-tz, line 1: does this belong into layout? not immediately clear to me

floss/qs/readme.md
mr-tz, line 1: is this information or parts of it persisted somewhere else now?

floss/tags/oss.py
```diff
@@ -1,11 +1,34 @@
...
 from typing import Dict, Sequence
 from dataclasses import dataclass
 
 import msgspec
 
-import floss.qs.db
+
+def _data_root() -> pathlib.Path:
+    """Legacy DB location: floss/qs/db/data."""
+    return pathlib.Path(__file__).resolve().parents[1] / "qs" / "db" / "data"
```
mr-tz: same definition here in various files, reuse one?

floss/main.py
```diff
@@ -508,12 +241,17 @@ def main(argv=None) -> int:
...
     if argv is None:
         argv = sys.argv[1:]
 
+    if argv and argv[0] == "quantum":
+        from floss.quantum import main as quantum_main
+
+        return quantum_main(argv[1:])
+
+    # use rich as default Traceback handler
+    rich.traceback.install(show_locals=True)
```
mr-tz: why down here now?

scripts/disassemblers/render-binja-import-script.py
```diff
@@ -22,7 +22,7 @@
 Usage:
 
   $ floss suspicious.exe -j > floss-results.json
-  $ python render-binja-import-script.py floss-results.json > apply_floss.py
+  $ python scripts/disassemblers/render-binja-import-script.py floss-results.json > apply_floss.py
```
mr-tz: would be fine to not include the full path for cases like this, the example is clear

scripts/tags/generate_gp_db.py
```diff
@@ -16,9 +28,14 @@
 import argparse
 import collections
 from typing import Dict, Tuple
+from pathlib import Path
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
```
mr-tz, line 35: hacky?

tests/conftest.py
```diff
@@ -14,8 +14,13 @@
 
 
 import os
+import sys
 from pathlib import Path
 
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
```
mr-tz: hacky? needed?

tests/test_gp_db.py
```diff
@@ -1,20 +1,34 @@
...
-import floss.qs.db.gp
+import floss.tags.gp
 
 CD = pathlib.Path(__file__).parent
 
 
 def test_load_db():
-    path = pathlib.Path(floss.qs.db.gp.__file__).parent / "data" / "gp" / "gp.jsonl.gz"
-    db = floss.qs.db.gp.StringGlobalPrevalenceDatabase.from_file(path)
+    path = pathlib.Path(floss.tags.gp.__file__).resolve().parents[1] / "qs" / "db" / "data" / "gp" / "gp.jsonl.gz"
```
mr-tz: this looks a little wild :)

tests/test_scripts.py
```diff
@@ -30,7 +30,7 @@
 
 
 def get_script_path(s) -> Path:
-    return CD / ".." / "scripts" / s
+    return CD / ".." / "scripts" / "disassemblers" / s
```
mr-tz: only testing those scripts?

mr-tz (commented):
some comments and points for further discussion

.github/pyinstaller/floss.spec
vee1e: its a CRLF -> LF conversion

- old: 23290d0a
- new: 23290a

.github/mypy/mypy.ini
vee1e: it adds the repo root `.` to mypy’s module search path. since scripts/ is typechecked now and uses package-style imports, without this, mypy only searches installed packages or the files it’s given, and those `scripts.*` imports often don’t resolve from the way pre-commit invokes mypy.

floss/layout/base.py
vee1e: fair but right now `base.py` is the layout type hierarchy and pe/elf/macho are the builders. SectionLayout and SegmentLayout are shared across formats, so they belong in base, and the format-specific subclasses could live next to their parsers. i’d rather leave that as maybe a follow-up refactor so this PR stays about the modular split/integration rather than reshuffling class homes.

floss/layout/types.py
vee1e: to me the rationale was `Tag` and `TaggedString` living here because `Layout.strings` holds them through extract -> tag and putting them under tags risked import cycles during the split.

also `ExtractedString` is layout-native (slice-backed); the tagged form is shared with tags/document.

floss/qs/readme.md
vee1e: it was removed mostly. only the database maintenance side was rewritten into scripts/tags/README.md (and the existing per-DB readmes under floss/qs/db/data/). The product overview, feature lists and screenshots were not moved anywhere. could be good to put as an addendum at the top of the main readme.md now the quantum is a first class feature. what do you think?

floss/tags/oss.py
vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/d03761423d4e1c140c26709b92c81391e4b1d1c4, put it in `tags/__init__.py`

floss/main.py
vee1e: only moved relative to the new early return for floss quantum. Classic FLOSS still installs rich at the start of its path. qs never hits this line and uses `rich.traceback.install()` in quantum.py instead.

scripts/disassemblers/render-binja-import-script.py
vee1e: makes sense, addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/1e2ba0b

scripts/tags/generate_gp_db.py
vee1e: agree it’s a bit hacky, but it’s only so `python scripts/tags/generate_gp_db.py` works without PYTHONPATH needing to be set because of scripts.tags.* imports after the reorg. we can drop it and document something like `python -m scripts.tags.generate_gp_db` or require `-e .` instead. i prefer the simpler user facing way but am open to opinions

tests/conftest.py
vee1e: same idea as above, needed only so tests can import scripts.tags...

scripts isn’t installed with the package. we can add something like

```py
# pyproject.toml
[tool.pytest.ini_options]
pythonpath = ["."]
```

instead of mutating sys.path for the whole suite.

tests/test_gp_db.py
vee1e: yup, addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/d93b6c4 similar to https://github.com/mandiant/flare-floss/pull/1337/commits/d03761423d4e1c140c26709b92c81391e4b1d1c4

tests/test_scripts.py
vee1e: yeah, this file only ever smoke-tested the five render import scripts. the path update is just for the reorg into disassemblers/. Tag DB and analysis scripts aren’t in this suite (e.g. `test_qs_build_oss_db.py` is for the OSS build testing). we can rename the helper/test to something like `test_disassembler_scripts` to make the scope clearer.

floss/layout/__init__.py
vee1e: yeah maybe we can add a clarifying docstring to it?

.github/pyinstaller/floss.spec
mr-tz: ah good

floss/layout/base.py
mr-tz: sounds good, thanks

floss/layout/types.py
mr-tz: ok, let's keep and iterate on it if we have to

floss/qs/readme.md
mr-tz: yes, we should keep the relevant info around

floss/tags/oss.py
mr-tz: thanks

floss/main.py
mr-tz: ah can that be simplified (track for later?) or will it be superseded anyway then?

tests/conftest.py
mr-tz: hm ok yeah whatever is clearer and easier to maintain, what do you think?

tests/test_scripts.py
mr-tz: that works

floss/qs/readme.md
vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/dfeb21c i transposed the missing stuff to a section in the main README

floss/main.py
vee1e: i think it'll be fine to simplify. restored it and added a new traceback install in quantum.py in https://github.com/mandiant/flare-floss/pull/1337/commits/018ef3d.

tests/conftest.py
vee1e: i think we can implement the pyproject change to reduce the hackiness a bit. implemented in https://github.com/mandiant/flare-floss/pull/1337/commits/0adb6bc

tests/test_scripts.py
vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/adda278

floss/layout/__init__.py
mr-tz: One of the last things to figure out here before merge.

scripts/tags/generate_gp_db.py
mr-tz: One of the last things to figure out here before merge.

floss/layout/__init__.py
vee1e: yeah true, what do you feel should be done for this?

scripts/tags/generate_gp_db.py
vee1e: sure, what path do you feel would be best here?

floss/layout/__init__.py
mr-tz: It's not a blocker for me but do you feel this is setting future usage and maintainability up best?

scripts/tags/generate_gp_db.py
mr-tz: This should be runnable as a script so if no better solution exists we can keep as is.

vee1e: yeah agreed. lets keep it easily runnable

floss/layout/__init__.py
vee1e: agreed on the long term part. i'd rather set a thin package: only compute_layout on floss.layout, drop `__all__` and import types / MIN_STR_LEN / extract helpers from base / extract / types (which most internal code already does).

the broad re-exports were convenience from the QS split, not a deliberate public API. we can add `__all__` later on if needed.

vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/d350f785a0ec1bde029b4fd8ae7f8d8abc854442

mr-tz: Thanks, that's good. What about the other usages of all or was this the only one left?

vee1e: there's one in `tags/` as well. addressed in https://github.com/mandiant/flare-floss/pull/1337/commits/f9210691f4e54de23e3e92a8b69b059d8e6746f5

mr-tz: thanks

mr-tz:
Nice work here towards integration of FLOSS and QuantumStrand, @vee1e!

PR #1347, opened 2026-07-26
feat: unify layout/tags pipeline into default floss (iteration 2)

### LINKED ISSUE #1295: Use the QS view as default FLOSS view (for static strings)
We should discuss if we want to merge QS into FLOSS.

I think it could simplify some things but also create some confusion/issues (at least initially).

For static strings I can see the QS view as the default but we'd have to see how deobfuscated strings are then shown.

I've got much more thoughts but just recording this here for tracking purposes.

### LINKED ISSUE #1336: QS: move db contents to tags folder
Part of #943 
Context: https://github.com/vee1e/flare-floss/pull/18#discussion_r3594792033

Refactor the current `qs/db` folder's remaining files incl. scripts, docs, dbs to the more appropriate `tags/` folder. 

Should be implemented post https://github.com/vee1e/flare-floss/pull/18

vee1e:
## Summary

Iteration 2 of the QUANTUMSTRAND → FLOSS productization: one analysis path, one `ResultDocument`, and layout-aware static strings by default.

- Closes https://github.com/mandiant/flare-floss/issues/1336
- Closes https://github.com/mandiant/flare-floss/issues/1295
- Addresses https://github.com/mandiant/flare-floss/issues/943

See the full plan at https://gist.github.com/vee1e/c0afe9950394839b8a9ac37d4174f97e

- Extend `floss.results` with plain `tags` / `section` / `structure` fields and an optional serializable `layout` tree
- Move orchestration into `floss.pipeline.analyze`; thin `floss.main` entry
- When PE/ELF/Mach-O layout succeeds, reuse quantum extract (tags, structures, FP-lib rules) for static strings; classic statics as fallback
- Enrich static + language only (stack / tight / decoded remain unenriched for now)
- Default text: quantum-style section tree with **no classic meta** when layout is present; recovered string sections follow at the end
- CLI: `--no-layout` / `--no-tags`
- Move tag DBs `floss/qs/db/data` → `floss/tags/data` (LFS preserved); remove residual `floss/qs`
- Update PyInstaller datas, OSS workflow paths, packaging, README, and bulk_analyze
- Note: ~280 ish changes are pure whitespace/newline changes that were swept up in this PR.

## Sample output

On `test-decode-to-stack.exe` (truncated)

image

<br>

On `go-hello.exe` (truncated)

image

## Test plan

- [x] `pytest tests/` — 264 passed, 2 skipped, 6 xfailed
- [x] Smoke: `floss --no-stack tight decoded` on PMA 01-01 → layout tree with tags
- [x] Smoke: `floss --no-layout` → classic static listing
- [x] Smoke: `-j` includes `layout`, `tags`, `section`/`structure` on statics
- [x] CI on this PR (tests.yml, mypy as configured)


.github/pyinstaller/floss.spec
```diff
@@ -101,18 +101,18 @@ excludes = [
...
     ["../../floss/main.py"],
     pathex=["floss"],
     binaries=[],
     datas=datas,
-    hiddenimports=quantum_hiddenimports,
+    hiddenimports=quantum_hiddenimports,
     hookspath=[".github/pyinstaller/hooks"],
     runtime_hooks=[],
-    excludes=excludes,
+    excludes=excludes,
```
mr-tz, line 113: What are the EOL changes? All to use LF only?

README.md
```diff
@@ -43,16 +43,18 @@ The strings FLOSS extracts specific to a compiler are much easier to inspect by
...
+expert rules, and more). Stack, tight, and decoded strings still appear after
+the layout-aware static listing when deobfuscation is enabled.
 
 ```console
-$ floss quantum sample.exe
-$ floss quantum sample.exe -j
+$ floss sample.exe
+$ floss sample.exe -j
+$ floss --no layout sample.exe    # classic static listing
+$ floss --no tags sample.exe      # layout without tag databases
```
mr-tz: I don't really like this mix of string type and output mode. It's confusing.

mr-tz: Let's discuss options first on if/how to handle. I think a new option would make sense but also am hesitant to add unneeded new options. Maybe a bigger todo to clean up the cli arguments.

mr-tz: Or do you think this option works well as is?

floss/document.py
```diff
@@ -12,97 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Interim layout-aware result document for static string analysis.
+"""Deprecated: layout types live in ``floss.results``.
```
mr-tz: I think remove these now?!

floss/main.py
```diff
@@ -13,176 +13,46 @@
...
+    Options,
+    PipelineError,
+    analyze,
+    load_vw,
+    get_signatures,
+    select_functions,
+)
 
+# re-exports for tests and plugins that import helpers from floss.main
+__all__ = [
```
mr-tz: Remove again and don't add moving forward.

floss/main.py
```diff
@@ -208,54 +78,66 @@ def get_default_root() -> Path:
...
 
     if argv is None:
         argv = sys.argv[1:]
 
+    # deprecated alias: floss quantum … → same pipeline with static-focused defaults
     if argv and argv[0] == "quantum":
-        from floss.quantum import main as quantum_main
-
-        return quantum_main(argv[1:])
+        logger.warning("'floss quantum' is deprecated; layout-aware static analysis is the default for floss")
```
mr-tz: I don't think we need to keep this around. It's not been used much/at all.

floss/main.py
```diff
@@ -208,54 +78,66 @@ def get_default_root() -> Path:
...
+        # rebuild as: --no stack tight decoded [quantum flags] path
+        mapped = ["--no", "stack", "tight", "decoded"]
+        # translate quantum-style path argument (no FileType wrapping issues if we pass through make_parser)
+        i = 0
+        path = None
+        min_length = None
+        want_json = False
+        quiet = False
+        debug = 0
+        while i < len(rest):
```
mr-tz: This will be removed I assume but otherwise should be done via argparse.

floss/quantum.py
mr-tz, line 1: Again, vote to remove this as not needed.

tests/test_layout.py
```diff
@@ -47,64 +48,57 @@ def analyzed_layout(pma_binary_path):
...
+    finally:
+        path.unlink()
+
+    assert one.metadata.file_path == two.metadata.file_path
+    assert one.layout is not None and two.layout is not None
+    assert one.layout.name == two.layout.name
+    assert len(one.strings.static_strings) == len(two.strings.static_strings)
+    assert one.strings.static_strings[0].tags == two.strings.static_strings[0].tags
 
-    # show the round trip works
```
mr-tz, line 64: Keep comments

tests/test_qs.py
```diff
@@ -47,64 +48,57 @@ def analyzed_layout(pma_binary_path):
...
-    assert one.meta.version != three.meta.version
-    assert one != three
-    assert one.model_dump_json(exclude_none=True) != three.model_dump_json(exclude_none=True)
+    three.metadata.version = "0"
+    assert one.metadata.version != three.metadata.version
 
 
 def test_string_extraction(analyzed_layout):
     strings = collect_strings(analyzed_layout)
-    # Check if a known string is extracted
```
mr-tz, line 81: Keep comments

tests/test_qs.py
```diff
@@ -47,64 +48,57 @@ def analyzed_layout(pma_binary_path):
...
     strings = collect_strings(analyzed_layout)
-    # Check if a string is correctly associated with a structure
     data_string = next(s for s in strings if s.string.string == "@.data")
     assert data_string.structure == "section header"
 
     close_string = next(s for s in strings if s.string.string == "CloseHandle")
     assert close_string.structure == "import table"
-
-
-def test_analysis_pipeline(pma_binary_path):
```
mr-tz, line 102: Keep?

vee1e: Yeah putting it back

floss/main.py
vee1e: The CI was failing without these, I'll find a workaround for it.

.github/pyinstaller/floss.spec
vee1e: Yeah either LF changes or whitespace stripping.

README.md
vee1e: We can make an issue to refine and address CLI args. This was just a poc and feedback is definitely needed.

vee1e: - https://github.com/mandiant/flare-floss/issues/1348

floss/document.py
vee1e: yup, done in https://github.com/mandiant/flare-floss/pull/1347/commits/96437babd152f2e6c11b759666d4da1f72e74d64

floss/main.py
vee1e: done, just had to change some imports https://github.com/mandiant/flare-floss/pull/1347/commits/c090b9404cc1b0192183764a51a8bc8ad1306b60

floss/main.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1347/commits/e06e5d70636c5bb610e590465f49c4b300a71faf

floss/quantum.py
vee1e: also done in https://github.com/mandiant/flare-floss/commit/e06e5d70636c5bb610e590465f49c4b300a71faf

tests/test_layout.py
vee1e: done https://github.com/mandiant/flare-floss/pull/1347/commits/dd0211f1b987a6ee68cca117fe113c7c5b2c9219

tests/test_qs.py
vee1e: done https://github.com/mandiant/flare-floss/pull/1347/commits/dd0211f1b987a6ee68cca117fe113c7c5b2c9219

tests/test_qs.py
vee1e: done https://github.com/mandiant/flare-floss/pull/1347/commits/cda9f33ed33b05f98e37ec45e68f74926e1de27b

floss/main.py
vee1e: yeah, removed in https://github.com/mandiant/flare-floss/pull/1347/commits/e06e5d70636c5bb610e590465f49c4b300a71faf. this was just a temp workaround before removing `quantum` entirely, but argparse should be used in the future for this work.

floss/render/default.py
```diff
@@ -363,17 +379,18 @@ def render(results: floss.results.ResultDocument, verbose, disable_headers, colo
...
 
-    if results.analysis.enable_stack_strings:
+    # recovered strings always after static/language (classic blocks)
+    if results.analysis.enable_stack_strings and results.strings.stack_strings:
         render_heading(f"FLOSS STACK STRINGS ({len(results.strings.stack_strings)})", console, verbose, disable_headers)
         render_stackstrings(results.strings.stack_strings, console, verbose, disable_headers)
         console.print("\n")
 
-    if results.analysis.enable_tight_strings:
+    if results.analysis.enable_tight_strings and results.strings.tight_strings:
```
mr-tz: right now we at least show 0 strings were extracted for this mode if selected

floss/results.py
```diff
@@ -173,6 +252,8 @@ class Runtime:
     stack_strings: float = 0
     decoded_strings: float = 0
     tight_strings: float = 0
+    layout: float = 0
+    tags: float = 0
```
mr-tz: worth tracking? maybe for the rendering step? granular enough?

tests/test_layout.py
```diff
@@ -47,33 +48,42 @@ def analyzed_layout(pma_binary_path):
...
+    with tempfile.NamedTemporaryFile("w", suffix=".json", delete=False) as f:
+        f.write(doc)
+        path = Path(f.name)
+    try:
+        two = ResultDocument.parse_file(path)
+    finally:
+        path.unlink()
 
     # show the round trip works
-    # first by comparing the objects directly,
```
mr-tz, line 65: keep all comments that make sense

vee1e: restored most of the ones i feel were needed in https://github.com/mandiant/flare-floss/pull/1347/commits/c1e027ef31915e778c4374877c22163405bc4954 and https://github.com/mandiant/flare-floss/pull/1347/commits/75c973a59fd1d9b0638b67dcc70a6876fb6c1ff2

floss/results.py
vee1e: **removed them both in https://github.com/mandiant/flare-floss/pull/1347/commits/75c973a59fd1d9b0638b67dcc70a6876fb6c1ff2 for now. i feel metrics and perf tracking is an open question and can be discussed further in future iterations, and shouldn't be a gate in this PR's scope but the next one should address it.**

floss/render/default.py
vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1347/commits/63fab3fe7d2c987f4fa27ef72fa6ec77a6d533ca, now we show the headers even if no strings were found

README.md
```diff
@@ -43,16 +43,18 @@ The strings FLOSS extracts specific to a compiler are much easier to inspect by
...
 Please consult the documentation to learn more about the [language-specific string extraction](doc/language_specific_strings.md).
 
-### Layout-aware strings (`floss quantum`)
+### Layout-aware static strings
 
-QUANTUMSTRAND-style analysis is available as the `floss quantum` subcommand. It
-augments traditional `strings.exe` output with context to aid malware analysis
-and reverse engineering: file structure next to strings, and mute/highlight of
-entries based on global prevalence, library association, expert rules, and more.
+For PE, ELF, and Mach-O samples, FLOSS enriches static strings by default with
```
mr-tz: Tags are also applied independently of file type.

README.md
mr-tz: For here let's add a temporary different arg and use the open issue to discuss further changes after that.

mr-tz: Please suggest options

floss/results.py
```diff
@@ -98,6 +101,7 @@ class StackString:
     original_stack_pointer: int
     offset: int
     frame_offset: int
+    tags: List[str] = field(default_factory=list)
```
mr-tz: Not needed (yet) - remove

floss/results.py
```diff
@@ -130,6 +134,9 @@ class DecodedString:
     encoding: StringEncoding
     decoded_at: int
     decoding_routine: int
+    tags: List[str] = field(default_factory=list)
```
mr-tz: Same: not needed yet

scripts/analysis/bulk_analyze.py
```diff
@@ -134,22 +153,31 @@ def main():
...
+                        "Failed to analyze file %s, exited with code %d",
+                        file_path,
+                        result.returncode,
+                    )
                     if result.stdout:
                         logger.error("stdout:\n%s", result.stdout)
                     if result.stderr:
                         logger.error("stderr:\n%s", result.stderr)
             except Exception as e:
-                logger.error("Failed to run analysis subprocess for file %s: %s", file_path, e, exc_info=True)
```
mr-tz, line 143: Are these line reformats necessary in this file esp?

mr-tz (commented):
Some more notes, good progress

scripts/analysis/bulk_analyze.py
vee1e: no, i think black ran with an incorrect config and this snuck in. fixed in https://github.com/mandiant/flare-floss/pull/1347/commits/03ee960612f26808fbb434af62fd85b7fe206bed

floss/results.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1347/commits/374262dbe43116b704bc969003bc4d3ef6d917dd

README.md
vee1e: fixed in https://github.com/mandiant/flare-floss/commit/1b05672

README.md
vee1e: for this PR a temporary `--no-layout` and `--no-tags` could work, defaults stay the same. what do you think?

floss/results.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1347/commits/374262dbe43116b704bc969003bc4d3ef6d917dd too

mr-tz:
LLM review:

Bugs / correctness

1. CI smoke test still calls the removed floss quantum subcommand — .github/workflows/build.yml:44-48 runs ./dist/floss quantum --help and ./dist/floss quantum ... -j | grep -q '#common', but this PR deletes the quantum argv dispatch from main.py and removes the quantum console-script entry from pyproject.toml. This job didn't fire on this PR (it only triggers on pull_request: branches: [master], and this PR targets quantumstrand), so it's silently broken and will fail the instant this lands on master. Needs a fix in this PR, not after.
2. Layout fallback promise is broken for exceptions past compute_layout — floss/pipeline.py:233-248 (_try_layout_static) only wraps compute_layout() in try/except. extract_layout_strings, live.tag_strings, live.mark_structures, load_databases(), and remove_false_positive_lib_strings all run unguarded. Since layout+tags are now default-on for every PE/ELF/Mach-O sample (previously only reachable via opt-in floss quantum), any exception in these later stages — a malformed section triggering a structure-marking bug, a corrupt/missing tag DB — now crashes the whole floss run instead of falling back to classic static strings, contradicting the function's own docstring ("returns None to fall back to classic statics"). This is the biggest behavioral risk from making the feature default.
3. Stale PyInstaller hidden-imports — .github/pyinstaller/floss.spec:21-28 still lists "floss.quantum" and "floss.document" in quantum_hiddenimports; both modules are deleted by this PR. The comment above it ("not imported by floss/main.py yet") is also now inaccurate — pipeline.py does import floss.layout/floss.tags (lazily, hence still needing the hidden-imports hint for the rest of the list).

Gaps

4. No test coverage for the two new CLI switches — --no layout / --no tags are the headline new behavior toggles, but tests/test_cli_args.py wasn't extended with any case for them (it only picked up an import change). Worth a couple of analyze()/CLI-level tests asserting --no layout yields classic statics and --no tags yields an untagged layout tree.
5. Duplicate file read on every default run — _try_layout_static calls sample.read_bytes() (floss/pipeline.py:230ish), re-reading the entire file that get_static_strings() already read moments earlier in analyze(). Previously this cost was paid only by opt-in floss quantum users; now it's paid by everyone, every run. Cheap to fix (pass the buffer through) and worth doing given perf/metrics tracking was explicitly punted to a future iteration in the review thread.

Nits / follow-up docs

6. floss/enrich.py:27 (layout_encoding_to_string_encoding) has a dead "utf-8" branch — floss/layout/extract.py's ExtractedString.encoding is Literal["ascii", "unicode"] only, so that case can never fire. Harmless but misleading.
7. Leftover pre-rename references the mechanical qs/db → tags/data sweep missed: scripts/README.md:8 still says "Build and maintain QuantumStrand tag databases", and floss/tags/data/oss/readme.md:87 still hard-codes an example path under floss/qs/db/data/oss/zlib.csv.

README.md
mr-tz: yes that works temporarily at least

floss/render/default.py
```diff
@@ -335,19 +345,25 @@ def render(results: floss.results.ResultDocument, verbose, disable_headers, colo
...
-            colored_str = heading_style(f"FLARE FLOSS RESULTS (version {results.metadata.version})\n")
-            console.print(colored_str)
-        render_meta(results, console, verbose)
-        console.print("\n")
-
-    if results.analysis.enable_static_strings:
-        render_staticstrings(results.strings.static_strings, console, verbose, disable_headers)
-        console.print("\n")
+    # layout-aware path: no classic meta table (quantum-style)
+    if results.layout is not None and results.analysis.enable_static_strings:
```
mr-tz, line 363: need to check logic here

README.md
vee1e: done https://github.com/mandiant/flare-floss/pull/1347/commits/b4c1c623853f01ddf6275a8a1f47720222bc0b09

vee1e:
addressed llm review in https://github.com/mandiant/flare-floss/pull/1347/commits/47715efaf086e63691b5d8ad03437f337d04189c

floss/render/default.py
vee1e: Only one small edge case is there. The flag `--load` does not clear `results.layout` or apply `enable_layout`. So a JSON saved with layout still renders layout UI even if we pass it with `--no-layout`. we can address this if its a blocker but if/when we deprecate the default view entirely in subsequent PRs, I don't think its a huge issue.

mr-tz: Yeah let's note as a todo/issue though so we keep it in munde.

vee1e: done https://github.com/mandiant/flare-floss/pull/1347/commits/77bb42ec96b5f27e224c26f1f3dd8213fe5cb974

mr-tz:
almost there, another LLM review before I do my final pass

---

New finding: incorrect runtime accounting on the (now-default) layout-success path

floss/pipeline.py, _try_layout_static call site in analyze():

if layout_doc is not None:
    results.layout = layout_doc
    results.strings.static_strings = static_strings_from_layout(layout_doc)
    # includes layout/tag work when structured layout succeeded
    results.metadata.runtime.static_strings = get_runtime_diff(time0)
else:
    results.strings.static_strings = static_strings
    results.metadata.runtime.static_strings = static_runtime

time0 is stamped before the initial classic string extraction. By the time this line runs, the following have all happened in between: classic extraction, identify_language_and_version, and — critically — the interactive input("Do you want to enable string deobfuscation? ...") prompt when stdout is a TTY and no --no/--only was passed. So get_runtime_diff(time0) bakes an open-ended human response time into a field labeled static_strings. Previously (pre-refactor) this field was always the tightly-scoped static_runtime measured right after extraction; the else-branch here still does that correctly, only the layout-success branch regressed.

- Blast radius: this is the default code path now (layout succeeds for any parseable PE/ELF/Mach-O). It doesn't show up in the default text UI (that branch skips render_meta entirely when results.layout is not None), but it does leak into -j JSON output's metadata.runtime.static_strings, which is presumably consumed by tooling/telemetry given the perf-tracking discussion already in this PR's thread.
- Fix: snapshot a timestamp immediately before calling _try_layout_static and report static_runtime + get_runtime_diff(that_timestamp), instead of diffing from time0.

Already-tracked, worth a one-line mention

floss/render/default.py's new layout branch keys off results.layout is not None only; floss.results.load() doesn't clear results.layout or honor enable_layout/enable_tags on --load. You already flagged this in review and the author added a TODO(#1348) comment plus filed the follow-up — no action needed, just noting it's real and reproducible (floss --load saved_with_layout.json --no-layout still renders the layout view).

Nits

- Doc/PR-description drift: the PR description says floss quantum "becomes a deprecated alias," but the actual diff deletes floss/quantum.py, floss/document.py, and the quantum = "floss.quantum:main" console-script entirely, with no alias or deprecation message. I see from the review thread this was an intentional, discussed decision (you voted to just remove it) — but as written, the PR description is now stale and worth a quick edit before merge so it doesn't mislead anyone skimming the changelog. Practically: anyone still typing floss quantum sample.exe gets swallowed by the positional sample argparse arg and a confusing can't open 'quantum': No such file or directory error rather than any guidance.
- floss/enrich.py: enrich_static_strings() rebuilds the full offset index (build_offset_index, a recursive tree walk) on every call, and it's called twice per language branch (once for language_strings, once for language_strings_missed). Cheap to hoist into a single index built once in analyze().
- _try_layout_static's catch-all logs only at logger.debug. Given layout+tags is now the headline default behavior, a user hitting a real bug there gets silently downgraded to classic statics with zero visible signal unless running with -d. Consider logger.warning so silent feature loss is at least noticeable.
- get_static_strings_from_bytes (floss/utils.py) logs "File is empty" — capitalized, inconsistent with the surrounding lowercase log-message convention (e.g., "file is very large..." two functions away).

floss/main.py
```diff
@@ -312,268 +134,45 @@ def main(argv=None) -> int:
...
-    time0 = time()
-    interim = time0
+    try:
+        analysis_results = analyze(options)
+    except PipelineError as e:
+        if e.exit_code in (1, 130):
+            logger.info("%s", e)
+        else:
+            logger.error("%s", e)
+        return e.exit_code
```
mr-tz, line 163: I forgot, did we discuss this? what is the use here?

floss/utils.py
```diff
@@ -602,6 +602,16 @@ def read_memory(vw, va: int, size: int) -> bytes:
     raise envi.exc.SegmentationViolation(va)
 
 
+def get_static_strings_from_bytes(buf: bytes, min_length: int) -> list:
```
mr-tz: is this needed?

scripts/analysis/bulk_analyze.py
```diff
@@ -85,7 +85,11 @@ def main():
             cmd = [
                 sys.executable,
                 "-m",
-                "floss.quantum",
+                "floss.main",
+                "--no",
+                "stack",
+                "tight",
+                "decoded",
```
mr-tz, line 94: the bulk script used to work differently running regular floss?! now it just does static?

tests/test_qs.py
```diff
@@ -47,33 +48,49 @@ def analyzed_layout(pma_binary_path):
...
+        path.unlink()
 
     # show the round trip works
-    # first by comparing the objects directly,
-    # which works thanks to pydantic model equality.
-    assert one == two
-    # second by showing their json representations are the same.
-    assert one.model_dump_json(exclude_none=True) == two.model_dump_json(exclude_none=True)
+    # first by comparing key fields after JSON serialize/parse
+    # (full ResultDocument equality is not used: nested Runtime timestamps and
```
mr-tz: what changed here making this necessary?

scripts/analysis/bulk_analyze.py
vee1e: this script is inherited from QS, not master. since qs layout is static only the functionality is preserved here by excluding stack/tight/decoded. we can mark this to be changed when we add deobf strings as first-class in qs output.

tests/test_qs.py
vee1e: JSON isn't identical anymore because of float coercion in the fields and a `runtime.start_date` being timezone-aware now.

floss/utils.py
vee1e: mostly yes, its so we do `read_bytes` on sample once, and get statics and layout from the buffer. inlining could be done but would be worse for readability in my opinion. what do you think?

floss/main.py
vee1e: yeah this is so that cli failures that were handled in `main` are handled here and raised as a more descriptive PipelineError.

vee1e:
2nd llm review addressed in https://github.com/mandiant/flare-floss/pull/1347/commits/6a34b48e0968c841fbaf14e132990d858a4574c5

scripts/analysis/bulk_analyze.py
mr-tz: Ah the doc is misleading then at least. Fine to mark as todo also for later changes. Although the doc update probably helps to avoid confusion moving forward.

vee1e: added a todo https://github.com/mandiant/flare-floss/pull/1347/commits/9d1f7dab4bb096d4d8e1db9de910e636da9fde03

floss/render/default.py
```diff
@@ -29,7 +29,9 @@
 import floss.language.identify
 from floss.render import Verbosity
 from floss.results import AddressType, StackString, TightString, DecodedString, ResultDocument, StringEncoding
+from floss.tags.filter import TagRules, hide_strings_by_rules
 from floss.render.sanitize import sanitize
+from floss.render.layout_text import render_strings
```
mr-tz, line 34: Should this function move to default? What about other parts that are default now?

floss/render/default.py
```diff
@@ -335,19 +345,27 @@ def render(results: floss.results.ResultDocument, verbose, disable_headers, colo
...
-        console.print("\n")
-
-    if results.analysis.enable_static_strings:
-        render_staticstrings(results.strings.static_strings, console, verbose, disable_headers)
-        console.print("\n")
+    # layout-aware path: no classic meta table (quantum-style)
+    # TODO(#1348): --load + --no-layout still hits this branch when the JSON
+    # already contains a layout tree (load does not clear results.layout).
+    if results.layout is not None and results.analysis.enable_static_strings:
+        layout_view = hide_strings_by_rules(results.layout, DEFAULT_TAG_RULES)
```
mr-tz, line 364: Move into the render function or what's the idea here?

floss/utils.py
mr-tz: Why not use extract_ascii_unicode_strings?

tests/test_qs.py
mr-tz: Not sure I follow. Can you show examples?

floss/render/default.py
vee1e: i feel it should remain in layout_text for now because it knows how to recursively render layout data. for this PR, the renderer is still hybrid, so default remains the coordinator and layout_text owns layout-specific rendering. When Rich rendering is used for every string type, we can consolidate the modules and remove the obsolete classic helper. we can add a TODO to clean this in the next PR where we deprecate the traditional view.

tests/test_qs.py
vee1e: real reproduction from the current code:

one.start_date:

```python
datetime.datetime(2026, 7, 31, 15, 46, 4, 737703)
```

two.start_date:

```python
datetime.datetime(
    2026, 7, 31, 15, 46, 4, 737703,
    tzinfo=TzInfo(0)
)
```

The emitted JSON was:

```json
// original
"runtime": {
  "decoded_strings": 0,
  "total": 0,
  "start_date": "2026-07-31T15:46:04.737703Z"
}
```

```json
// after JSON parse and re-render
"runtime": {
  "decoded_strings": 0.0, ← changed from 0 to 0.0
  "total": 0.0, ← changed from 0 to 0.0
  "start_date": "2026-07-31T15:46:04.737703+00:00Z" ← added "+00:00Z"
}
```

So the round-trip preserves the result content, but `one == two` is false and the JSON strings differ because the timestamp gains timezone information and integer runtime values become floats.

floss/utils.py
vee1e: yeah, actually taking a second look at it i don't feel we lose much here, the only extra branch is the empty check. changed in https://github.com/mandiant/flare-floss/pull/1347/commits/c4d8105

floss/render/default.py
vee1e: its [already inside](https://github.com/vee1e/flare-floss/blob/c4d81050901c26b9e4d78c780b7d153e5e3b1f4a/floss/render/default.py#L344-L355) `default.render()` though, unless you meant something else?


```python
def render(results: floss.results.ResultDocument, verbose, disable_headers, color):
    sys.__stdout__.reconfigure(encoding="utf-8")  # type: ignore [union-attr]
    console = Console(file=io.StringIO(), color_system=get_color(color), highlight=False, soft_wrap=True)


    # layout-aware path: no classic meta table (quantum-style)
    # TODO(#1348): --load + --no-layout still hits this branch when the JSON
    # already contains a layout tree (load does not clear results.layout).
    if results.layout is not None and results.analysis.enable_static_strings:
        layout_view = hide_strings_by_rules(results.layout, DEFAULT_TAG_RULES)
        render_strings(console, layout_view, DEFAULT_TAG_RULES)
        console.print()
    else:
```

floss/render/default.py
mr-tz: That works for me. Thanks.

floss/render/default.py
mr-tz: I meant the render_strings function called right after.

tests/test_qs.py
mr-tz: Ok and why does that happen? Floats for those fields doesn't make sense. The timezone is debatable. Could the entire comparison just overwrite the impacted field and then do a == comparison?

vee1e: > Floats for those fields doesn't make sense

the `get_runtime_diff()` function [intentionally records fractional seconds](https://github.com/vee1e/flare-floss/blob/c4d81050901c26b9e4d78c780b7d153e5e3b1f4a/floss/utils.py#L390-L391), and the pipeline assigns values accordingly to these fields. Changing the fields to integers would lose timing precision and affect persisted result files.

i've changed the values to all be floats in https://github.com/mandiant/flare-floss/pull/1347/commits/538a51058c9b2e6dc35f10fb2293e42d770dc0bd, but if you feel they would be better as ints we can revert the change.

also updated `start_date` to be UTC aware as well to finish making it a simpler one == two comparison.

floss/render/default.py
vee1e: i'd leave it in layout_text for now because it specifically renders layout trees, default only decides when to call it. plus, moving it now would separate it from the other layout-rendering helpers (eg. `render_string_string`, `render_string_tags`).

when all rendering is unified in the next PR, we can move the whole renderer _and_ helpers together.

.github/workflows/build.yml
mr-tz: need to add quantumstrand here

mr-tz: that should then result in an error for ./dist/${{ matrix.artifact_name }} --no stack tight decoded tests/data/test-decode-to-stack.exe -j | grep -q '#common'
--no is declared with nargs="+" and choices=[...].

same issue in scripts/analysis/bulk_analyze.py

floss/document.py
```diff
@@ -1,108 +0,0 @@
...
-            strings=result_strings,
-            children=result_children,
-        )
-
-
-class Sample(BaseModel):
-    md5: str
-    sha1: str
-    sha256: str
-    path: str
```
mr-tz, line 91: are we loosing this? can you double check please that no useful data goes missing?

floss/render/default.py
```diff
@@ -335,19 +345,27 @@ def render(results: floss.results.ResultDocument, verbose, disable_headers, colo
...
-        render_meta(results, console, verbose)
-        console.print("\n")
-
-    if results.analysis.enable_static_strings:
-        render_staticstrings(results.strings.static_strings, console, verbose, disable_headers)
-        console.print("\n")
+    # layout-aware path: no classic meta table (quantum-style)
+    # TODO(#1348): --load + --no-layout still hits this branch when the JSON
+    # already contains a layout tree (load does not clear results.layout).
+    if results.layout is not None and results.analysis.enable_static_strings:
```
mr-tz, line 363: via LLM: Since layout is on by default for any PE/ELF/Mach-O, render_meta (file path, identified language, runtime, version, imagebase, string counts, function-analysis stats in verbose mode) is now never shown by default, regardless of --verbose/quiet state — this used to be controlled by disable_headers/-q, now it's implicitly gated by layout success. This is called out in the PR description as intentional ("no classic meta when layout is present"), but it removes previously-always-visible diagnostic info (sample path, language ID, timing) from the common path with no opt-back-in short of --no-layout. Worth confirming this is the desired UX, not just a byproduct of the merge.

floss/pipeline.py
```diff
@@ -14,7 +14,538 @@
...
+        if enable_tags:
+            remove_false_positive_lib_strings(live)
+
+        return ResultLayout.from_layout(live)
+    except Exception as e:
+        logger.warning("layout-aware static analysis failed; using classic statics: %s", e)
+        return None
+
+
+def analyze(options: Options) -> Optional[ResultDocument]:
```
mr-tz, line 257: via LLM: Static-string extraction now always reads the whole file into memory
Previously floss.utils.get_static_strings used mmap for classic extraction (memory-efficient for large files). The new pipeline.analyze() does:
sample_buf = sample.read_bytes()
unconditionally, before any MAX_FILE_SIZE/--large-file gate (that gate only applies to the vivisect/deobfuscation path further down). get_static_strings (utils.py) is now dead for the main flow (still used by floss/language/go/coverage.py, but no longer from main.py/pipeline.py). For very large samples this is a real memory-footprint regression relative to the mmap-based path it replaces — worth at least gating full-buffer reads behind a size check, or keeping mmap and only materializing bytes for layout when needed.

tests/test_layout.py
mr-tz, line 1: this filename is outdated with the qs removal

vee1e: renamed in https://github.com/mandiant/flare-floss/pull/1347/commits/60e77920411743ab7f295a595c5e35cddb6f64a3, do we want to do the same for the other test_qs_* files?

floss/document.py
vee1e: yeah qs-viewer consumes these hashes, do we want to keep this for now or remove them and put them back in later? that viewer is already incompatible with unified `floss -j` (field names and top-level shape), not only hashes. Updating it is a separate port, not a reason to restore Sample for this PR.

floss/render/default.py
vee1e: yeah it matches PR intent, not a merge accident. we can keep as-is, or later add meta above the tree / under -v without --no-layout. No code bug to fix unless you want that opt-in.

.github/workflows/build.yml
vee1e: done https://github.com/mandiant/flare-floss/pull/1347/commits/b2a62e9b94c4f5015960c18b1a18b4e10b812bd0 and changed the arg order to be sample first

floss/pipeline.py
vee1e: we can have mmap for classic extraction and only read_bytes (or mmap + copy) when enable_layout and you actually run layout. plus add an optional size warn/gate before full materialization. not urgent for typical malware sizes unless we're doing multi‑GB samples on the static-only path. fixed in https://github.com/mandiant/flare-floss/pull/1347/commits/6890d7652802b5e259287fd7c15a4c6c94a92449

tests/test_layout.py
mr-tz: yes

vee1e: sure done in https://github.com/mandiant/flare-floss/pull/1347/commits/ae271e6b1ad94a015fbdb5d390e8436777c579b3

floss/pipeline.py
mr-tz: ah, we need the full buffer - since this will be the default / only option maybe worth keeping as was or just a note for future refinements in the next phase?!

floss/render/default.py
mr-tz: ok, lets leave this then

floss/document.py
mr-tz: this is for all consumer scripts and handy information to have, ie. what sample is this report for

floss/render/default.py
mr-tz: ok sounds good

floss/pipeline.py
vee1e: yeah reverted and added a TODO for later in https://github.com/mandiant/flare-floss/pull/1347/commits/bedeab4364a6fef2985afc9143ad1f34cce79d26

tests/test_qs.py
mr-tz: thats good now

floss/main.py
mr-tz: ok thanks

floss/document.py
vee1e: fair enough, added the hashes back in https://github.com/mandiant/flare-floss/pull/1347/commits/d5f00fa40dff65544668dcb78d1e89781c1979ac, now they're filled during analysis from the sample buffer. path was already there as `metadata.file_path` so fine to discard.

mr-tz (approved):

mr-tz:
Good progress @vee1e towards full integration. Thanks!

PR #1353, opened 2026-08-10
feat(zig): add zig strings extraction

CX330Blake:
## Summary

- add Zig binary identification using the `ZIG_PROGRESS` runtime string
- add `Language.ZIG` to the supported language identifiers
- add a Zig PE identification test
- document the Zig identification heuristic

## Test data

https://github.com/mandiant/flare-floss-testfiles/pull/28

The test-data PR adds a minimal Zig source file and a Windows x86-64 PE compiled with Zig 0.16.0.

### DIFF
--- a/floss/language/identify.py
+++ b/floss/language/identify.py
@@ -35,6 +35,7 @@ class Language(Enum):
     GO = "go"
     RUST = "rust"
     DOTNET = "dotnet"
+    ZIG = "zig"
     UNKNOWN = "unknown"
     DISABLED = "none"
 
@@ -44,6 +45,11 @@ def identify_language_and_version(sample: Path, static_strings: Iterable[StaticS
     if is_rust:
         logger.info("Rust binary found with version: %s", version)
         return Language.RUST, version
+
+    is_zig, version = get_if_zig_and_version(static_strings)
+    if is_zig:
+        logger.info("Zig binary found")
+        return Language.ZIG, version
 
     # open file as PE for further checks
     try:
@@ -94,6 +100,21 @@ def get_if_rust_and_version(static_strings: Iterable[StaticString]) -> Tuple[boo
             else:
                 logger.debug("hash %s not found in Rust commit hash database", matches["hash"])
                 return True, VERSION_UNKNOWN_OR_NA
+
+    return False, VERSION_UNKNOWN_OR_NA
+
+
+def get_if_zig_and_version(static_strings: Iterable[StaticString]) -> Tuple[bool, str]:
+    """
+    Return whether the binary appears to be compiled with Zig.
+
+    ZIG_PROGRESS is used as a Zig compiler/runtime heuristic, but it does not
+    contain enough information to determine the Zig compiler version.
+    """
+
+    for static_string_obj in static_strings:
+        if "ZIG_PROGRESS" in static_string_obj.string:
+            return True, VERSION_UNKNOWN_OR_NA
 
     return False, VERSION_UNKNOWN_OR_NA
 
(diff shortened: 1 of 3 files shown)


floss/language/identify.py
```diff
@@ -35,6 +35,7 @@ class Language(Enum):
     GO = "go"
     RUST = "rust"
     DOTNET = "dotnet"
+    ZIG = "zig"
```
vee1e, line 39: This branch predates `Language.AUTO` (added on master after your base `98b1fa5`, about 6 weeks old now). Correction to an earlier concern of mine: a 3-way merge keeps both sides, so AUTO survives the merge. Still, please rebase, so this branch builds and tests against the current enum rather than a stale copy without it.

floss/language/identify.py
```diff
@@ -98,6 +104,110 @@ def get_if_rust_and_version(static_strings: Iterable[StaticString]) -> Tuple[boo
...
+    if (
+        ".tls" in sections
+        and tls_directory
+        and tls_directory.VirtualAddress
+        and tls_directory.Size
+    ):
+        score += 4
+        has_structure = True
+
+    if "RtlExitUserProcess" in imports.get("ntdll.dll", set()):
```
vee1e: DLL names are lowercased at line 122, but import names compare exact-case here and in the lock trio below. An oddly-cased import name slips past. Please lowercase both sides of the comparison.

floss/language/identify.py
```diff
@@ -98,6 +104,110 @@ def get_if_rust_and_version(static_strings: Iterable[StaticString]) -> Tuple[boo
...
+        ".rdata",
+        ".buildid",
+        ".data",
+        ".pdata",
+        ".tls",
+        ".reloc",
+    }
+
+    # should be exactly the same as the one of these two
+    if section_names in (legacy_sections, modern_sections):
```
vee1e: Exact set equality here means one extra section voids these 2 points. A real-world Zig binary with `.rsrc` (icons, version info) matches neither bundle and loses the score. Consider subset matching (the binary sections include the bundle) instead of requiring an exact match.

tests/test_language_id.py
```diff
@@ -17,6 +17,8 @@
             VERSION_UNKNOWN_OR_NA,
         ),
         ("data/language/rust/rust-hello/bin/rust-hello.exe", Language.RUST, "1.69.0"),
+        ("data/language/zig/zig-hello/bin/zig-hello.exe", Language.ZIG, VERSION_UNKNOWN_OR_NA),
```
vee1e: Both Zig positives depend on the tests/data pin `702d965`, which fetches from neither the upstream nor the fork testfiles remote, and test-data PR mandiant/flare-floss-testfiles#29 is still open. So these two cases cannot run anywhere yet. Please land the test-data PR first so the true-positive path actually executes before this merges.

floss/language/identify.py
```diff
@@ -98,6 +104,110 @@ def get_if_rust_and_version(static_strings: Iterable[StaticString]) -> Tuple[boo
...
+        score += 2
+        has_structure = True
+
+    mapped_data = b"".join(
```
vee1e: Gluing every section into one blob means a marker split across two neighboring sections can match across the join, crediting bytes that sit nowhere contiguous in the binary. Please test each section bytes separately and OR the hits instead.

CX330Blake: After looking more into it, the runtime markers will only appears in `.rdata` section. I'll update the code for this.

vee1e:
@CX330Blake Is this PR still a work in progress? I noticed that Zig detection and rendering are added, but the main pipeline only invokes language-specific extraction for Go and Rust, so `extract_zig_strings()` is not currently called. Is the pipeline integration intentionally deferred to a follow-up, or should it be included here?

CX330Blake:
Thanks for catching this. This wasn't intentional — I made a mistake while cleaning up the git history, and the pipeline integration for `extract_zig_strings()` was accidentally dropped.

It should be included in this PR. I've already restored the integration and make sure the full Zig extraction path is wired into the main pipeline.

We can also squash the commits once the PR is ready to merge.

PR #1355, opened 2026-08-11
Folding of QS CLI - Part 1

vee1e:
## Finish folding of QS CLI 

Reference Doc is at https://docs.google.com/document/d/19i0fkJZBPHb15-NHeALzk3RRMYLafnCLNScAOdCiFfM/edit

## Summary

This change implements the first batch of the FLOSS migration development spec.
It completes the four small tasks in the plan. The medium and large tasks stay in later batches.
All related tests pass. The linters and the type checker pass.

## Task list

- [x] Point the tag scripts at the current tag database location.
- [x] Add the layout and tags runtime timing fields to the results document.
- [x] Record the elapsed time of the layout and tag steps in the pipeline.
- [x] Define the help, version, and no-argument behavior.
- [x] Remove the `--no-layout` and `--no-tags` flags.
- [x] Remove the `--load` flag and detect JSON results documents automatically.
- [x] Remove the `--no` and `--only` flags.
- [x] Add the `--string-type` and `--no-string-type` flags with mutual exclusion.
- [x] Rename `--functions` to `--analyze-functions`.

## To be done after this PR

- [ ] Move the string render function into the default render module.
- [ ] Remove the classic view and the obsolete classic helpers.
- [ ] Add centered headings for the stack, tight, and decoded string sections.
- [ ] Add the `--columns` option for column visibility.
- [ ] Add the `--section` and `--no-section` flags.
- [ ] Add the `--structure` and `--no-structure` flags.
- [ ] Add the `--tag` and `--no-tag` flags.
- [ ] Add the `--interesting` shortcut to exclude the noisy tags.
- [ ] Add the `--query` flag to filter strings by content.
- [ ] Add the `--max-strings` flag to cap the strings per section.
- [ ] Emit the full results document as JSON with sorted keys.
- [ ] Add the `--summary` output.
- [ ] Emit structured JSON errors in JSON mode.
- [ ] Add the analysis cache module.
- [ ] Compute the cache key from the sample content hash and the tool version.
- [ ] Store the results document atomically in the cache.
- [ ] Load a valid cached document and skip analysis.
- [ ] Support the environment variables for the cache directory and the cache switch.
- [ ] Drop an invalid cache entry and re-analyze.


floss/main.py
```diff
@@ -44,6 +45,41 @@ def is_running_standalone() -> bool:
...
+    """
+    try:
+        with sample.open("rb") as f:
+            for byte in iter(lambda: f.read(1), b""):
+                if not byte.strip():
+                    continue
+                return byte == b"{"
+    except OSError:
+        return False
+    return False
```
vee1e: this could be done via json.loads? or would it impact perf too much

README.md
```diff
@@ -89,16 +87,15 @@ Extract obfuscated strings from a malware binary:
 
 Only extract stack and tight strings:
 
-    $ floss --only stack tight -- suspicious.exe
+    $ floss --string-type stack tight -- suspicious.exe
```
vee1e, line 90: before the spec `--only` already used `nargs="+"` so values were space-separated. should we still keep comma-separated syntax for `--columns` considering this fact for consistency?

floss/main.py
mr-tz: lets check it's a proper floss doc based on some field, if we need to load the entire file for that do so, if not then even better

README.md
mr-tz: lets use spaces then

.github/workflows/build.yml
```diff
@@ -125,7 +125,7 @@ jobs:
       - name: Set zip name
         run: echo "zip_name=floss-${GITHUB_REF#refs/tags/}-${{ matrix.asset_name }}.zip" >> $GITHUB_ENV
       - name: Zip ${{ matrix.artifact_name }} into ${{ env.zip_name }}
-        run: zip ${{ env.zip_name }} ${{ matrix.artifact_name }}
+        run: zip "$zip_name" "${{ matrix.artifact_name }}"
```
mr-tz, line 128: undo?

doc/usage.md
```diff
@@ -18,9 +18,7 @@ Since FLOSS also extracts static strings (like `strings.exe`),
 Here's a summary of the command line flags and options you
  can provide to FLOSS to modify its behavior.
 
-See `floss -h` for all supported arguments and usage examples. This displays the most used arguments only.
-
-To see all supported arguments run `floss -H`.
+See `floss -h` for all supported arguments and usage examples.
```
mr-tz, line 21: are we still doing a limited help when run without args or is there only one help screen for all cases?

doc/usage.md
```diff
@@ -137,7 +136,7 @@ This can improve performance as FLOSS by perhaps one-third
   to always manually identify decoding routines).
 Specify functions by using their hex-encoded virtual address.
 
-    floss.exe --functions 0x401000 0x402000 malware.exe
+    floss.exe --analyze-functions 0x401000 0x402000 malware.exe
```
mr-tz: does this even work? without the -- between function offsets and the sample?

floss/cli.py
```diff
@@ -89,46 +93,34 @@ def make_parser(argv):
...
           do not extract static strings
-            floss --no static -- suspicious.exe
+            floss --no-string-type static -- suspicious.exe
 
           only extract stack and tight strings
-            floss --only stack tight -- suspicious.exe
+            floss --string-type stack tight -- suspicious.exe
 
-          classic static strings without layout/tags
-            floss --no-layout -- suspicious.exe
```
mr-tz, line 110: show `--plain` usage here

doc/usage.md
mr-tz: thinking a very basic help with examples on run without args may be worth considering

mr-tz: but maybe also not worth the extra effort

floss/cli.py
```diff
@@ -146,33 +138,22 @@ def make_parser(argv):
...
-        help="disable layout-aware static analysis (classic flat static listing)",
-    )
-    analysis_group.add_argument(
-        "--no-tags",
-        action="store_true",
-        help="disable tag databases (layout sections without prevalence/library tags)",
+        help="do not extract specified string type(s); valid values: %s" % ", ".join(STRING_TYPE_CHOICES),
     )
 
     advanced_group = parser.add_argument_group("advanced arguments")
```
mr-tz, line 162: do we still want the advanced part? can you share a screenshot of the help with these changes as of now?

floss/cli.py
```diff
@@ -188,86 +169,57 @@ def make_parser(argv):
...
-            else argparse.SUPPRESS
-        ),
+        help="only analyze the specified functions, hex-encoded like 0x401000, space-separate multiple functions",
     )
     advanced_group.add_argument(
         "--disable-progress",
         action="store_true",
-        help="disable all progress bars" if show_all_options else argparse.SUPPRESS,
+        help="disable all progress bars",
     )
```
mr-tz: keep or do this via --quiet?

floss/main.py
```diff
@@ -69,11 +87,15 @@ def main(argv=None) -> int:
...
 
-    parser = make_parser(argv)
+    parser = make_parser()
     try:
+        if not argv:
+            # no arguments: print the full option list and exit with code 1
+            parser.print_help()
+            return 1
         args = parser.parse_args(args=argv)
         if args.enabled_types and args.disabled_types:
```
mr-tz: tbd rename the variable names?

floss/main.py
```diff
@@ -108,17 +130,17 @@ def main(argv=None) -> int:
         if StringType.STATIC.value not in disabled:
             disabled.append(StringType.STATIC.value)
 
-    # layout/tags: temporary dedicated flags (not --no string types); see issue #1348
+    # layout/tags are always on: automatic and detected from the sample content
```
mr-tz, line 129: what does "detected from the sample content" mean?

README.md
vee1e: updated the spec to this, in future iterations columns will implemented as space separated

floss/pipeline.py
```diff
@@ -219,10 +219,11 @@ def _try_layout_static(
     buf: bytes,
     min_length: int,
     enable_tags: bool,
-) -> Optional[ResultLayout]:
+) -> Tuple[Optional[ResultLayout], float]:
     """
     Quantum-style layout extract when PE/ELF/Mach-O parse succeeds.
-    Returns serializable ResultLayout, or None to fall back to classic statics.
+    Returns serializable ResultLayout (or None to fall back to classic statics)
+    along with the elapsed seconds spent in the tag database matching step.
```
mr-tz: I think it's better to do the timing outside of the function (ideally with some wrapper function) - or why do you suggest to do it in the function and return the time?

floss/results.py
```diff
@@ -283,7 +285,7 @@ class Analysis:
 @dataclass
 class Metadata:
     file_path: str
-    # sample identity for consumers (path + content hashes); hashes empty when unknown (e.g. --load of old JSON)
+    # sample identity for consumers (path + content hashes); hashes empty when unknown (e.g. loaded old JSON)
```
mr-tz: remove this comment? stale or outdated?

scripts/tags/query_string.py
```diff
@@ -12,21 +12,20 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 def load_db_gp():
-    gpfile = os.path.join(os.path.dirname(__file__), "..", "..", "floss", "qs", "db", "data", "gp", "gp.jsonl.gz")
-    compress = gpfile.endswith(".gz")
-    return StringGlobalPrevalenceDatabase.from_file(pathlib.Path(gpfile), compress=compress)
+    gpfile = data_root() / "gp" / "gp.jsonl.gz"
+    compress = gpfile.suffix == ".gz"
```
mr-tz, line 27: is suffix for .json.gz resulting in .gz?

tests/test_main.py
```diff
@@ -21,13 +21,27 @@
 
 
 def test_main_help():
-    for help_str in ("-h", "-H"):
+    for help_str in ("-h", "--help"):
```
mr-tz, line 25: test running without args here also

tests/test_main.py
```diff
@@ -21,13 +21,27 @@
...
+        floss.main.main(["--version"])
+    assert pytest_wrapped_e.type == SystemExit
+    assert pytest_wrapped_e.value.code == 0
+
+
+def test_main_no_args_prints_help_and_exits_1(capsys):
+    assert floss.main.main([]) == 1
+    out = capsys.readouterr().out
+    assert "usage:" in out
+    assert "--analyze-functions" in out
```
mr-tz: not a good test, combine with above also instead

.github/workflows/build.yml
vee1e: previously it was resulting in a zizmor [CI failure](https://github.com/mandiant/flare-floss/actions/runs/31494294204/job/93788539890)

floss/main.py
vee1e: updated https://github.com/mandiant/flare-floss/pull/1355/commits/7de508c5e24fa45b14e50c5c0bfdff0129b0c57b

- peeks the first non-whitespace byte as a fast path so large binaries are never read for the check.
- only when the content looks like JSON, does a full `json.loads` and confirms the top-level ResultDocument fields (metadata, analysis, strings) are present.

doc/usage.md
vee1e: i'm still not sure on this, with the number of args we have it would be nice for a user with a simpler use case to get instant info and not trudge through the deetailed cli spec. wdyt?

doc/usage.md
vee1e: nope thanks for catching, we do need `--`. fixed https://github.com/mandiant/flare-floss/pull/1355/commits/790f787b77c5cafbb48d8b53b1128fb15bbed5ba

.github/workflows/build.yml
mr-tz: ah yeah, we're opted into that now - good fix then :)

doc/usage.md
mr-tz: yeah I think a short TLDR help would be great - ideally something we reuse for the full help also if possible

doc/usage.md
mr-tz: dont add that to the help though its common usage

floss/cli.py
```diff
@@ -188,86 +169,58 @@ def make_parser(argv):
...
         type=lambda x: int(x, 0x10),
         default=None,
         nargs="+",
-        help=(
-            "only analyze the specified functions, hex-encoded like 0x401000, space-separate multiple functions"
-            if show_all_options
-            else argparse.SUPPRESS
-        ),
+        help="only analyze the specified functions, hex-encoded like 0x401000, space-separate multiple functions, "
+        "end with -- before the sample",
```
```suggestion
```

floss/main.py
```diff
@@ -44,6 +45,37 @@ def is_running_standalone() -> bool:
...
+    """
+    try:
+        with sample.open("rb") as f:
+            for byte in iter(lambda: f.read(1), b""):
+                if not byte.strip():
+                    continue
+                if byte != b"{":
+                    return False
+                break
+    except OSError:
```
mr-tz: what would be the error here?

floss/main.py
```diff
@@ -44,6 +45,37 @@ def is_running_standalone() -> bool:
...
+        return False
+
+    try:
+        data = json.loads(sample.read_bytes())
+    except (OSError, ValueError):
```
mr-tz: also json decoding errors?

floss/main.py
```diff
@@ -44,6 +45,37 @@ def is_running_standalone() -> bool:
...
+        return False
+
+    return isinstance(data, dict) and {"metadata", "analysis", "strings"} <= set(data)
```
mr-tz: interesting syntax with the `<=`

README.md
mr-tz: thanks

floss/cli.py
vee1e: https://github.com/mandiant/flare-floss/pull/1355/commits/beef9957b5d6639969a675e4b1a514e692e5466d

floss/cli.py
vee1e: yes, we still want the advanced part. per spec §3.4.2, `--help` shows the full option list including advanced options. Removing `-H` would mean there's no core-only toggle anymore, the "advanced arguments" group header is now just visual organization, and every option is always visible.

output:

```
usage: __main__.py [-h] [-n MIN_LENGTH]
                   [--string-type {static,stack,tight,decoded} [{static,stack,tight,decoded} ...]]
                   [--no-string-type {static,stack,tight,decoded} [{static,stack,tight,decoded} ...]]
                   [-f {auto,pe,sc32,sc64}] [--language {go,rust,dotnet,none}]
                   [--analyze-functions FUNCTIONS [FUNCTIONS ...]]
                   [--disable-progress] [--signatures SIGNATURES] [-L]
                   [--version] [-j] [-v] [-d] [-q]
                   [--color {auto,always,never}]
                   sample

The FLARE team's open-source tool to extract ALL strings from malware.
  __main__.py 3.1.1 - https://github.com/mandiant/flare-floss/

FLOSS extracts the following string types:
 1. static strings:  "regular" ASCII and UTF-16LE strings
 2. stack strings:   strings constructed on the stack at run-time
 3. tight strings:   special form of stack strings, decoded on the stack
 4. decoded strings: strings decoded in a function

Language-specific strings:
 1. Go:   strings from binaries written in Go
 2. Rust: strings from binaries written in Rust

By default, static strings are layout-aware with tags for PE/ELF/Mach-O
(section context, prevalence/library/expert tags).

positional arguments:
  sample                path to sample to analyze

options:
  -h, --help            show this help message and exit
  -n MIN_LENGTH, --minimum-length MIN_LENGTH
                        minimum string length

analysis arguments:
  --string-type {static,stack,tight,decoded} [{static,stack,tight,decoded} ...]
                        only extract specified string type(s); valid values:
                        static, stack, tight, decoded
  --no-string-type {static,stack,tight,decoded} [{static,stack,tight,decoded} ...]
                        do not extract specified string type(s); valid values:
                        static, stack, tight, decoded

advanced arguments:
  -f {auto,pe,sc32,sc64}, --format {auto,pe,sc32,sc64}
                        select sample format, auto: (default) detect file type
                        automatically, pe: Windows PE file, sc32: 32-bit
                        shellcode, sc64: 64-bit shellcode
  --language {go,rust,dotnet,none}
                        use language-specific string extraction, auto-detect
                        language by default, disable using 'none'
  --analyze-functions FUNCTIONS [FUNCTIONS ...]
                        only analyze the specified functions, hex-encoded like
                        0x401000, space-separate multiple functions, end with
                        -- before the sample
  --disable-progress    disable all progress bars
  --signatures SIGNATURES
                        path to .sig/.pat file or directory used to identify
                        library functions, use embedded signatures by default
  -L, --large-file      allow processing files larger than 16 MB
  --version             show program's version number and exit

rendering arguments:
  -j, --json            emit JSON instead of text
  -v, --verbose         enable verbose results, e.g. including function
                        offsets (does not affect JSON output)

logging arguments:
  -d, --debug           enable debugging output on STDERR, specify multiple
                        times to increase verbosity
  -q, --quiet           disable all status output on STDOUT except fatal
                        errors
  --color {auto,always,never}
                        enable ANSI color codes in results, default: only
                        during interactive session

examples:
  extract all strings from an executable
    floss suspicious.exe

  classic flat list of strings without layout and tags
    floss --plain suspicious.exe

  do not extract static strings
    floss --no-string-type static -- suspicious.exe

  only extract stack and tight strings
    floss --string-type stack tight -- suspicious.exe

  extract strings from 32-bit shellcode
    floss -f sc32 shellcode.bin

  only decode strings from the specified functions
    floss --analyze-functions 0x401000 0x401100 -- suspicious.exe

  extract strings from a binary written in Go (if automatic language identification fails)
    floss --language go program.exe
```

floss/cli.py
vee1e: removed https://github.com/mandiant/flare-floss/pull/1355/commits/7cb25013fd5da25d1d50a09d8efc4ecde8797f9b 

we already discussed removing it in spec §3.4.8

floss/main.py
vee1e: if file detected as PE/ELF/MachO -> show rich layout and tags
else only show flat layout

floss/main.py
vee1e: done here only in https://github.com/mandiant/flare-floss/pull/1355/commits/5d9ebdfb075aff1e663eee4df81647e55af88b65

floss/results.py
vee1e: removed in https://github.com/mandiant/flare-floss/pull/1355/commits/44b888de67962c258c7aea0caa8bc30816c1fa01

floss/pipeline.py
vee1e: agreed, addressed in https://github.com/mandiant/flare-floss/pull/1355/commits/873c8bdd62c3fa54c67b34523cca7f26dc6b1fb7

- `_compute_layout` now computes the structured layout and extracts strings, no timing.
- `_tag_layout` tags strings, marks structures, and drops false positives.
- `_try_layout_static` is now the timing wrapper, it times the whole layout+tag step and the tag DB matching step (runtime.tags).

scripts/tags/query_string.py
vee1e: yeah, `.suffix` returns only the final suffix, so `Path("gp.jsonl.gz").suffix` is .gz. If you wanted all suffixes, you'd use .suffixes, giving `.jsonl` and `.gz`.

tests/test_main.py
vee1e: merged them both in https://github.com/mandiant/flare-floss/pull/1355/commits/e178418470d625b3a45ff4c1ca4dff449dfca5d6

floss/main.py
vee1e: any failure to open/read the file on re-open. `OSError` covers stuff like:

- `FileNotFoundError` (eg. file was removed between the initial argparse open and this re-open)
- `PermissionError` (eg. not readable)
- `IsADirectoryError` (eg. path is a directory)

treating those as "not a results document" avoids crashing in the race between the first open and this one.

floss/main.py
vee1e: covered already. `JSONDecodeError` and `UnicodeDecodeError` are both subclasses of ValueError.

floss/main.py
vee1e: lol. yeah `<=` can be used to find subsets. made a more one in https://github.com/mandiant/flare-floss/pull/1355/commits/48095d9532c3e82fd9725ae856d497ccb8aa7302

floss/pipeline.py
```diff
@@ -241,17 +240,61 @@ def _try_layout_static(
             return None
 
         extract_layout_strings(live, min_length)
-        # tag_strings always converts ExtractedString → TaggedString (needed for mark_structures)
-        taggers = load_databases() if enable_tags else []
-        live.tag_strings(taggers)
-        live.mark_structures()
-        if enable_tags:
-            remove_false_positive_lib_strings(live)
+        return live
```
mr-tz: what is live? Ideally we dont have the implicit modification here

floss/utils.py
```diff
@@ -471,11 +471,11 @@ def get_call_funcname(api):
...
 
 
-def is_string_type_enabled(type_, disabled_types, enabled_types):
-    if disabled_types:
-        return type_ not in disabled_types
-    elif enabled_types:
-        return type_ in enabled_types
+def is_string_type_enabled(type_, no_string_types, string_types):
+    if no_string_types:
+        return type_ not in no_string_types
```
mr-tz: here the enabled/disabled syntax is definitely easier to follow, what do you think for maintainability moving forward?

mr-tz: use `enabled_string_types` maybe?

scripts/tags/query_string.py
mr-tz: thanks

tests/test_load.py
```diff
@@ -94,9 +97,33 @@ def test_load(tmp_path):
...
+def test_is_results_document_true_for_results_json(tmp_path):
+    p = tmp_path / "results.json"
+    p.write_text(RESULTS)
+    assert floss.main.is_results_document(p)
+
+
+def test_is_results_document_false_for_binary(tmp_path, exefile):
+    from pathlib import Path
+
+    assert not floss.main.is_results_document(tmp_path / "results.json")
```
mr-tz: why is this binary? im confused compared to the prior test

mr-tz: its empty here?

tests/test_main.py
```diff
@@ -20,13 +20,29 @@
...
+def test_main_help(capsys):
+    for help_str in ("-h", "--help"):
         # via https://medium.com/python-pandemonium/testing-sys-exit-with-pytest-10c6e5f7726f
         with pytest.raises(SystemExit) as pytest_wrapped_e:
             floss.main.main([help_str])
         assert pytest_wrapped_e.type == SystemExit
         assert pytest_wrapped_e.value.code == 0
+        out = capsys.readouterr().out
+        assert "usage:" in out
+        assert "--analyze-functions" in out
```
mr-tz: is this a good test string here? same for usage although that's more clear

floss/pipeline.py
```diff
@@ -241,17 +240,61 @@ def _try_layout_static(
...
+    except Exception as e:
+        logger.warning("layout-aware static analysis failed; using classic statics: %s", e)
+        return None
+
+
+def _tag_layout(
```
mr-tz: is naming inconsistent here with using the `_` prefixes only here?

floss/pipeline.py
```diff
@@ -241,17 +240,61 @@ def _try_layout_static(
...
+
+
+def _try_layout_static(
+    buf: bytes,
+    min_length: int,
+    enable_tags: bool,
+    runtime: Runtime,
+) -> Optional[ResultLayout]:
+    """
+    Layout-aware static extraction with timing.
```
mr-tz, line 259: with wrapping i meant something potentially along the lines of https://github.com/mandiant/capa/blob/b58647299a349e59a9aa1f33ad5058865d286db3/capa/main.py#L143

vee1e: refactored in https://github.com/mandiant/flare-floss/pull/1355/commits/e0eefaf to match capa's timing context-manager pattern. added `_measure_runtime` and rewrote `_try_layout_static` to wrap the layout and tag steps in nested `with` blocks.

floss/pipeline.py
vee1e: yup right, `get_file_type`, `load_vw`, `get_signatures` are internal-only as well but not prefixed. fixed https://github.com/mandiant/flare-floss/pull/1355/commits/5f7a5d2

tests/test_main.py
vee1e: its fine imo but if we want stable options we can change it. i made it assert `--json` in https://github.com/mandiant/flare-floss/pull/1355/commits/c5c2224

tests/test_load.py
vee1e: yeah that was a bad test. removed the `results.json` assert and hoisted path to the top of file in  https://github.com/mandiant/flare-floss/pull/1355/commits/235963c

floss/utils.py
vee1e: yeah agreed, changed it to enabled/disabled terminology in  https://github.com/mandiant/flare-floss/pull/1355/commits/60b6254

floss/pipeline.py
vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1355/commits/7dbe90d now we have a better name than `live` and make the layout extraction explicit.

floss/main.py
```diff
@@ -69,11 +101,15 @@ def main(argv=None) -> int:
     if argv is None:
         argv = sys.argv[1:]
 
-    parser = make_parser(argv)
+    parser = make_parser()
     try:
+        if not argv:
+            # no arguments: print the full option list and exit with code 1
+            parser.print_help()
```
mr-tz, line 76: from above: "yeah I think a short TLDR help would be great - ideally something we reuse for the full help also if possible" - https://github.com/mandiant/flare-floss/pull/1355/changes#r3758834222

WDYT? we can also track for further consideration, dont have to do it here

vee1e: yeah lets track this for later once we finalize and add all the cli args

floss/main.py
```diff
@@ -99,26 +135,26 @@ def main(argv=None) -> int:
...
-        enable_tight_strings=is_string_type_enabled(StringType.TIGHT, disabled, enabled),
-        enable_decoded_strings=is_string_type_enabled(StringType.DECODED, disabled, enabled),
-        enable_layout=not args.no_layout,
-        enable_tags=not args.no_tags,
+        enable_static_strings=is_string_type_enabled(StringType.STATIC, disabled_string_types, enabled_string_types),
+        enable_stack_strings=is_string_type_enabled(StringType.STACK, disabled_string_types, enabled_string_types),
+        enable_tight_strings=is_string_type_enabled(StringType.TIGHT, disabled_string_types, enabled_string_types),
+        enable_decoded_strings=is_string_type_enabled(StringType.DECODED, disabled_string_types, enabled_string_types),
+        enable_layout=True,
+        enable_tags=True,
```
mr-tz: do we still need these flags here?

floss/pipeline.py
```diff
@@ -135,7 +137,7 @@ def select_functions(vw, asked_functions: Optional[List[int]]) -> Set[int]:
     return asked_functions_
 
 
-def get_file_type(sample_file_path: Path) -> bytes:
+def _get_file_type(sample_file_path: Path) -> bytes:
```
mr-tz: hm should this be combined with is_results_document or at least exist in the same file?

floss/pipeline.py
```diff
@@ -215,40 +217,91 @@ def get_signatures(sigs_path: Path) -> List[Path]:
...
     """
-    from floss.tags import load_databases, remove_false_positive_lib_strings
     from floss.layout import compute_layout
     from floss.ranges import Slice
     from floss.layout.extract import extract_layout_strings
 
     try:
         file_slice = Slice.from_bytes(buf=buf)
-        live = compute_layout(file_slice)
+        parsed = compute_layout(file_slice)
```
mr-tz: use `parsed_layout` or just `layout` as var name?

and adjust the other layout if needed

floss/layout/extract.py
```diff
@@ -68,14 +68,14 @@ def extract_strings(slice: Slice, n: int = MIN_STR_LEN) -> Iterable[ExtractedStr
     )
 
 
-def extract_layout_strings(layout: Layout, min_len: int):
+def extract_layout_strings(layout: Layout, min_len: int) -> Layout:
```
mr-tz: well, we don't win much here do we?

then may just keep as was or can you think of another improvement?

floss/pipeline.py
mr-tz: i was thinking to not add the `_` to the names as we don't do that at all(?) across the codebase

just want to keep some consistency one way or another - is pipeline supposed to be a API kinda layer for others to use?

floss/pipeline.py
```diff
@@ -215,40 +217,91 @@ def get_signatures(sigs_path: Path) -> List[Path]:
...
+    # tag_strings always converts ExtractedString → TaggedString (needed for mark_structures)
+    taggers = load_databases() if enable_tags else []
+    layout.tag_strings(taggers)
+    layout.mark_structures()
+    if enable_tags:
+        remove_false_positive_lib_strings(layout)
+
+
+@contextlib.contextmanager
+def _measure_runtime(runtime: Runtime, field: str) -> Iterator[None]:
```
mr-tz: would this make sense as a method of the Runtime class instead / alternatively?

we may want to measure other things though (are we right now?) which would be a counter argument

floss/pipeline.py
```diff
@@ -376,7 +429,9 @@ def analyze(options: Options) -> Optional[ResultDocument]:
         # only layout/tag work for static_strings runtime — not language ID or the TTY prompt above
         layout_t0 = time()
```
mr-tz: use timing wrapper for these also or track to do later on

tests/test_main.py
```diff
@@ -20,13 +20,29 @@
...
+    # running without arguments prints the same help and exits with code 1
+    assert floss.main.main([]) == 1
+    out = capsys.readouterr().out
+    assert "usage:" in out
+    assert "--json" in out
+
+
+def test_main_version():
+    with pytest.raises(SystemExit) as pytest_wrapped_e:
+        floss.main.main(["--version"])
```
mr-tz, line 44: also test that the correct version is part of the output

floss/pipeline.py
vee1e: ah okay, i think we don't use it then. i don't see pipeline being an API either. removed the prefix from all of them in https://github.com/mandiant/flare-floss/pull/1355/commits/585cceb14ecf855427f751cf00cc6d6b29265a0f

floss/pipeline.py
vee1e: yeah, moved the context manager onto the `Runtime` class itslef in https://github.com/mandiant/flare-floss/pull/1355/commits/35432977f995668bd8a51228b023a407a198d1c0. in floss/results.py.

re: the counter-argument, right now everything we time goes into a Runtime field, so Runtime.measure covers today's usage; if we ever time non-runtime things, we can add a standalone helper back. i kept the manual get_runtime_diff() patterns as-is since they measure spans across multiple steps.

floss/layout/extract.py
vee1e: we can make string extraction a method on the `Layout` class instead of a standalone function. right now the steps are split, `tag_strings` and `mark_structures` are methods on Layout, but string extraction is a free function extract_layout_strings(layout, min_len). should we move it into Layout as well?

for now i removed the spurious return in https://github.com/mandiant/flare-floss/pull/1355/changes/9c4da76

floss/pipeline.py
vee1e: just `layout` was actually the first draft but i thought it would be too confusing for two vars to have the same name. also since `extract_layout_strings` mutates and returns the same tree, the redundant `layout` variable is gone and the function now returns the extraction result directly, https://github.com/mandiant/flare-floss/pull/1355/commits/68ec89f0834464cb8b9a65ed2ed91a9a6ed665a1:

```py
parsed_layout = layout_compute(file_slice)
...
return extract_layout_strings(parsed_layout, min_length)
```

floss/pipeline.py
vee1e: yeah combined in  https://github.com/mandiant/flare-floss/pull/1355/commits/3244737 and https://github.com/mandiant/flare-floss/pull/1355/commits/1164d7d56479b8453a5e1c1ec1474bc86b8a6a52

floss/main.py
vee1e: nope, they're redundant since `Analysis` already defaults them to True, fix https://github.com/mandiant/flare-floss/pull/1355/commits/3244737cda5b0e8845d124a3ed96c1e8899727cd

floss/main.py
mr-tz: #1359

floss/main.py
```diff
@@ -99,26 +103,24 @@ def main(argv=None) -> int:
...
 
-    # layout/tags: temporary dedicated flags (not --no string types); see issue #1348
+    # layout/tags are always on: automatic and detected from the sample content
     analysis = Analysis(
-        enable_static_strings=is_string_type_enabled(StringType.STATIC, disabled, enabled),
-        enable_stack_strings=is_string_type_enabled(StringType.STACK, disabled, enabled),
-        enable_tight_strings=is_string_type_enabled(StringType.TIGHT, disabled, enabled),
-        enable_decoded_strings=is_string_type_enabled(StringType.DECODED, disabled, enabled),
-        enable_layout=not args.no_layout,
-        enable_tags=not args.no_tags,
```
mr-tz, line 118: and do we need the fields on Analysis?

floss/pipeline.py
```diff
@@ -215,40 +203,79 @@ def get_signatures(sigs_path: Path) -> List[Path]:
...
+
+    Runs the layout computation and the tag matching steps, and records
+    the elapsed time of the layout and tag step in ``runtime.layout`` and
+    the tag database matching step in ``runtime.tags``.
+
+    Returns a serializable ResultLayout, or None to fall back to classic
+    statics. Default-on layout must not crash the whole run.
+    """
+    try:
+        with runtime.measure("layout"):
```
mr-tz: maybe `measure_and_set_time` or similar to better indicate the implied mod here?

floss/layout/extract.py
mr-tz: moving to Layout could make sense yeah

floss/pipeline.py
mr-tz: ok

floss/main.py
mr-tz: see other comment: do we need the fields in Analysis still?

floss/pipeline.py
vee1e: done here only in https://github.com/mandiant/flare-floss/pull/1355/commits/2bb403666e635213ab9d9bf5aad84cc1617c96a8. scope:
- `static_strings` (accumulates static_runtime + layout/tag work)
- `language_strings` (Go/Rust)
- `vivisect`
- `find_features`
- `stack_strings`
- `tight_strings`
- `decoded_strings`

floss/layout/extract.py
vee1e: done, https://github.com/mandiant/flare-floss/pull/1355/commits/f62f41d1005e38ad9f667dd02b68868ed3381076

tests/test_main.py
vee1e: done https://github.com/mandiant/flare-floss/pull/1355/commits/2bf26d14769a6209ca20a4b1d74a99fecc14f098

floss/main.py
vee1e: yeah, `test_cli_args` tests the `False` paths in it. otherwise there's no way to turn those steps on/off. what we removed earlier was just the redundant `True` assignments, since the default is already "on".

floss/pipeline.py
vee1e: yeah agreed https://github.com/mandiant/flare-floss/pull/1355/commits/4235e7b8b3e219ad2e550c1cd8a791407cc14cd3

mr-tz (approved):
nice work, thanks

PR #1360, opened 2026-08-13
Folding of QS CLI - Part 2

vee1e:
## Summary

Iteration 4 of the FLOSS migration spec: the render-time filters, columns, summary output, and JSON error surface.

## What's included

**Render consolidation (§2.3)**
- Layout-aware string rendering moved into `floss/render/layout.py`; `default.py` orchestrates the classic and recovered-string sections.
- Centered lowercase headings for the stack/tight/decoded sections (§1.2).

**Filtering flags (§3.1.2)**
- Symmetrical `--section/--no-section`, `--structure/--no-structure`, `--tag/--no-tag`, plus `--interesting`. Include/exclude pairs are mutually exclusive; values OR across repeats.
- `--query` (repeatable, ORed regexes) and `--max-strings` (per-section, relevance-ordered) (§3.2.1).
- `--columns tags offset structure encoding` (§3.2.2).
- `--string-type`/`--no-string-type` now accept `language` and `all`; language extraction is independent of static (§3.4.3).

**Output (§3.3)**
- `--summary` (metadata, counts, section counts, tag histogram, high-value strings) with escaped/sanitized strings.
- Structured JSON errors on STDERR in JSON mode; text-mode errors also on STDERR.
- JSON top-level keys emitted `metadata`-first so saved documents stay detectable.

## Tests

- New `tests/test_render_filters.py` covering the filters, columns, summary, and main-level flags.
- Additions to `test_cli_args.py`, `test_load.py`, `test_main.py`, `test_results_json_roundtrip.py`.

Caching (spec §3.5) is intentionally out of scope for this batch.


floss/render/default.py
```diff
@@ -317,10 +333,31 @@ def render_heading(heading, console, verbose, disable_headers):
...
+
+         ─────────────────────
+               stack strings
+         ─────────────────────
+    """
+    if disable_headers:
+        return
+    style = ""
+    if verbose != Verbosity.DEFAULT:
+        style = "cyan"
```
mr-tz: why does Verbosity affect coloring?

floss/render/default.py
```diff
@@ -353,15 +390,41 @@ def get_color(color):
...
 
-def render(results: floss.results.ResultDocument, verbose, disable_headers, color):
+def render(
+    results: floss.results.ResultDocument,
+    verbose,
+    disable_headers,
+    color,
+    columns: Sequence[str] = DEFAULT_COLUMNS,
+    layout_filter: Optional[LayoutFilter] = None,
+    plain: bool = False,
```
mr-tz, line 407: can plain and verbose be combined?

floss/render/default.py
```diff
@@ -353,15 +390,41 @@ def get_color(color):
...
+                    layout = filtered
+
+            # when the user expressed tag intent (--tag or --interesting), don't let
+            # the default hide rules (e.g. #code, #reloc) drop strings the filter
+            # deliberately kept. tag-aware filtering already narrowed the set.
+            tag_rules = DEFAULT_TAG_RULES
+            if layout_filter is not None and (layout_filter.include_tags or layout_filter.interesting):
+                tag_rules = {tag: "default" if rule == "hide" else rule for tag, rule in DEFAULT_TAG_RULES.items()}
+
+            layout_view = hide_strings_by_rules(layout, tag_rules)
```
mr-tz: worth creating a helper function for this?!

floss/render/filter.py
```diff
@@ -0,0 +1,312 @@
...
+
+import re
+from typing import Set, List, Tuple, Optional, Sequence
+
+from floss.results import ResultLayout, ResultString
+from floss.tags.oss import DEFAULT_FILENAMES
+from floss.tags.filter import TagRules
+
+# noisy tags that the --interesting shortcut excludes
+NOISY_TAGS: Set[str] = {"#common", "#duplicate", "#code", "#reloc", "#code-junk"}
```
mr-tz, line 33: should we show these in the help to be transparent and keep these always in sync if we change them

floss/render/filter.py
```diff
@@ -0,0 +1,312 @@
...
+
+    ``import-table``, ``import_table``, and ``import table`` all compare equal.
+    """
+    return name.strip().lower().replace("_", "-").replace(" ", "-")
+
+
+def tag_matches(user_tag: str, string_tags: Sequence[str]) -> bool:
+    """true if ``user_tag`` (already normalized) matches any of ``string_tags``.
+
+    ``oss`` is a meta tag that matches any OSS library tag.
```
mr-tz: what if we want to add more meta tags? #common is also a meta tag right now in a sense, so is every directory in the tags/data/ path - can we leverage that?

floss/render/filter.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    normalized = user_tag
+    if normalized == "oss":
+        return any(normalize_tag(tag) in OSS_TAG_NAMES_NORMALIZED for tag in string_tags)
+    return any(normalize_tag(tag) == normalized for tag in string_tags)
+
+
+def is_arch_wrapper(layout: ResultLayout) -> bool:
```
mr-tz: improve name

floss/render/filter.py
```diff
@@ -0,0 +1,312 @@
...
+    @property
+    def active(self) -> bool:
+        return bool(
+            self.include_sections
+            or self.exclude_sections
+            or self.include_structures
+            or self.exclude_structures
+            or self.include_tags
+            or self.exclude_tags
+            or self.interesting
```
mr-tz, line 195: would it make sense to early on translate --interesting to the respective tags inclusions/exclusions and just use that throughout? con: some implied context later down in the code

floss/render/summary.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def render_summary(results: ResultDocument, color: str = "auto") -> str:
+    """render a concise summary of ``results`` as text.
+
+    The summary is a human- and agent-consumable view over the same
+    ``ResultDocument`` that ``--json`` emits.
+    """
+    sys.__stdout__.reconfigure(encoding="utf-8")  # type: ignore [union-attr]
+    console = Console(file=io.StringIO(), color_system=get_color(color), highlight=False, soft_wrap=True)
```
mr-tz, line 139: do we do this elsewhere? should we?

floss/render/summary.py
```diff
@@ -0,0 +1,181 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Concise, token-efficient summary of a FLOSS result document.
+
+The summary is built from the same ``ResultDocument`` as the ``--json`` output,
+so it is programmatic and consistent. It reports sample metadata, per-type
+string counts, section counts, tag histograms, and the strings that carry
+high-value (non-noisy) tags. Meant for human consumers and agents, so it is
+pretty printed with tables.
```
mr-tz: I think for agents we dont need table characters

floss/cli.py
```diff
@@ -159,6 +184,88 @@ def make_parser():
...
+            "exclude_sections",
+            "NAME",
+            None,
+            "do not show static strings in the given binary section(s)",
+        ),
+        (
+            "--structure",
+            "include_structures",
+            "NAME",
+            "e.g. import-table, export-table, rich-header, elf-header, macho-header; names are slugs and "
```
mr-tz: list all available here, can these be pulled from the code or a constant?

floss/cli.py
```diff
@@ -159,6 +184,88 @@ def make_parser():
...
+            "exclude_structures",
+            "NAME",
+            None,
+            "do not show static strings in the given binary structure(s)",
+        ),
+        (
+            "--tag",
+            "include_tags",
+            "TAG",
+            "e.g. winapi, crypto, or the oss meta tag",
```
mr-tz: list all or point to summary to see what's there

floss/cli.py
mr-tz: or point to summary to see what's there for a specific sample

floss/cli.py
```diff
@@ -159,6 +184,88 @@ def make_parser():
...
+            metavar=metavar,
+            default=[],
+            help=help_text,
+        )
+    filter_group.add_argument(
+        "--interesting",
+        action="store_true",
+        dest="interesting",
+        default=False,
+        help="exclude strings with only noisy tags: #common, #duplicate, #code, #reloc, #code-junk",
```
mr-tz: ah it's here, pull these from the constant

floss/main.py
```diff
@@ -59,6 +63,51 @@ def get_default_root() -> Path:
...
+    sys.stderr.write(json.dumps({"error": message, "code": code}) + "\n")
+
+
+JSON_SHORT_FLAG = re.compile(r"^-[a-zA-Z]*j")
+
+
+def json_requested(argv) -> bool:
+    """best-effort detection of a JSON output mode before argument parsing completes."""
+    argv = argv or []
+    return "--json" in argv or any(JSON_SHORT_FLAG.match(arg) for arg in argv)
```
mr-tz: it should be `--json` or `-j` only

floss/main.py
```diff
@@ -78,8 +128,24 @@ def main(argv=None) -> int:
...
+        if args.max_strings is not None and args.max_strings <= 0:
+            parser.error("--max-strings must be a positive integer")
+        for pattern in args.queries:
+            try:
+                re.compile(pattern)
+            except re.error as e:
+                parser.error("invalid --query regular expression %r: %s" % (pattern, e))
     except ArgumentValueError as e:
-        print(e)
+        if json_requested(argv):
```
mr-tz: reuse the json_mode variable above?

mr-tz: turn this into a helper function overall

floss/results.py
```diff
@@ -397,12 +399,23 @@ def check_set_string_types(results: ResultDocument, wanted_analysis: Analysis) -
 
 
 def filter_functions(results: ResultDocument, functions: List[int]) -> None:
-    filtered_scores = dict()
+    # a function is valid if it appears in any string category; decoding
+    # functions additionally have a score. don't require a decoding score for
+    # stack/tight-only functions.
```
mr-tz, line 404: good improvement

floss/results.py
```diff
@@ -423,3 +436,24 @@ def filter_string_len(results: ResultDocument, min_length: int) -> None:
...
+    )
+    results.strings.language_strings_missed = list(
+        filter(lambda s: len(s.string) >= min_length, results.strings.language_strings_missed)
+    )
+    if results.layout is not None:
+        results.layout = _filter_layout_string_len(results.layout, min_length)
+
+
+def _filter_layout_string_len(layout: ResultLayout, min_length: int) -> ResultLayout:
+    """recursively drop layout-tree strings shorter than min_length."""
```
mr-tz, line 467: are these extracted? why do we have to drop here if so? they shouldn't be captured in the first place

tests/test_cli_args.py
```diff
@@ -60,6 +60,62 @@ def test_args_analysis_type_conflict(exefile):
...
+def test_language_auto_accepted(exefile, capsys):
+    """--language auto is the default and is accepted by the parser."""
+    assert floss.main.main([exefile, "--language", "auto"]) == 0
+    capsys.readouterr()
+    assert floss.main.main([exefile]) == 0
+    capsys.readouterr()
+
+
+def test_filter_flags_sample_last_parses(exefile, capsys):
+    """multi-value filter flags need a -- terminator before the sample."""
```
mr-tz: do we need to test this?

tests/test_cli_args.py
```diff
@@ -60,6 +60,62 @@ def test_args_analysis_type_conflict(exefile):
...
+            sample=sample,
+            min_length=6,
+            analysis=Analysis(
+                enable_static_strings=False,
+                enable_stack_strings=False,
+                enable_tight_strings=False,
+                enable_decoded_strings=False,
+                enable_language_strings=True,
+            ),
+            prompt_deobfuscation=False,
```
mr-tz: what's this?

floss/utils.py
```diff
@@ -539,6 +539,19 @@ def is_string_type_enabled(type_, disabled_string_types, enabled_string_types):
         return True
 
 
+def expand_string_types(string_types):
+    """expand the `all` alias into the full set of concrete string types."""
```
mr-tz: can this be done in a simpler fashion?

floss/results.py
mr-tz: for later runs on a cached report I could see this is needed maybe

tests/test_main.py
```diff
@@ -31,6 +31,8 @@ def test_main_help(capsys):
         out = capsys.readouterr().out
         assert "usage:" in out
         assert "--json" in out
+        assert "import-table" in out
+        assert "export-table" in out
```
mr-tz:
```suggestion
```

floss/render/summary.py
```diff
@@ -0,0 +1,181 @@
...
+def render_summary(results: ResultDocument, color: str = "auto") -> str:
+    """render a concise summary of ``results`` as text.
+
+    The summary is a human- and agent-consumable view over the same
+    ``ResultDocument`` that ``--json`` emits.
+    """
+    sys.__stdout__.reconfigure(encoding="utf-8")  # type: ignore [union-attr]
+    console = Console(file=io.StringIO(), color_system=get_color(color), highlight=False, soft_wrap=True)
+
+    console.print(f"FLOSS SUMMARY (version {results.metadata.version})\n")
```
mr-tz, line 141: TBD does --summary only focus on static (the data is only that right now?!) or also leverage more of the decoded strings data if available?

floss/render/summary.py
```diff
@@ -0,0 +1,181 @@
...
+    section_counts: Dict[str, int] = Counter()
+    tag_histogram: Counter = Counter()
+    high_value: List[ResultString] = []
+
+    def walk(node: ResultLayout, section: str, depth: int) -> None:
+        section_counts[section] += len(node.strings)
+        for s in node.strings:
+            tag_histogram.update(s.tags)
+            if any(tag not in NOISY_TAGS for tag in s.tags):
+                high_value.append(s)
```
mr-tz, line 71: this should use the relevance algorithm instead?!

floss/utils.py
vee1e: yeah, moved `CONCRETE_STRING_TYPES` and `StringType` into a module both can import without circularity in https://github.com/mandiant/flare-floss/pull/1360/commits/9ddc1ac

tests/test_cli_args.py
vee1e: It verifies the fix where Go/Rust extraction no longer depends on `enable_static_strings` i.e. `--string-type language` (or `--no-string-type static`) still extracts language strings. it calls the pipeline directly on a Go sample rather than going through `main` to avoid the deobfuscation prompt/TTY path. added a short comment to the test in https://github.com/mandiant/flare-floss/pull/1360/commits/9ddc1ac

tests/test_cli_args.py
vee1e: i added it because `nargs="+"` silently swallows a trailing positional sample, and the epilog documents sample-last usage. Without a test, someone could "fix" the docs or parsing and regress the `--` terminator behavior. if you think it's over-testing a CLI-ergonomics quirk we can drop it, but i'd personally keep at least one assertion that the documented sample-last form parses.

floss/results.py
vee1e: yup, this drop is needed only for cached documents (a saved JSON may contain strings shorter than a new `-n`, or be filtered with a larger `-n` than the analysis used). Since we moved to auto-detecting saved documents, load-time re-filtering is the correct place, not extraction. made the comment explain this is the load path in https://github.com/mandiant/flare-floss/pull/1360/commits/9ddc1ac

floss/main.py
vee1e: added `report_error` in https://github.com/mandiant/flare-floss/pull/1360/commits/8957bb0

floss/main.py
vee1e: sure, removed in https://github.com/mandiant/flare-floss/pull/1360/commits/8957bb02ec4344f48193e780a54d215fb34fdcab

floss/cli.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1360/commits/4b53c625f220ab9e608f5a692d97e53a5718e214

floss/cli.py
vee1e: added in https://github.com/mandiant/flare-floss/pull/1360/commits/c5b6eea a `KNOWN_STRUCTURE_SLUGS` to pull from and modify easily and added it to help in https://github.com/mandiant/flare-floss/pull/1360/commits/4b53c625f220ab9e608f5a692d97e53a5718e214

floss/render/summary.py
vee1e: added a docstring in https://github.com/mandiant/flare-floss/pull/1360/commits/7f03cc2, should we add a `--summary -j` option for a json summary? i think we discussed this in the spec before.

floss/render/summary.py
vee1e: https://github.com/mandiant/flare-floss/pull/1360/commits/7f03cc2 reused the existing style helpers instead.

floss/render/filter.py
vee1e: yeah i considered exactly this. but `interesting` as its own flag rather than expanding it into `exclude_tags` is because the semantics differ, `--interesting` drops a string only when it has no non-noisy tag, whereas `--no-tag common` drops any string carrying `#common`.

so if we expanded `--interesting` into per-tag exclusions up front, we'd lose that "at least one meaningful tag" behavior added a comment in https://github.com/mandiant/flare-floss/pull/1360/commits/c5b6eea the attribute explaining this so the context isn't implicit.

floss/render/filter.py
vee1e: renamed to `is_macho_arch_wrapper` in https://github.com/mandiant/flare-floss/pull/1360/commits/c5b6eea75a6b6b0e5f45c06eda279b3ea1959ff9

floss/render/filter.py
vee1e: fair point, `#common`, `#code`, etc. are produced by the tag sources in `floss/tags/data/*` so there's a natural mapping: each directory is a tag family. I think the right generalization is to derive meta tags from the tag-source directories (winapi → `#winapi`, oss → all `#<lib>` tags, gp → `#common`, etc.) rather than hardcoding `oss`. what do you think?

floss/render/filter.py
vee1e: addressed in https://github.com/mandiant/flare-floss/pull/1360/commits/4b53c625f220ab9e608f5a692d97e53a5718e214

floss/render/default.py
vee1e: agreed, added `effective_tag_rules` in https://github.com/mandiant/flare-floss/pull/1360/commits/16a090f566f6fa5067d31c47c8ca15306a898d44

floss/render/default.py
vee1e: yes, `plain` selects the flat (non-layout) view, `verbose` controls detail level within either view. so `--plain -v` would render the flat list with offsets/functions, and `--plain` alone renders it tersely.

floss/render/default.py
vee1e: the cyan heading in verbose mode is legacy behavior carried over from the pre-existing `render_heading`. updated in https://github.com/mandiant/flare-floss/pull/1360/commits/16a090f566f6fa5067d31c47c8ca15306a898d44

floss/render/summary.py
vee1e: right now `--summary`'s layout-derived sections cover static strings only because those need the layout/tag data that only statics carry. but **string-counts** reports stack/tight/decoded/language counts from the document, so it does show them. those strings aren't tagged, so they wouldn't appear in a tag histogram anyway. i think the summary's static focus is correct for today, if we later want decoded strings highlighted but we should keep it as-is for now. wdyt?

floss/render/summary.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1360/commits/7f03cc275ba36899b308ede942ddfc8d8d0a633b

floss/render/filter.py
mr-tz: agreed

floss/render/summary.py
mr-tz: commit seems unrelated

floss/render/summary.py
mr-tz: we agreed to not use json for agents I thought

mr-tz: id keep the table but remove the unneeded chars maybe

floss/main.py
mr-tz: do we need `-j ` (with the space) to not catch other `-j..` args?

tests/test_cli_args.py
mr-tz: I think we need doc in the code in a appropriate place also?

floss/render/summary.py
mr-tz: then default --summary to be static only? or require that to be called explicitly? otherwise we spin decoding strings for a summary that doesnt benefit from the extra work

floss/render/summary.py
vee1e: yeah the commit bundled several changes. it also contained the relevance ordering and the module-docstring change, so the heading-style bit was bundled with unrelated changes instead of being its own atomic commit. My mistake on commit granularity. The `heading_style` reuse is small and self-contained.

floss/main.py
vee1e: right now it checks `"-j" in argv`, which would also match something like `-jsonfile` if it ever appeared as an argument (argparse wouldn't normally produce that for our flags, but a value token could).

since we only define `-j` exactly, I should match it precisely: `arg == "-j"`. I'll change it to check `any(a == "-j" for a in argv)` so clustered/adjacent `-jxxx` tokens aren't misdetected.

tests/test_cli_args.py
vee1e: we can add a docstring to `test_language_extraction_independent_of_static`, explaining it gives the pipeline directly to avoid the deobf prompt. If you mean documentation in the source not just the test, the behavior itself is documented in `pipeline.py` where language extraction was pulled out of the `enable_static_strings` block.

that comment explains language strings are independent of static. If you meant a comment in the test module header or the spec, let me know where you'd like it and we'll add it.

mr-tz (commented):
Overall, an excellent iteration. The fixes implemented in the latest batch of commits cleanly resolve the issues pointed out in the previous review. I completely agree with your rejection of the two invalid points regarding Mach-O section nodes and the argparse positional arg behavior; those design constraints are valid and well-reasoned.

All newly added functionality, from JSON fixed-key fallbacks to deduplicating expanded string types, looks robust and is well-tested. 

There is just one lingering edge-case in the new `--language` override guard that should be addressed before merging:

### 1. The manual `--language` override can still be clobbered by a false-positive auto-detection

In `096e4428` (`floss/pipeline.py`), you added this guard:
```python
        if selected_lang in (Language.GO, Language.RUST, Language.DOTNET) and lang_id == Language.UNKNOWN:
            # a concrete manual selection wins over failed auto-detection...
            results.metadata.language = selected_lang.value
```
While this fixes the issue when auto-detection *fails* (`Language.UNKNOWN`), it falls short when auto-detection *succeeds but detects the wrong language*.

If a user explicitly runs `floss --language go`, but `identify_language_and_version` falsely identifies the binary as Rust, `lang_id == Language.UNKNOWN` evaluates to `False`. The `if` block is skipped, execution falls to the `else` block, and `results.metadata.language` is set to `"rust"`. This completely ignores the user's manual `--language go` override. (Currently, the pipeline just logs a warning but proceeds to use the auto-detected language anyway).

**Recommendation:** Drop the `and lang_id == Language.UNKNOWN` condition. A concrete manual selection should unilaterally win over auto-detection:

```python
        if selected_lang in (Language.GO, Language.RUST, Language.DOTNET):
            results.metadata.language = selected_lang.value
            results.metadata.language_version = lang_version if lang_id == selected_lang else ""
        else:
            ...
```

### Verdict

Once the language override condition is simplified to unconditionally honor the user's concrete choice, this is **ready to be merged**. Great work!

floss/render/summary.py
vee1e: sure, done in https://github.com/mandiant/flare-floss/pull/1360/commits/dbfa870

floss/main.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1360/commits/0785a29

floss/render/summary.py
vee1e: done in https://github.com/mandiant/flare-floss/commit/01a808c now tables are markdown style, easy and helpful to parse for agents

floss/cli.py
```diff
@@ -222,6 +331,13 @@ def make_parser():
 
     output_group = parser.add_argument_group("rendering arguments")
     output_group.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
+    output_group.add_argument(
+        "--summary",
+        action="store_true",
+        default=False,
+        help="emit a concise summary (metadata, counts, tag histogram, high-value strings); "
+        "static-only by default unless string types are explicitly selected",
```
mr-tz, line 340: Manual selection of other types doesn't offer any benefit?! Should we just reject that for now?

floss/render/default.py
mr-tz: Hm I think overall it's a bit complicated with all the types and options and what affects what. Please at least open an issue if we don't address in this PR.

floss/render/filter.py
mr-tz: Is this done?

floss/render/filter.py
```diff
@@ -0,0 +1,336 @@
...
+
+    relevance order within a section:
+      1. strings with a highlighted tag first
+      2. then untagged strings
+      3. then strings with any non-noisy tag
+      4. then the rest (only noisy tags)
+    within each group, ascending by offset.
+    """
+    has_highlight = any(tag_rules.get(tag) == "highlight" for tag in s.tags)
+    has_non_noisy = any(tag not in NOISY_TAGS for tag in s.tags)
```
mr-tz, line 108: This logic is off as common strings will be shown if they also have a different tag.

floss/cli.py
vee1e: agreed, done https://github.com/mandiant/flare-floss/pull/1360/commits/83f5788d324ca1b9c5ea0743555d5f2dba610e09 --language choices to auto, go, rust, none acc to spec §3.4.7. `--language dotnet` is now rejected (invalid choice) since .NET extraction is unimplemented.

floss/render/filter.py
vee1e: > Hm I think overall it's a bit complicated with all the types and options and what affects what. Please at least open an issue if we don't address in this PR. 

 https://github.com/mandiant/flare-floss/issues/1362

> Is this done?

~tracked for now, i'll implement it along with the correctness bugs from below.~ done in https://github.com/mandiant/flare-floss/pull/1360/commits/5aee9ab3790ac2c0f3f5ff2b163134b1f01b20f8 and https://github.com/mandiant/flare-floss/pull/1360/changes/78f14686c84072de1059968d1abf589256aebe3b

> This logic is off as common strings will be shown if they also have a different tag.

fixed, it now drops any string carrying a noisy tag, even alongside a non-noisy tag, matching the spec's "excludes the noisy tags" wording.

floss/cli.py
```diff
@@ -177,8 +287,10 @@ def make_parser():
     advanced_group.add_argument(
         "--language",
         type=str,
-        choices=[l.value for l in Language if l != Language.UNKNOWN],
-        default=Language.UNKNOWN.value,
+        # spec 3.4.7: auto, go, rust, none. dotnet is not offered because .NET
+        # string extraction is not implemented yet.
```
```suggestion
```

floss/cli.py
mr-tz: this comment is on the summary flag, please revisit/revert

floss/main.py
```diff
@@ -103,8 +172,15 @@ def main(argv=None) -> int:
...
-    enabled_string_types = list(args.enabled_string_types or [])
+    disabled_string_types = expand_string_types(list(args.disabled_string_types or []))
+    enabled_string_types = expand_string_types(list(args.enabled_string_types or []))
+
+    if args.summary and not disabled_string_types and not enabled_string_types:
+        # the summary's layout-derived sections cover static strings only, so
+        # don't spin up the slow deobfuscation for stack/tight/decoded unless
+        # the user explicitly requested them
+        logger.info("--summary is static-only; skipping stack/tight/decoded extraction")
+        disabled_string_types.extend([StringType.STACK.value, StringType.TIGHT.value, StringType.DECODED.value])
```
mr-tz, line 199: WDYT on summary only working for static not allowing other args basically?

mr-tz: maybe rename to --static-summary?

floss/results.py
```diff
@@ -417,9 +430,36 @@ def filter_functions(results: ResultDocument, functions: List[int]) -> None:
 
 
 def filter_string_len(results: ResultDocument, min_length: int) -> None:
+    """filter strings below min_length.
+
+    Applied when loading a saved results document (and future cache hits):
+    extraction already respects min_length, but a document loaded with a
+    different -n may contain shorter strings, so re-filter here.
```
mr-tz, line 437: we should warn/abort if the min length selected is below what's available to not confuse users
e.g. extraction ran with min 6 but later the user says min is 4

floss/utils.py
```diff
@@ -539,6 +539,24 @@ def is_string_type_enabled(type_, disabled_string_types, enabled_string_types):
         return True
 
 
+def expand_string_types(string_types):
```
mr-tz, line 542: this seems very complex for this task?
maybe all can only be used by itself?

what are the invocations you're envisioning here to catch/guard against?

tests/test_cli_args.py
```diff
@@ -60,6 +60,133 @@ def test_args_analysis_type_conflict(exefile):
...
+def test_language_auto_accepted(exefile, capsys):
+    """--language auto is the default and is accepted by the parser."""
+    assert floss.main.main([exefile, "--language", "auto"]) == 0
+    capsys.readouterr()
+    assert floss.main.main([exefile]) == 0
+    capsys.readouterr()
+
+
+def test_filter_flags_sample_last_parses(exefile, capsys):
+    """multi-value filter flags need a -- terminator before the sample."""
```
mr-tz: this is testing that the user needs to call it this way? remove

tests/test_cli_args.py
```diff
@@ -60,6 +60,133 @@ def test_args_analysis_type_conflict(exefile):
     assert floss.main.main([exefile, "--string-type", "stack", "--no-string-type", "tight"]) == -1
 
 
+def test_language_auto_accepted(exefile, capsys):
+    """--language auto is the default and is accepted by the parser."""
```
mr-tz: should we then test all types and all args? remove

tests/test_cli_args.py
```diff
@@ -60,6 +60,133 @@ def test_args_analysis_type_conflict(exefile):
...
+    ):
+        assert floss.main.main(args + ["--", exefile]) == 0, args
+        capsys.readouterr()
+
+
+def test_language_extraction_independent_of_static():
+    """language strings are extracted even when static strings are disabled.
+
+    drives the pipeline directly (not via floss.main) to avoid the language
+    deobfuscation prompt/TTY path, on a Go sample whose language is detectable.
```
mr-tz: can we add a flag or detection for this? like `-y` or infer from the other options to not prompt?

mr-tz: the below tests would also benefit and/or need to be improved

tests/test_render_filters.py
mr-tz, line 1: did you double check all these?

floss/main.py
vee1e: > WDYT on summary only working for static not allowing other args basically?

yeah agreed would prevent shadow args not working

> maybe rename to --static-summary?

i'd push back on this, are we only ever going to do `--summary` for static only or are there plans to extend it to decoded/stack/tight? if the latter is true we should keep it static but have it documented that its only static for now + the above.

floss/results.py
vee1e: yeah those short strings are already gone at extraction time, so re-filtering is misleading. now warn when the requested `-n` is below the stored extraction min length, pointing out they can't be recovered. fixed in https://github.com/mandiant/flare-floss/pull/1360/commits/cc66e7e

floss/utils.py
vee1e: right, `all` combining with other types was the whole reason for the dedupe. restricted it so it can only be used alone (parse-time error otherwise) and simplified `expand_string_types` with no more chain/dedupe. fixed in https://github.com/mandiant/flare-floss/pull/1360/commits/7932b3d

tests/test_cli_args.py
vee1e: agreed, that's just locking in an argparse convention. removed in https://github.com/mandiant/flare-floss/pull/1360/commits/cac88eb

tests/test_cli_args.py
vee1e: fair, same reasoning. if we test one default we'd have to test them all. removed https://github.com/mandiant/flare-floss/pull/1360/commits/8ef8bc1

tests/test_cli_args.py
vee1e: yup, added `-y`/`--yes` flag in https://github.com/mandiant/flare-floss/pull/1360/commits/2c39629. it suppresses the deobfuscation prompt and defaults to not running deobfuscation (same as the prompt's default `n`)

vee1e: and the language extraction test now goes through `floss.main` with `-y` instead of calling the pipeline directly. https://github.com/mandiant/flare-floss/pull/1360/commits/4731c8d

floss/main.py
mr-tz: we could extend it at some point

vee1e: alright, added the gate for static and other args in https://github.com/mandiant/flare-floss/commit/f73c7ef but lets keep it `--summary` for now so other paths are there for it in the future

tests/test_render_filters.py
vee1e: ran a few reviews on this + manual pass: https://github.com/mandiant/flare-floss/pull/1360/commits/e1bc7f9f83e06ac21bfdf935b239c51c432c3a02

removed 2 redundant tests:
- `test_filter_tag_family_winapi` was duplicating `test_filter_by_tag` (the `winapi` family is a single-element set, so the family path and direct-tag path are indistinguishable).
- `test_main_columns_accumulate` was duplicating `test_main_columns_flag` and would pass even if accumulation broke.

added 3 tests:

- `test_filter_tag_normalization`
- `test_summary_section_counts_thread_fat_macho` to test `analyze_layout` Mach-O fat-arch section threading
- `test_layout_none_warns_filters_ignored`, the warning emitted when an active layout filter has no layout tree.

floss/results.py
mr-tz: abort instead?

tests/test_cli_args.py
mr-tz: there's more tests that could use the updated calling

floss/results.py
vee1e: done https://github.com/mandiant/flare-floss/pull/1360/commits/f3263ff186dc1d74e6292940455a57dd024e5181

tests/test_cli_args.py
vee1e: https://github.com/mandiant/flare-floss/commit/7bdd55cdc4f16a40a75a2450b8459953b54c99c5#commitcomment-196524254


> the only tests left that drive the pipeline directly are:
> - `test_no_layout_yields_classic_static`
> - `test_no_tags_skips_tag_databases`
> 
> they set internal `enable_layout`/`enable_tags` flags that have no CLI flag, so they can't go through `-y` plus `-j`. so not applicable, reverted the commit entirely.

if i missed any other tests let me know and we'll change it to use `-y`.

tests/test_render_filters.py
mr-tz: ok thanks

tests/test_cli_args.py
mr-tz: what about

test_manual_language_override_wins_over_auto_detect

test_manual_language_override_beats_wrong_auto_detect

mr-tz: got it

mr-tz: although how could this happen if we cannot force it via args?

floss/render/default.py
mr-tz: https://github.com/mandiant/flare-floss/issues/1362

floss/cli.py
vee1e: > the `--language` code was also an issue hence why i didn't revert it, i don't think there's any advantage to letting the user select a language imo

tests/test_cli_args.py
vee1e: > although how could this happen if we cannot force it via args?

it can though. `--language go/rust` is a real CLI flag, so a user genuinely can force the override from the command line (e.g. `--language go` on a binary auto-detect would misidentify).

the only thing the test can't do via args is fake the auto-detection result (the monkeypatch that makes it say rust). that's just a test convenience to simulate a real misdetection case. so no code change needed; the two tests are legit and already on the new pattern.

mr-tz (approved):
another big step, thanks!

PR #1367, opened 2026-08-17
Folding of QS CLI - Part 3

vee1e:
## Summary

Implemented §3.5 caching.

## Key decisions

- Cache files use the format `{sha256}-{version}.json`.
- `covers()` checks that a cached result matches the requested analysis settings. If it does not, the cache is ignored and the analysis runs normally.
- Caching is disabled for non-default analysis variants to avoid returning a mismatched document.
- Cache writes use locks and are atomic. If the lock cannot be acquired, the write is skipped with a warning.
- `FLOSS_CACHE_DIR` is supported. Without it, the cache uses `platformdirs.user_cache_dir("floss")`.
- Invalid, corrupted, checksum-mismatched, or version-mismatched cache entries are ignored.

### DIFF
--- a/floss/cache.py
+++ b/floss/cache.py
@@ -64,20 +64,20 @@ _ENABLED_VALUES = ("1", "true", "yes", "y")
 def cache_enabled() -> bool:
     """Whether result caching is enabled.
 
-    ``FLOSS_CACHE_ENABLE`` disables caching when set to 0 (or false/no/n);
-    caching is enabled by default.
+    ``FLOSS_CACHE_ENABLE`` disables caching when set to 0 (or false/no/n,
+    case-insensitive); caching is enabled by default.
     """
-    return os.environ.get(ENV_CACHE_ENABLE, "1") not in _DISABLED_VALUES
+    return os.environ.get(ENV_CACHE_ENABLE, "1").lower() not in _DISABLED_VALUES
 
 
 def cache_refresh() -> bool:
     """Whether the current run should bypass the cache and overwrite its entry.
 
-    ``FLOSS_CACHE_REFRESH=1`` (or true/yes/y) forces a re-analysis: the cached
-    document is ignored and the fresh result is stored on top of it. Unset by
-    default.
+    ``FLOSS_CACHE_REFRESH=1`` (or true/yes/y, case-insensitive) forces a
+    re-analysis: the cached document is ignored and the fresh result is stored
+    on top of it. Unset by default.
     """
-    return os.environ.get(ENV_CACHE_REFRESH, "") in _ENABLED_VALUES
+    return os.environ.get(ENV_CACHE_REFRESH, "").lower() in _ENABLED_VALUES
 
 
 def get_cache_dir() -> Path:
--- a/floss/cache.py
+++ b/floss/cache.py
@@ -103,15 +103,28 @@ def load(cache_dir: Path, key: str, sha256: str, version: str) -> Optional[Resul
         doc = ResultDocument.parse_file(path)
     except (OSError, UnicodeDecodeError, ValueError) as e:
         logger.warning("dropping invalid cache entry %s: %s", path.name, e)
-        path.unlink(missing_ok=True)
+        _drop_cache_entry(path)
         return None
 
     if doc.metadata.sha256 != sha256 or doc.metadata.version != version:
         logger.warning("dropping stale cache entry %s (checksum/version mismatch)", path.name)
-        path.unlink(missing_ok=True)
+        _drop_cache_entry(path)
         return None
 
     return doc
+
+
+def _drop_cache_entry(path: Path) -> None:
+    """Best-effort removal of a stale cache entry; never raises.
+
+    The entry may be held open by another reader or an antivirus scanner (e.g.
+    on Windows), so removal can fail with a PermissionError. Leave it and let
+    the caller continue.
+    """
+    try:
+        path.unlink(missing_ok=True)
+    except OSError as e:
+        logger.warning("could not remove cache entry %s: %s", path.name, e)
 
 
 def store(cache_dir: Path, key: str, doc: ResultDocument) -> bool:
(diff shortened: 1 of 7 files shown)


floss/cache.py
```diff
@@ -0,0 +1,231 @@
...
+    (shorter strings were dropped at extraction time and cannot be recovered).
+    """
+    for field in STRING_TYPE_FIELDS:
+        if getattr(wanted, field) and not getattr(cached.analysis, field):
+            return False
+
+    if wanted.enable_layout and not cached.analysis.enable_layout:
+        return False
+    if not wanted.enable_layout and cached.layout is not None:
+        return False
```
mr-tz: cant we drop the layout later? things that are not wanted but there shouldnt be an issue?

floss/cache.py
```diff
@@ -0,0 +1,231 @@
...
+    if wanted.enable_tags and not cached.analysis.enable_tags:
+        return False
+    if not wanted.enable_tags and cached.analysis.enable_tags:
+        return False
```
mr-tz: same here

floss/cache.py
```diff
@@ -0,0 +1,231 @@
...
+
+    try:
+        if sys.platform == "win32":
+            import msvcrt
+
+            msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
+        else:
+            import fcntl
+
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
mr-tz, line 310: are there os agnostic alternatives for this?

floss/cache.py
```diff
@@ -0,0 +1,231 @@
...
+    except OSError:
+        os.close(fd)
+        return None
+
+    return fd
+
+
+def _release_lock(fd: int, lock_path: Path) -> None:
+    """Release a lock previously acquired by :func:`_acquire_lock`."""
+    if sys.platform == "win32":
```
mr-tz, line 320: no else needed?

floss/main.py
```diff
@@ -170,6 +172,18 @@ def main(argv=None) -> int:
 
     set_log_config(args.debug, args.quiet)
 
+    # caching applies to the default analysis variant only: an explicit format,
+    # language, or custom signatures change the analysis, so the content-
+    # addressed cache could return a mismatched document. disable it for those.
```
mr-tz, line 177: how much more complex is it to support any invocation and warn on missing further processing?

does that make cache file handling annoying? can I overwrite a cached file easily on purpose with other args etc.

floss/pipeline.py
```diff
@@ -311,6 +316,17 @@ def analyze(options: Options) -> Optional[ResultDocument]:
     results.metadata.sha1 = hashlib.sha1(sample_buf).hexdigest()
     results.metadata.sha256 = hashlib.sha256(sample_buf).hexdigest()
 
+    # result caching: keyed on the sample bytes + FLOSS version. on a valid hit
+    # we skip extraction, layout, and tagging entirely and render from the cache.
+    # --analyze-functions changes which functions are analyzed, so it is a miss.
```
mr-tz: how does the --analyze-functions comment relate to the code here?

tests/test_cache.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def test_get_cache_dir_override(monkeypatch, tmp_path):
+    monkeypatch.setenv(floss.cache.ENV_CACHE_DIR, str(tmp_path))
+    assert floss.cache.get_cache_dir() == tmp_path
+
+
+def test_get_cache_dir_default(monkeypatch):
+    monkeypatch.delenv(floss.cache.ENV_CACHE_DIR, raising=False)
+    assert str(floss.cache.get_cache_dir()).endswith("floss")
```
mr-tz: is that a good enough check?

vee1e: nope, i updated it in https://github.com/mandiant/flare-floss/pull/1367/commits/fba96b0fb6dd4f42d61ccd96a2cf52a4f8fb4053. The `endswith` check was also wrong on Windows, so it now asserts equality with `platformdirs.user_cache_dir("floss")`, exactly what `get_cache_dir` returns with no override.

floss/pipeline.py
vee1e: it documents the `not options.functions` guard on the next line, `--analyze-functions` restricts which functions are analyzed, and the key doesn't cover that, so a hit could wrongly serve a full-function result.

it's defense-in-depth since `main` already sets `cache_dir=None` for `--analyze-functions`.

i moved the comment inline with the condition in https://github.com/mandiant/flare-floss/pull/1367/commits/bc94b6cb27ff95854b4119e5842194cd537a16b7 so the connection is obvious.

floss/main.py
vee1e: full-variant support is the big one: the key is only `{sha256}-{version}`, so doing it properly means either adding the variant to the key or partial-cache reconciliation (detect what's missing, run just those steps, merge into the cached doc). that's a meaningful chunk of work beyond §3.5's full-document cache.

the overwrite gap is fair: identical-args runs overwrite via the atomic replace, but for a different variant there is no write at all, so the only forced-refresh today is deleting `{sha256}-{version}.json` (`FLOSS_CACHE_ENABLE=0` disables read *and* write).

we can add an explicit way to force a refresh (`--no-cache`/`--refresh`) or document the `rm`, which do you prefer?

floss/cache.py
vee1e: yup. on POSIX `flock` is released when the last fd to the open file description is closed, so the `os.close(fd)` after the if-block handles it. Only Windows needs the explicit `LK_UNLCK` before close.

floss/cache.py
vee1e: technically yes, `filelock` is the standard cross-platform one. it's not currently a dependency, which is why I kept the inline branch: we add ~15 lines vs. adding a dep to a project that also ships as a PyInstaller binary. wdyt?

floss/cache.py
vee1e: same as above, for tags it's a real miss: a tags-enabled cached doc has had
false-positive strings removed at extraction, so it's a proper subset and can't be
recovered. keeping it a miss.

floss/cache.py
vee1e: fair, for layout we *could* drop it and treat this as a hit. rendering would fall back to the flat static list from `strings.static_strings`, which is the same data.

the tags case is different: a doc built with tags enabled has already had false-positive library strings removed (`remove_false_positive_lib_strings`) so it can't faithfully serve a tags-off request. also, the CLI always enables layout+tags, so these branches only work for `Options` callers.

i dropped the layout in `materialize` and kept tags as a miss in https://github.com/mandiant/flare-floss/pull/1367/commits/bc94b6cb27ff95854b4119e5842194cd537a16b7

mr-tz: See some comments in the commit, thanks

floss/cache.py
mr-tz: will defer to later track via issue please

floss/cache.py
mr-tz: ok, that's fine as is then

floss/main.py
mr-tz: maybe we need to add (hashed) options to the key then? maybe defer to later improvements tracked in a TBD issue

on the rm option I don't like adding yet another option - so lets avoid if possible, can we leverage the ENV vars for this?

floss/cache.py
vee1e: this actually landed already in https://github.com/mandiant/flare-floss/pull/1367/commits/1571c403e08b7fac7314f4f6fade5630e9e52216: `covers` now lets a tags-enabled cache serve a no-tags request, and `materialize` redacts the tags across the static, language, and layout strings. `remove_false_positive_lib_strings` only strips tags, never strings, so the cached data is a faithful superset. no tracking issue needed for this one.

floss/main.py
vee1e: Agreed on both:

- opened #1369 to track hashing the analysis options into the cache key.
- implemented FLOSS_CACHE_REFRESH=1 in https://github.com/mandiant/flare-floss/pull/1367/commits/a5e7aa00976557b765d56d23d30f5147ab0ed5bd, it forces a miss, re-analyzes, and overwrites the cache entry. Documented in --help.

mr-tz: ok thanks

tests/test_cache.py
mr-tz: good

floss/cli.py
```diff
@@ -143,6 +143,11 @@ def make_parser():
 
           extract strings from a binary written in Go (if automatic language identification fails)
             floss --language go program.exe
+
+        environment variables:
+          FLOSS_CACHE_DIR       directory for the analysis result cache (default: platform cache directory)
+          FLOSS_CACHE_ENABLE    set to 0 to disable result caching (default: enabled)
+          FLOSS_CACHE_REFRESH   set to 1 to ignore cached results and overwrite the entry (default: unset)
```
```suggestion
          FLOSS_CACHE_REFRESH   set to 1 to ignore cached results and overwrite the entry (default: disabled)
```

mr-tz (approved):

Issue #1375, opened 2026-08-18
Fix 10 correctness/regression issues found in review of #1372 before merging QUANTUMSTRAND into master


mr-tz:
Reviewed the full diff of #1372 (`quantumstrand` → `master`, closes #943). CI is green across the board, but a source-level review turned up 10 issues, several of which silently produce wrong output or regress existing behavior. Filing these to track fixes before merge; plan is to address all of them and rebase the branch on top of `master` afterward.

### High priority — silently wrong output / behavior contradicts docs

- [x] **XOR-decoded PE/ELF layouts fail the structured-layout check** (`floss/enrich.py:34`)
  `is_structured_layout()` does an exact match on `"pe"`/`"elf"`, but an XOR-obfuscated header gets `" (XOR decoded with key: 0x...)"` appended to `layout.name` (`floss/layout/pe.py:342`, `floss/layout/elf.py:260`). The check then fails and `compute_layout()` silently falls back to classic static strings — losing section names, code/reloc/XOR tags, and structure annotations for exactly the obfuscated-malware case FLOSS is built to handle.

- [x] **`-y`/`--yes` doesn't skip deobfuscation for Go/Rust binaries** (`floss/pipeline.py:396`)
  The flag's help text says deobfuscation "defaults to not running it," but that only happens inside the interactive-prompt branch, which `-y` bypasses entirely. Only `.NET` gets `enable_stack/tight/decoded_strings` force-disabled elsewhere (`floss/pipeline.py:387-389`); Go/Rust binaries keep the `Analysis()` default of `True`. A non-interactive/CI run with `-y` on a Go or Rust sample runs full deobfuscation anyway.

- [ ] **Static-strings path now reads the whole file into memory, bypassing the large-file guard** (`floss/pipeline.py:309`)
  `sample.read_bytes()` runs unconditionally before the `MAX_FILE_SIZE`/`--large-file` check (`floss/pipeline.py:504`), which now only guards the deobfuscation stage. `master`'s equivalent path used `mmap` (`floss/utils.py: get_static_strings`) specifically to avoid this. Regression risk: OOM or a large perf hit on big samples even for `--string-type static`-only runs.

### Medium priority — regressions / silent misbehavior

- [x] **`--disable-progress` CLI flag silently removed** (`floss/cli.py`)
[...]

PR #1376, opened 2026-08-18
Fix review issues from #1375 (quantumstrand correctness/regression fixes)

vee1e:
Addresses the valid issues from https://github.com/mandiant/flare-floss/issues/1375

### DIFF
--- a/floss/layout/base.py
+++ b/floss/layout/base.py
@@ -235,15 +235,19 @@ class Layout(BaseModel, abc.ABC):
         such as a PE file and all its data.
         """
         if structures:
-            for string in self.strings:
-                for structures_by_address in structures:
-                    structure = structures_by_address.get(string.offset)
-                    if structure:
-                        string.structure = structure.name
-                        break
+            self._mark_string_structures(structures)
 
         for child in self.children:
             child.mark_structures(structures=structures, **kwargs)
+
+    def _mark_string_structures(self, structures) -> None:
+        """attach the first matching structure name to this node's own strings."""
+        for string in self.strings:
+            for structures_by_address in structures:
+                structure = structures_by_address.get(string.offset)
+                if structure:
+                    string.structure = structure.name
+                    break
 
 
 class SectionLayout(Layout):
@@ -291,6 +295,9 @@ class PELayout(Layout):
         super().tag_strings(taggers)
 
     def mark_structures(self, structures=(), **kwargs):
+        # apply the PE structures to this node's own strings too (e.g. strings
+        # in the header gap that are attached to the root layout node)
+        self._mark_string_structures((structures or ()) + (self.structures_by_address,))
         for child in self.children:
             if isinstance(child, (SectionLayout, SegmentLayout)):
                 # expected child of a PE
@@ -332,6 +339,9 @@ class ELFLayout(Layout):
         super().tag_strings(taggers)
 
     def mark_structures(self, structures: Optional[Tuple[Dict[int, Structure], ...]] = (), **kwargs):
+        # apply the ELF structures to this node's own strings too (the ELF
+        # header/program-header gap strings attach to the root layout node)
+        self._mark_string_structures((structures or ()) + (self.structures_by_address,))
         for child in self.children:
             if isinstance(child, (SectionLayout, SegmentLayout)):
                 child.mark_structures(structures=(structures or ()) + (self.structures_by_address,), **kwargs)
--- a/floss/main.py
+++ b/floss/main.py
@@ -152,6 +152,13 @@ def main(argv=None) -> int:
             # the summary's layout-derived sections cover static strings only, so
             # a non-static string-type selection is a shadow arg that does nothing.
             # reject it instead of silently ignoring it.
+            if args.analyze_functions:
+                # the summary is static-only, --analyze-functions cannot show
+                # static, so the combination leaves nothing to analyze
+                parser.error(
+                    "--summary only covers static strings, which --analyze-functions does not show; "
+                    "these flags cannot be combined"
+                )
             for flag, values in (
                 ("--string-type", args.enabled_string_types),
                 ("--no-string-type", args.disabled_string_types),
--- a/floss/main.py
+++ b/floss/main.py
@@ -211,13 +211,10 @@ def main(argv=None) -> int:
     disabled_string_types = expand_string_types(list(args.disabled_string_types or []))
     enabled_string_types = expand_string_types(list(args.enabled_string_types or []))
 
-    if not enabled_string_types and not disabled_string_types and not args.analyze_functions:
-        # string deobfuscation is opt-in: default to static (and language)
-        # strings only, so a plain run doesn't spin up the slow vivisect stage
-        # for stack/tight/decoded. enable it explicitly with
-        # --string-type stack/tight/decoded (--analyze-functions requests
-        # function-level deobfuscation and is therefore not affected).
-        logger.info("string deobfuscation is off by default; use --string-type stack/tight/decoded to enable")
+    if args.summary and not disabled_string_types and not enabled_string_types:
+        # the summary's layout-derived sections cover static strings only, so
+        # don't spin up the slow deobfuscation for stack/tight/decoded
+        logger.info("--summary is static-only; skipping stack/tight/decoded extraction")
         disabled_string_types.extend([StringType.STACK.value, StringType.TIGHT.value, StringType.DECODED.value])
 
     if args.analyze_functions:
--- a/floss/main.py
+++ b/floss/main.py
@@ -211,10 +211,13 @@ def main(argv=None) -> int:
     disabled_string_types = expand_string_types(list(args.disabled_string_types or []))
     enabled_string_types = expand_string_types(list(args.enabled_string_types or []))
 
-    if args.summary and not disabled_string_types and not enabled_string_types:
-        # the summary's layout-derived sections cover static strings only, so
-        # don't spin up the slow deobfuscation for stack/tight/decoded
-        logger.info("--summary is static-only; skipping stack/tight/decoded extraction")
+    if not enabled_string_types and not disabled_string_types and not args.analyze_functions:
+        # string deobfuscation is opt-in: default to static (and language)
+        # strings only, so a plain run doesn't spin up the slow vivisect stage
+        # for stack/tight/decoded. enable it explicitly with
+        # --string-type stack/tight/decoded (--analyze-functions requests
+        # function-level deobfuscation and is therefore not affected).
+        logger.info("string deobfuscation is off by default; use --string-type stack/tight/decoded to enable")
         disabled_string_types.extend([StringType.STACK.value, StringType.TIGHT.value, StringType.DECODED.value])
 
     if args.analyze_functions:
@@ -285,7 +288,6 @@ def main(argv=None) -> int:
         large_file=args.large_file,
         quiet=args.quiet,
         verbose=args.verbose,
-        no_prompt=args.no_prompt,
         cache_dir=cache_dir,
     )
 
--- a/floss/main.py
+++ b/floss/main.py
@@ -149,21 +149,15 @@ def main(argv=None) -> int:
             if len(values) > 1 and StringType.ALL.value in values:
                 parser.error("%s: 'all' cannot be combined with other string types" % flag)
         if args.summary:
-            # the summary's layout-derived sections cover static strings only, so
-            # a non-static string-type selection is a shadow arg that does nothing.
-            # reject it instead of silently ignoring it.
             if args.analyze_functions:
                 parser.error(
                     "--summary only covers static strings, which --analyze-functions does not show; "
                     "these flags cannot be combined"
                 )
-            for flag, values in (
-                ("--string-type", args.enabled_string_types),
-                ("--no-string-type", args.disabled_string_types),
-            ):
-                for value in values:
-                    if value != StringType.STATIC.value:
-                        parser.error("%s: '--summary' only covers static strings and does not take %s" % (flag, value))
+            if args.enabled_string_types or args.disabled_string_types:
+                # --summary is its own static-only view; reject any string-type
+                # selection rather than accepting shadow args
+                parser.error("--summary only covers static strings and does not take --string-type/--no-string-type")
         if args.max_strings is not None and args.max_strings <= 0:
             parser.error("--max-strings must be a positive integer")
         for pattern in args.queries:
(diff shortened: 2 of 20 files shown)


floss/layout/base.py
```diff
@@ -291,6 +295,9 @@ def check_is_code_tagger(s: ExtractedString) -> Sequence[Tag]:
         super().tag_strings(taggers)
 
     def mark_structures(self, structures=(), **kwargs):
+        # apply the PE structures to this node's own strings too (e.g. strings
+        # in the header gap that are attached to the root layout node)
+        self._mark_string_structures((structures or ()) + (self.structures_by_address,))
```
mr-tz, line 300: Do we have an example or test for this?
I'm not sure we actually need the association here.

floss/main.py
```diff
@@ -152,6 +152,13 @@ def main(argv=None) -> int:
             # the summary's layout-derived sections cover static strings only, so
             # a non-static string-type selection is a shadow arg that does nothing.
             # reject it instead of silently ignoring it.
+            if args.analyze_functions:
+                # the summary is static-only, --analyze-functions cannot show
```
mr-tz: Comment dup of output. Remove.

tests/test_cli_args.py
```diff
@@ -191,3 +191,17 @@ def all_tags(layout):
     for s in results.strings.static_strings:
         for db_tag in ("#common", "#winapi", "#capa", "#msvc", "#openssl"):
             assert db_tag not in s.tags
+
+
+def test_yes_skips_deobfuscation_for_go(capsys):
+    """-y/--yes on a Go sample defaults to not running deobfuscation."""
```
mr-tz: Is that the right interpretation?
I thought yes means go ahead but it's already confusing so we should tweak.

vee1e: yeah should have a better name than yes considering its purpose. Changing

floss/layout/base.py
vee1e: Tests added in tests/test_layout.py:

 - `test_pe_root_strings_get_structure_annotations`
- `test_elf_root_strings_get_structure_annotations`.

Rationale for the association: ELF header/program-header/section-header gap strings are attached to the root layout node (elf.py doesn't tile a header child the way pe.py does), so without this `--structure elf-header` can never match them even though the structure was recorded at that offset. PE root strings (header gap) are the same case.

floss/main.py
vee1e: Removed in eedabce.

tests/test_cli_args.py
vee1e: Renamed to `--no-prompt` (skip the prompt, default to not running) in 8c2451c; dest/Options field are now `no_prompt`.

floss/layout/base.py
mr-tz: can you create a test sample and show the before and after?

tests/test_cli_args.py
mr-tz: hm, the implicit not running is still confusing
I'll need to look at more of the history on this issue with tty vs. auto runs and how that relates to the arg here.

vee1e: yeah there was an issue on this behaviour too #1176

floss/layout/base.py
vee1e: verified on a real sample,`tests/data/elf/687e79cd...929.elf`.

the root-attached string `@8\t@` at 0x34 (inside the ELF header region) previously had empty `structure`, so `--structure elf-header` could never match it even though the structure was recorded at that offset. with this change it resolves to `structure='elf header'`.

caveat: this only happens on binaries with header-gap strings attached to the root node (the other ELF sample in the corpus has none), which is the case #1375 flagged. should we keep it, or is that case marginal enough to drop?

mr-tz: can you show the output screenshot diffs maybe? I'm not sure I follow

tests/test_cache.py
```diff
@@ -199,7 +201,7 @@ def test_cache_file_path(tmp_path):
 
 def test_store_and_load_roundtrip(tmp_path):
     doc = make_doc()
-    key = floss.cache.compute_key(doc.metadata.sha256, __version__)
+    key = floss.cache.compute_key(doc.metadata.sha256, __version__, "auto")
```
mr-tz: make `auto` be the default on this function?

floss/pipeline.py
```diff
@@ -393,25 +393,34 @@ def analyze(options: Options) -> Optional[ResultDocument]:
...
-                try:
-                    prompt = input("Do you want to enable string deobfuscation? (this could take a long time) [y/N] ")
-                except KeyboardInterrupt:
-                    raise PipelineError("aborted by user", exit_code=130)
-                except EOFError:
-                    raise PipelineError("aborted by user", exit_code=1)
-            else:
-                prompt = "n"
+        if not enabled_string_types and not disabled_string_types:
+            if not options.no_prompt:
```
mr-tz: can we remove this and rely on --string-type arguments to drive this behavior exclusively?
we can provide a warning that certain combinations will take a long time but leave it to the user to handle that using the selected or deselected string types

vee1e: sure, so i'll remove `--no-prompt` and the prompt input block entirely and make all deobf switches off by default, so the user has to manually enable it

mr-tz: the default is all so all would be on?

mr-tz: I think that's the issue here with some implicit/explicit selection and what should be opt-in vs. opt-out.

String deobfuscation IMO for Rust/Go doesn't make much sense so we should discourage/default to not do it like FLOSS master tries to do it but improve the UX around it.

floss/layout/base.py
vee1e: Actual CLI output diff on the real sample `tests/data/elf/687e79cd...929.elf` (cache disabled, `--string-type static --structure elf-header`):

**Before** (no fix):
```
(no strings matched — the header-gap string has empty structure, so the filter excludes it)
```

**After** (with fix):
```
───────────────────────────────────── elf ─────────────────────────────────────┐
@8    @                                                                00000034│
───────────────────────────────────────────────────────────────────────────────┘
```

The root-attached string `@8\\t@` at 0x34 (ELF header region) previously had `structure=''`, so `--structure elf-header` matched nothing; now it resolves to `elf header` and the filter finds it. (Note: an earlier confusion — running the old code returned the new document because of the analysis cache; with `FLOSS_CACHE_ENABLE=0` the true before/after is as above.)

floss/pipeline.py
vee1e: default should be changed then to static only imho

floss/layout/base.py
mr-tz: I see, what's the output without --structure set for the start?


also: aren't we escaping `\t` - separate bug?

also 2: is it worth adding the cache gotcha to AGENTS.md?

floss/pipeline.py
mr-tz: FLOSS tries to give you ALL the strings but I agree with added scope (language strings, more challenging binaries) that's not ideal and why we had the prompt there...

I'm still leaning towards enabling all for now with the proper warning and let's open an issue to further tackle this later?

tests/test_cache.py
vee1e: Done — `compute_key(sha256, version, format='auto')` in f700666, so the default variant doesn't need to pass it.

floss/pipeline.py
vee1e: Done in 0abe0c2. Removed the prompt block and `--no-prompt` entirely; deobfuscation is now opt-in:

- A plain run defaults to static + language strings only (no vivisect).
- Enable stack/tight/decoded explicitly via `--string-type stack tight decoded`.
- `--analyze-functions` keeps deobfuscation on (it requests function-level analysis), so it isn't hit by the new default.
- The existing Go/Rust 'may be inaccurate and take a long time' warnings still fire when those types are on.

This also resolves #1176's silent redirect behavior — there's no implicit 'n' anymore, just an explicit default.

tests/test_cli_args.py
vee1e: Resolved as part of 0abe0c2 — the whole prompt (and this flag) is gone. Deobfuscation is now opt-in: plain runs default to static + language only, and stack/tight/decoded are enabled via `--string-type`. No implicit 'not running' anywhere — it's just the default, and `--string-type stack tight decoded` turns it on. No TTY-vs-redirect distinction remains, so #1176 is addressed too.

floss/pipeline.py
mr-tz: I think that would be a drastic usage change... @williballenthin thoughts on this specific question?

1. default floss runs produces all strings (for Go/Rust we know it doesn't make sense but we run anyway but warn the user)
2. default floss runs produce only static strings, opt-in for any deobfuscation
3. another idea...

vee1e: I'll put back the all functionality and track an issue for this since it requires more debate

mr-tz: let's discuss this first before implementing next time

vee1e: Yeah I asked the agent only to review the comments but it ran amok and pushed them without telling me :\

vee1e: #1379

floss/main.py
mr-tz: logic is off with no string type?

also just simplify to not allow --string-type/--no-string-type with summary for now?

floss/pipeline.py
```diff
@@ -390,32 +387,6 @@ def analyze(options: Options) -> Optional[ResultDocument]:
         analysis.enable_tight_strings = False
         analysis.enable_decoded_strings = False
 
-    enabled_string_types = options.enabled_string_types or []
```
mr-tz, line 393: where do we warn now?

tests/test_cache.py
mr-tz: remove auto in tests then

floss/pipeline.py
vee1e: https://github.com/vee1e/flare-floss/blob/0abe0c2103af2d5dca5d2bfa96355f9134c9cca5/floss/pipeline.py#L368-L377

it warns right after language identification

mr-tz: great, thanks

scripts/tags/extract_strings.py
```diff
@@ -263,7 +263,7 @@ def main():
     if max_len == -1:
         if args.libs:
             max_len = MAX_LEN_PES
-        elif args.libs:
+        elif args.pes:
             max_len = MAX_LEN_LIBS
```
mr-tz, line 265: is the order off here? libs -> pes?

tests/test_cli_args.py
```diff
@@ -65,14 +65,14 @@ def test_args_analysis_type_conflict(exefile):
 def test_language_extraction_independent_of_static(capsys):
     """language strings are extracted even when static strings are disabled.
 
-    uses --string-type language (so only language extraction runs) and -y so the
-    deobfuscation prompt is skipped, on a Go sample whose language is detectable.
+    uses --string-type language (so only language extraction runs) on a Go sample
+    whose language is detectable, on a Go sample whose language is detectable.
```
mr-tz: dup info

mr-tz: are these tests still useful?

tests/test_cli_args.py
```diff
@@ -191,3 +189,17 @@ def all_tags(layout):
     for s in results.strings.static_strings:
         for db_tag in ("#common", "#winapi", "#capa", "#msvc", "#openssl"):
             assert db_tag not in s.tags
+
+
+def test_deobfuscation_on_by_default(capsys):
```
mr-tz: not useful IMO

tests/test_cli_args.py
mr-tz: https://github.com/mandiant/flare-floss/issues/1379

floss/pipeline.py
mr-tz: deferring to issue / later discussion

tests/test_cache.py
vee1e: done https://github.com/mandiant/flare-floss/pull/1376/commits/3569962ff39f8dcaa1b51300028f9136cc82e3c9

floss/layout/base.py
vee1e: Full layout output (no --structure) shows the string under the root section:

```
───────────────────────────────────── elf ─────────────────────────────────────┐
@8    @                                                                00000034│
─────────────────────────────────── .interp ──────────────────────────────────┐│
...
```

also `\t` is made into spaces intentionally, not a bug: the layout module `json.dumps()` the string (escaping the tab to `\t`) then deliberately widens `\t` to 4 spaces so the row stays one line. (Edge case: a literal backslash+t in the source also matches that replace, but that's rare.)

Cache gotcha documented in AGENTS.md in https://github.com/mandiant/flare-floss/pull/1376/commits/e94f0a1111876d6676aaac6edcafd267dd9062f1

floss/main.py
vee1e: done in https://github.com/mandiant/flare-floss/pull/1376/commits/864e689552a9e5db9fcf57ca6bb2ff23d4c988cf now summary rejects any string type arg

scripts/tags/extract_strings.py
vee1e: yeah, the MAX_LEN_LIBS was applied to pes and vice versa fixed https://github.com/mandiant/flare-floss/pull/1376/commits/1413961f624ccf3a0fa2da64cc16be88efdf237e

tests/test_cli_args.py
vee1e: agreed, it only reasserted the defaults, removed https://github.com/mandiant/flare-floss/pull/1376/commits/461113d04fc37366649946d6b882c78b8a0215d2

tests/test_cli_args.py
vee1e: fixed dup in https://github.com/mandiant/flare-floss/pull/1376/commits/7bdd3fe9c4847cddee37f80e0a4f24f76618a46c.


the tests cover distinct cases and thus still useful IMO
- language extraction with static off
- manual override vs unknown auto-detect
- manual override vs wrong auto-detect

floss/layout/base.py
mr-tz: I think we want to escape all special characters in output \t \n \r etc.

mr-tz: Thanks for the other stuff, that looks good. Let's get escaping right then good to go.

vee1e: done https://github.com/mandiant/flare-floss/pull/1376/commits/f87a49197eb91f430dbfcbf42f19b29272622b0e

mr-tz (approved):

PR #1395, opened 2026-08-20
Serve the web viewer from the standalone FLOSS binary

### LINKED ISSUE #1371: Serve the web viewer from the standalone FLOSS binary
The standalone FLOSS binary cannot open the web viewer. Today you have to build the viewer yourself and upload a results file by hand.

Add a server flag and a port flag. After analysis, FLOSS should start a small local web server, serve the bundled viewer, and hand it the results so you do not need to upload a file. If you pass the server flag with no sample, the viewer still starts and you can upload a results file as usual.

Example: floss --server --port 8080 suspicious.exe

### How it should work

The viewer build is already one self-contained HTML file (scripts and styles inlined, about 320 kB). Pack that single file into the PyInstaller binary and serve it with the Python standard library. No new dependencies.

The server should:

- Serve the viewer at the root URL
- Expose the analysis results as JSON so the viewer can load them on its own
- Listen on 127.0.0.1 by default (local only; document that this is not meant to be exposed)
- Default to port 8080; the port flag chooses another (0–65535)
- Print the URL and stop cleanly on Ctrl+C
- If the viewer was not bundled, return a clear message instead of a blank page

The sample path should be optional when the server flag is set, so you can start the viewer alone.

The viewer should try to fetch results from the local server when it loads. If that fails (static file, or the local dev server), keep today's preview and upload behavior.

CI should build the viewer first and pass the built file into the binary packaging step.

### Tests

- Server with no results: viewer page is 200 if the viewer was built, otherwise a clear 404; results endpoint is 404
- Server with a minimal result document: results endpoint is 200 and the JSON loads in the viewer
- Unknown paths return 404

### Out of scope

- Shipping the viewer in a pip install. From a source checkout it should work if the viewer has been built. From a bare pip install, the results API can still work but the page itself will be missing unless we later copy the built viewer into the package.
- Auto-opening a browser. Easy to add later.
- When the server flag is set, it wins over text and JSON render flags.

cc: @mr-tz

vee1e:
Closes #1371

Adds `--server [PORT]` so the standalone FLOSS binary serves the bundled web viewer and hands it the analysis results over HTTP, instead of requiring a manual results-file upload.


doc/usage.md
```diff
@@ -82,6 +82,26 @@ content, so you can explore results without re-running the analysis.
     floss.exe malware_floss_results.json
 
 
+### View results in the web viewer (`--server`)
```
mr-tz, line 88: mention that this is the same as http://mandiant.github.io/flare-floss/ but local

floss/main.py
```diff
@@ -160,6 +178,27 @@ def main(argv=None) -> int:
                 parser.error("--summary only covers static strings and does not take --string-type/--no-string-type")
         if args.max_strings is not None and args.max_strings <= 0:
             parser.error("--max-strings must be a positive integer")
+        if args.server is not None:
+            value = args.server
+            args.server = True
+            if value != "":
+                if args.sample is None and _server_value_is_sample(value):
```
mr-tz, line 187: is this all to handle usage where `--` has not been provided to diff the sample from other args? should we handle it differently here vs. in the other cases?

floss/main.py
```diff
@@ -199,109 +238,112 @@ def main(argv=None) -> int:
 
         args.signatures = sigs_path
 
-    sample = Path(args.sample.name)
```
mr-tz, line 202: weird diff here :(
not your fault

mr-tz (commented):
let me think a bit more on the serving this page vs. just opening a local html file

PR #1399, opened 2026-08-21
feat(viewer): add download button and getting started page for standalone offline viewer

### LINKED ISSUE #1397: add option to download standalone HTML page to viewer
So people can obtain a local offline version easily

### LINKED ISSUE #1400: web viewer: add a getting-started landing page like capa Explorer Web
capa Explorer Web shows a real landing page when no results are loaded: logo, one-line description, quick-start steps, and a pointer to its offline download button: https://mandiant.github.io/capa/

The FLOSS viewer's empty state currently only says "Drag a JSON file or use the upload button" (`viewer/src/App.tsx`).

A matching landing page would help first-time users:

1. Install floss, e.g. `pip install flare-floss`
2. Analyze a sample and save the JSON results: `floss -j /path/to/sample > results.json`
3. Load `results.json` into the viewer (drag and drop or Upload)

It should also mention the download button (#1399), so people see they can grab a standalone copy of the viewer for offline use, same as capa does.

vee1e:
Adds a download button to the viewer that saves the page as a single self-contained HTML file (`floss-viewer-<version>-<commit>.html`) people can keep and open without the hosted site.

Live at: https://flare-floss-viewer.vercel.app/

What changes:

- Getting-started landing page with Upload / Demo / Download viewer actions.
- The sidebar stays hidden until a results document is loaded; a persistent Upload button and the theme toggle are pinned top-right.
- Download: on the hosted site it fetches the page source and saves it; if fetch is unavailable (a local copy opened via `file://`), it serializes the current DOM instead. The filename embeds the FLOSS version and git commit, derived automatically at build time.
- Fully self-contained offline: JS/CSS plus the logo and favicon are inlined as base64 data URIs; fonts are the system monospace stack, so there are no external requests.
- README updated.

Fixes #1397
Fixes #1400

## Screenshots

Landing page (dark):

Landing page, dark mode

Landing page (light):

Landing page, light mode

Loaded results document (dark):

Loaded results document, dark mode

Loaded results document (light):

Loaded results document, light mode


viewer/src/App.tsx
```diff
@@ -524,6 +524,33 @@ const App: React.FC = () => {
...
+      } catch {
+        // fetch can fail on file:// or restricted origins; fall back to serializing the live DOM
+      }
+    }
+    if (!html) {
+      const clone = document.documentElement.cloneNode(true) as HTMLElement;
+      const root = clone.querySelector('#root');
+      if (root) root.innerHTML = '';
+      html = '<!doctype html>\n' + clone.outerHTML;
+    }
```
mr-tz, line 576: does the live DOM saving result in the same file as other download (same SHA256s)?

vee1e: no, they produce different files. test results:

| download | sha256 |
|---|---|
| served `dist/index.html` | `d093144c…` |
| over http (fetch path) | `d093144c…`, same as served |
| via file:// (DOM path) | `40c0bc58…` |
| file:// again, nothing changed | `40c0bc58…` |
| file:// after toggling the theme | `074808a2…` |

the fetch path saves the exact served bytes, the DOM path can't, for two reasons:

1. Runtime state gets baked into the clone. React sets `data-theme` on `<html>` (App.tsx:316), so the saved copy carries whatever theme was active. Toggling the theme alone changes the hash.
2. Serializing the DOM (turning the live tree back into HTML text) is not byte-preserving: `crossorigin` becomes `crossorigin=""`, trailing newlines collapse.

Same browser plus same state reproduces the same file, but the hash depends on who downloaded it and when. If you want one canonical file per release, I can inline the page source at build time instead (a small Vite plugin that embeds `dist/index.html` as a string), which makes every download byte-identical. WDYT?

mr-tz: yeah, ideally we can provide a consistent download and name it based on a version string and/or commit hash

or do you have other thoughts?

vee1e: yeah something like `viewer-<version>.html` could work

mr-tz: and ideally that gets bumped automatically somehow because we always forget

viewer/scripts/generate-version.mjs
```diff
@@ -0,0 +1,24 @@
...
+
+let commit = 'unknown';
+try {
+  commit = execSync('git rev-parse --short HEAD', { cwd: root }).toString().trim();
+} catch {
+  // not a git checkout (e.g. release tarball); fall back to 'unknown'
+}
+
+const out = `// Generated by scripts/generate-version.mjs — do not edit.
+export const VIEWER_VERSION = '${pkg.version}';
```
mr-tz: does this need to be set somehow manullay or what value will be used?

vee1e: no, when npm build runs, the script executes `git rev-parse --short HEAD` in the repo, so the filename always gets the commit you're building from.

The only case it can't do that is building outside a git checkout (e.g. from a release tarball), then it falls back to unknown.

the version part comes from the `version` field in package.json so that only changes if/when we decide to bump it.

mr-tz: the version is what I meant, we should look into automation to help us do/remember to bump

viewer/src/App.tsx
mr-tz: TODO?

viewer/scripts/generate-version.mjs
vee1e: what's the oracle or cutoff for a new version of this to be released? can that be codified in some way?

viewer/src/App.tsx
vee1e: yeah unless we have a definite threshold to version numbers

mr-tz: how do you mean?

viewer/scripts/generate-version.mjs
mr-tz: at least highlight that something needs to be bumped - maybe in sync with FLOSS or better independent?

vee1e: yeah I think floss releases can be a good solution for now, easier to follow just one version spec. wdyt? We can have a script to extract the ver. and hardcode it during the build step.

viewer/src/App.tsx
vee1e: more context in the later comment but basically what constitutes a major release, etc.

I feel making it the same version as floss is the best option

viewer/scripts/generate-version.mjs
mr-tz: yeah, for sure but floss and the viewer could advance independently so that coupling then could be confusing/a hindrance

vee1e: Also true, but are there any alternatives you would suggest then for versioning?

mr-tz: Let's start with floss version and git hash so it's automatic.

vee1e: sure

viewer/index.html
```diff
@@ -5,10 +5,7 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <title>FLOSS Graphical Viewer</title>
-  <link rel="icon" type="image/png" href="favicon.png" />
-  <link rel="preconnect" href="https://fonts.googleapis.com">
-  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-  <link href="https://fonts.googleapis.com/css2?family=Geist+Mono:wght@100..900&display=swap" rel="stylesheet">
+  <link rel="icon" type="image/png" href="/favicon.png" />
```
mr-tz: Base commit `d713054` (already has #1398 merged) has `href="favicon.png"` here (relative). This PR's diff flips it back to `href="/favicon.png"` — reintroducing the absolute path that #1398 just fixed. The PR description says 'This branch leaves those lines untouched, so the two merge in any order,' but that doesn't match this hunk. Absolute paths break under a non-root base path and, more importantly, break when the downloaded standalone HTML is opened via `file://` (resolves against the filesystem root, not the app's directory).

viewer/src/App.tsx
```diff
@@ -617,44 +673,15 @@ const App: React.FC = () => {
...
-            </button>
-            <button className="btn-ghost" onClick={handlePreview}>Preview</button>
-            <label htmlFor="file-upload" className="btn-ghost" style={{ cursor: 'pointer' }}>
-              Upload
-            </label>
-            <input {...getInputProps()} id="file-upload" />
-          </div>
-        </div>
-
+      {data && (
```
mr-tz, line 689: Once a results file is loaded, there's no more way to load a different one except drag-and-drop. The old sidebar-header (Upload button + Preview button) was removed and replaced by controls that only exist in the `!data` landing branch further down. The dropzone is configured with `noClick: true, noKeyboard: true` (line 439-440), so clicking the app doesn't open a file picker either. On top of that, this PR adds mobile CSS breakpoints, but drag-and-drop from a file manager is impractical on touch devices — so mobile users are effectively stuck once a file is loaded. Worth keeping a persistent Upload control (e.g. next to the floating theme toggle) that survives `data` being set.

viewer/src/App.tsx
```diff
@@ -524,6 +552,34 @@ const App: React.FC = () => {
...
+      const root = clone.querySelector('#root');
+      if (root) root.innerHTML = '';
+      clone.removeAttribute('data-theme');
+      html = '<!doctype html>\n' + clone.outerHTML;
+    }
+    const blob = new Blob([html], { type: 'text/html' });
+    const url = URL.createObjectURL(blob);
+    const anchor = document.createElement('a');
+    anchor.href = url;
+    anchor.download = `floss-viewer-${VIEWER_VERSION}-${VIEWER_COMMIT}.html`;
```
mr-tz: Even setting aside the absolute-path regression noted above, the 'fully self-contained' claim doesn't fully hold: `vite-plugin-singlefile` only inlines the JS/CSS/asset module graph it builds — it doesn't touch `floss-logo.png` or `favicon.png` in `viewer/public/`, which are referenced as plain URL strings rather than imported. So those two images were never fetched/embedded into the single file, and will fail to load once the downloaded HTML is opened offline via `file://`. Worth either inlining them as base64 data URIs (they're small) or dropping the offline claim's scope to match reality.

viewer/index.html
vee1e: Fixed: the href is now relative (favicon.png) again, matching #1398. Additionally it gets inlined to a base64 data URI at build time (tiny vite plugin in vite.config.ts), so the standalone download stays fully self-contained when opened via file://.

viewer/src/App.tsx
vee1e: Fixed: there's now a persistent Upload button pinned top-right next to the theme toggle that survives data being set (the hidden file input is hoisted to the App level and both the landing Upload and the topbar Upload label point at it). It now uses the same opaque surface background as the theme toggle.

Dark mode:

Persistent Upload control next to the theme toggle, dark mode

Light mode:

Persistent Upload control next to the theme toggle, light mode

viewer/src/App.tsx
vee1e: Fixed: both images are now inlined as base64 data URIs at build time — floss-logo.png is moved from public/ into src/ and imported, so vite-plugin-singlefile's asset inlining kicks in; favicon.png is inlined by a small build-time plugin in vite.config.ts. Verified with a server that 404s all image requests: the page renders the logo (naturalWidth 2124) with zero external PNG fetches, and the downloaded standalone HTML contains both data URIs.

viewer/src/App.tsx
mr-tz: TODO consistent download

viewer/vite.config.ts
```diff
@@ -1,8 +1,24 @@
...
+// Inline the favicon as a data URI so the downloaded standalone HTML
+// stays fully self-contained when opened offline via file://.
+const inlineFavicon = (): Plugin => ({
+  name: 'inline-favicon',
+  apply: 'build',
+  transformIndexHtml(html) {
+    const dataUri = `data:image/png;base64,${readFileSync(join(publicDir, 'favicon.png')).toString('base64')}`
+    return html.replace('href="favicon.png"', `href="${dataUri}"`)
+  },
+})
```
mr-tz: do we need a plugin for this vs. just embedding directly?

vee1e: Nope, the favicon is now embedded directly as a static data URI in index.html (it changes rarely, so hardcoding it is fine), which also removes the @types/node dependency the plugin needed.

viewer/src/App.tsx
vee1e: Added a TODO at handleDownloadViewer documenting that the file:// fallback (DOM serialization) bakes in runtime state and is not byte-identical to the fetch path, with the build-time embedding idea as the eventual fix.

mr-tz: ok, fair for now

viewer/src/App.tsx
```diff
@@ -524,6 +553,38 @@ const App: React.FC = () => {
...
```
mr-tz: last thought, how about to include date of the commit for easier human understanding like 26-08-24 (yy-mm-dd)?

viewer/.gitignore
```diff
@@ -23,3 +23,6 @@ dist-ssr
 *.sln
 *.sw?
 .vercel
+
+# generated at build time
+src/generated
```
mr-tz, line 28: how is this used?

viewer/README.md
```diff
@@ -8,6 +8,7 @@ This is a web-based viewer for analyzing the output of the `floss` tool. It allo
 - Filter strings by search term, minimum length, tags, and structures.
 - Toggle display of columns (tags, encoding, offset/structure).
 - Copy filtered strings to the clipboard.
+- Download the viewer itself as a standalone HTML file for offline use (download icon in the header). The saved file is fully self-contained and can be opened directly in a browser without an internet connection. The file is named after the viewer version and git commit it was built from (e.g. `floss-viewer-0.0.0-abc1234.html`), so downloads are identifiable and reproducible per build.
```
mr-tz: update version string if finalized if needed

vee1e: Done both https://github.com/mandiant/flare-floss/commit/ec828bcabe48c940e690414bc2290f763e2169bf

viewer/src/App.tsx
vee1e: Yeah sounds good

viewer/.gitignore
vee1e: This ignores the `src/generated/` directory, which holds a single build-time-generated file: `version.ts`, exporting the FLOSS version, commit hash, and commit date that are stamped into the download filename. It is written by `scripts/generate-version.mjs`, which runs before every `npm run dev` / `npm run build` (it reads `floss/version.py` and runs `git rev-parse --short HEAD`). It is gitignored because its content changes on every commit (the hash is part of it), so committing it would just add churn to every PR.

mr-tz: thx

viewer/README.md
```diff
@@ -8,6 +8,7 @@ This is a web-based viewer for analyzing the output of the `floss` tool. It allo
 - Filter strings by search term, minimum length, tags, and structures.
 - Toggle display of columns (tags, encoding, offset/structure).
 - Copy filtered strings to the clipboard.
+- Download the viewer itself as a standalone HTML file for offline use (download icon in the header). The saved file is fully self-contained and can be opened directly in a browser without an internet connection. The file is named after the viewer version, the commit date, and the git commit it was built from (e.g. `floss-viewer-3.1.1-26-08-24-abc1234.html`), so downloads are identifiable, reproducible, and sortable per build.
```
mr-tz:
```suggestion
- Download the viewer itself as a standalone HTML file for offline use (download icon in the header). The saved file is fully self-contained and can be opened directly in a browser without an internet connection. The file is named after the viewer version, the commit date, and the git commit it was built from (e.g. `floss-viewer-3.1.1-26-08-24-abc1234.html`).
```

mr-tz (approved):

viewer/README.md
```diff
@@ -25,6 +26,8 @@ npm run dev
 
 This will start a local server, and you can view the application in your browser. The server supports Hot Module Replacement (HMR), so changes to the source code will be reflected live without a full page reload.
 
+TODO: add basic viewer tests (e.g. a CI job that runs `npm run lint` and `npm run build`, plus a small smoke test for load/upload/download flows), see the discussion in mandiant/flare-floss#1399.
+
```
mr-tz: remove here, add issue for this instead

vee1e: https://github.com/mandiant/flare-floss/issues/1402

PR #1407, opened 2026-08-25
Add shell completion scripts (--print-completion)

### LINKED ISSUE #1350: QS/FLOSS: Add completion scripts
Add zsh/bash/fish completion scripts for FLOSS args so that you don't have to run `--help` every time, just a tab will work.

vee1e:
Fixes #1350

Adds tab completion for FLOSS arguments via [`shtab`](https://github.com/tqdm/shtab), which generates the scripts directly from our `argparse` parser. Because they're derived from the argument definitions, flags, choice values, and file paths stay in sync automatically as the CLI evolves.

    floss --print-completion {bash,zsh,tcsh,fish,powershell}

Sample output:

```
$ floss -
--analyze-functions      -- only analyze the specified functions, hex-encoded like 0x401000, space-separate multiple functions
--color                  -- enable ANSI color codes in results, default: only during interactive session
--columns                -- columns to show in the layout view; valid values: tags, offset, structure, encoding. Default: tags, offset.
--debug              -d  -- enable debugging output on STDERR, specify multiple times to increase verbosity
--format             -f  -- select sample format, auto: (default) detect file type automatically, pe: Windows PE file, sc32: 32-bit shellcode, sc64: 64-bit shellcode
--help               -h  -- show this help message and exit
--interesting            -- exclude strings with noisy tags: #code, #code-junk, #common, #duplicate, #reloc
--json               -j  -- emit JSON instead of text
--language               -- use language-specific string extraction, auto-detect language by default, disable using 'none'
--large-file         -L  -- allow processing files larger than 64 MB
--max-strings            -- cap the emitted strings per section to the top N by relevance (highlighted, then untagged, then tagged, ascending by offset)
--minimum-length     -n  -- minimum string length
--no-section             -- do not show static strings in the given binary section(s)
--no-string-type         -- do not extract specified string type(s); valid values: static, stack, tight, decoded, language, all
--no-structure           -- do not show static strings in the given binary structure(s)
--no-tag                 -- do not show strings with the given semantic tag(s)
--plain                  -- render the classic flat list of strings without layout and tags
--print-completion       -- print shell completion script for the given shell
--query                  -- only show strings matching the given regular expression(s); repeatable, patterns are ORed
--quiet              -q  -- disable all status output on STDOUT except fatal errors
--section                -- only show static strings in the given binary section(s); e.g. .rdata
--signatures             -- path to .sig/.pat file or directory used to identify library functions, use embedded signatures by default
--string-type            -- only extract specified string type(s); valid values: static, stack, tight, decoded, language, all
--structure              -- only show static strings in the given binary structure(s); e.g. import-table, export-table, rich-header, section-header, elf-header, program-header, string-table, sy
--summary                -- emit a concise summary (metadata, counts, tag histogram, high-value strings); static-only by default unless string types are explicitly selected
--tag                    -- only show strings with the given semantic tag(s); e.g. winapi, crypto, or a tag family: crt, expert, gp, oss, winapi; run --summary to see the tags present in a spec
--verbose            -v  -- enable verbose results, e.g. including function offsets (does not affect JSON output)
--version                -- show program's version number and exit
```

## What completes

- all flags and options (`floss --string-<TAB>` offers `--string-type`)
- the `sample` positional completes file paths
- per-shell installation instructions added to [doc/usage.md](doc/usage.md)

## Notes

- `shtab` is a small pure-python dependency with no transitive deps; pinned in `requirements.txt` (`shtab==1.12.0`), loose bound in `pyproject.toml` (`shtab>=1.7`) per the library-dependency policy
- also works from PyInstaller standalone builds, where users can't pip-install extras
- parser now sets `prog="floss"` so generated scripts and usage text are stable across entry points (entry point, `python -m floss`, standalone exe)
- new tests in `tests/test_completion.py`: script generation + exit codes for bash/zsh/fish, content markers, and invalid-shell error path

cc: @mr-tz

### DIFF
--- a/floss/cli.py
+++ b/floss/cli.py
@@ -24,6 +24,8 @@ from enum import Enum
 from typing import List, Optional
 from pathlib import Path
 
+import shtab
+
 import floss.utils
 import floss.logging_
 from floss.const import (
@@ -151,6 +153,7 @@ def make_parser():
         """)
 
     parser = ArgumentParser(
+        prog="floss",
         description=desc,
         epilog=epilog,
         formatter_class=argparse.RawDescriptionHelpFormatter,
@@ -164,11 +167,13 @@ def make_parser():
         help="minimum string length",
     )
 
-    parser.add_argument(
+    sample_argument = parser.add_argument(
         "sample",
         type=argparse.FileType("rb"),
         help="path to sample to analyze",
     )
+    # enable file path completion for the sample positional argument
+    setattr(sample_argument, "complete", shtab.FILE)
 
     analysis_group = parser.add_argument_group("analysis arguments")
     analysis_group.add_argument(
@@ -334,6 +339,12 @@ def make_parser():
             action=floss.utils.UninstallContextMenu,
             help="uninstall FLOSS from the right-click context menu for Windows Explorer and exit",
         )
+
+    shtab.add_argument_to(
+        parser,
+        ["--print-completion"],
+        help="print shell completion script for the given shell",
+    )
 
     output_group = parser.add_argument_group("rendering arguments")
     output_group.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
--- a/floss/cli.py
+++ b/floss/cli.py
@@ -343,7 +343,7 @@ def make_parser():
     shtab.add_argument_to(
         parser,
         ["--print-completion"],
-        help="print shell completion script for the given shell",
+        help="print shell completion script for bash, zsh, tcsh, fish, or powershell",
     )
 
     output_group = parser.add_argument_group("rendering arguments")
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -1,0 +1,68 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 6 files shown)


floss/cli.py
```diff
@@ -335,6 +340,12 @@ def make_parser():
             help="uninstall FLOSS from the right-click context menu for Windows Explorer and exit",
         )
 
+    shtab.add_argument_to(
+        parser,
+        ["--print-completion"],
+        help="print shell completion script for the given shell",
+    )
```
mr-tz, line 347: can this show what shells are supported? how can users know?

mr-tz (commented):
overall I think this is very neat and strikes a balance between making this ok to setup vs. us maintaining more of the install process via a `--install-completion` option with full setup for example

this may get buried in the usage doc a bit - should we highlight it elsewhere as a nice QoL feature?

floss/cli.py
vee1e: good point done in 9f9fdec1

mr-tz (approved):

tests/test_completion.py
```diff
@@ -0,0 +1,68 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import pytest
+
+import floss.main
+
+
+@pytest.mark.parametrize("shell", ["bash", "zsh", "fish"])
```
mr-tz: what about the others mentioned in the docs? not worth testing?

vee1e: ah my bad done in aa1c5398

PR #1414, opened 2026-08-29
Simplify and dedupe core modules; fix amd64 imm32 xref decoding

vee1e:
## What

A cleanup pass across the core modules, driven by a ponytail-review-style hunt for over-engineering. The changes are deduplication, dead-code removal, and rewrites to idiomatic equivalents, plus the bugs that surfaced while doing them. No new features.

## Changes

- Remove dead symbols: `DemoHook`, the module-level `hex()` shadow, the unused `strings.py` main block, two no-op `__init__` overrides, and a redundant `LayoutFilter.relevance_key` staticmethod.
- Collapse six near-identical xref scanners (LEA/PUSH/MOV, i386/amd64) into one helper.
- Dedupe the language-string stripping logic shared by `get_extract_stats` and `get_missed_strings`.
- Dedupe the Go/Rust language extraction branches in the pipeline.
- Merge the PE/ELF xor/reloc/code taggers and the structure-marking recursion into shared `Layout` helpers.
- Replace `list(filter(lambda ...))` with comprehensions, `map(lambda ...)` with comprehensions, `type(x)` checks with `isinstance`, and `operator.getitem` with a subscript.
- Restore the disabled `PrintfHook` (kept out of the active hook set for future iteration) and the original parenthesized import layout.

## Bugs found while refactoring

- The amd64 PUSH and MOV scanners captured a 4-byte imm32 but unpacked it as a 64-bit value, so `struct.unpack` errored on every real match. Unpack as 32-bit.
- The shared xref iterator passed a base address to those amd64 scanners, which took only one argument, so push/mov decoding raised `TypeError` on amd64 PEs. Accept the base address, matching the LEA scanner.
- Swapping `util.hex` for the builtin `hex()` changed output for negative `frame_offset` values (`0x-8` became `-0x8`). Use a format string that matches the original for both signs.
- The merged structure-marking helper silently dropped `**kwargs`. Thread it through to the child recursion.
- Replacing an `assert` with explicit branching in `log_result` meant a malformed string raised `AssertionError`, which Python strips under `-O` and then surfaced as `AttributeError`. Restore an explicit `ValueError`.

### DIFF
--- a/floss/api_hooks.py
+++ b/floss/api_hooks.py
@@ -407,6 +407,18 @@ class SnprintfHook:
         return False
 
 
+class PrintfHook:
+    # TODO disabled for now as incomplete (need to implement string format) and could result in FP strings as is
+    def __call__(self, emu, api, argv):
+        # TODO vfprintf, vfwprintf, vfprintf_s, vfwprintf_s, vsnprintf, vsnwprintf, etc.
+        if fu.contains_funcname(api, ("vsprintf", "vswprintf", "wvsprintfA")):
+            buf, format_, *va_list = argv
+            format_str = fu.readStringAtRva(emu, format_, maxsize=MAX_STR_SIZE)
+            emu.writeMemory(buf, format_str)
+            fu.call_return(emu, api, argv, buf)
+            return True
+
+
 class ExitExceptionHook:
     def __call__(self, emu, api, argv):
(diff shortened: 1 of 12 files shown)


mr-tz (approved):
thanks, good fixes

PR #1418, opened 2026-09-01
Emit parseable Python for decoded strings in the Ghidra script

### LINKED ISSUE #1351: fix(ghidra): generated import script fails Python parsing
## Summary

`scripts/render-ghidra-import-script.py` emits syntactically invalid Python whenever a FLOSS result contains a non-empty decoded string. The renderer exits successfully, but the generated PyGhidra script cannot be parsed or run.

Target repository: https://github.com/mandiant/flare-floss

## Steps to reproduce

1. Check out current FLARE-FLOSS revision `98b1fa594bac442567f7b340fd1703cdf443474f` and install it in a virtual environment.
2. Save this minimal valid result document as `/tmp/floss-results.json`:

```json
{
  "metadata": {
    "file_path": "fixture.exe"
  },
  "strings": {
    "decoded_strings": [
      {
        "address": 4096,
        "address_type": "STACK",
        "string": "DECODED_TEST",
        "encoding": "ASCII",
        "decoded_at": 4199808,
        "decoding_routine": 4199654
      }
    ]
  }
}
```

3. Render and parse-check the generated script:

```sh
python scripts/render-ghidra-import-script.py /tmp/floss-results.json > /tmp/apply_floss.py
python -m py_compile /tmp/apply_floss.py
```

The first command exits successfully. The second command fails.

## Expected behavior

The renderer should emit a valid Python 3 script for a valid FLOSS result containing decoded strings. `python -m py_compile /tmp/apply_floss.py` should exit successfully, allowing the script to proceed to Ghidra-specific runtime validation.

A regression check should parse or compile the generated script rather than checking only the renderer process's exit status.

## Actual behavior

The generated script contains this shape:

```python
print(f"FLOSS: string \"{decode_string(\"REVDT0RFRF9URVNU\")}\" decoded at VA 0x401580")
```

Python rejects it before execution:

```text
SyntaxError: unexpected character after line continuation character
```

## Affected area

- `scripts/render-ghidra-import-script.py`, specifically the generated print commands for decoded strings with both global and non-global address types
- `tests/test_scripts.py`, whose current renderer test checks only the generator's return code and does not parse the generated Python
- The documented Ghidra 12.0+ PyGhidra import workflow

## Runtime or environment

- FLARE-FLOSS source revision: `98b1fa594bac442567f7b340fd1703cdf443474f`
- FLARE-FLOSS version: `3.1.1`
- Local minimized reproduction: Python 3.12
- Independent clean-container reproduction: Ubuntu 26.04, Linux AArch64, Python 3.14.4
- Ghidra is not required to reproduce the syntax failure

## Evidence

The renderer constructs the decoded-string print commands using a generated f-string whose expression contains backslash-escaped quotes. This occurs in both decoded-string branches of `render_ghidra_script()`.

The minimized result above is accepted by `ResultDocument.parse_file()`. The renderer returns exit status 0 and writes the output, after which Python's compiler rejects that output at the nested `decode_string(...)` expression.

The same failure was independently reproduced in a clean container using a benign PE fixture. FLOSS recovered the expected decoded and tight strings, generated the Ghidra import script, and that script failed `py_compile` with the same syntax error. The fixture was never executed.

A live search of the repository's open and closed issues and pull requests on 2026-08-07 found no report with the same generated-script parsing failure. Nearby items include the merged PyGhidra migration in #1190 and earlier Ghidra integration work in #323, #394, and #402; those track the integration or its design rather than this syntax failure.

## Impact

Any result document containing a non-empty decoded string produces an unusable Ghidra import script. Because the renderer itself exits successfully, automated workflows may treat generation as successful and discover the failure only when Ghidra or another Python parser attempts to load the artifact.

## Additional context

The failure is independent of the decoded string's content: the base64 payload contains no user-controlled Python syntax, but the surrounding generated f-string is already invalid. A complete closure signal is that generated output for both decoded-string address branches passes Python parsing, while existing stack-string and tight-string rendering behavior remains intact.

arpitjain099:
Closes #1351.

`render-ghidra-import-script.py` builds the decoded-string print lines as a nested f-string whose expression contains backslash-escaped quotes, so the generated script comes out as:

```python
print(f"FLOSS: string \"{decode_string(\"REVDT0RFRF9URVNU\")}\" decoded at VA 0x401580")
```

which Python will not parse. The renderer still exits 0, so the failure only shows up when Ghidra, or anything else, tries to load the script.

Reproduced on the current master with the minimal result document from the issue:

```
$ python scripts/disassemblers/render-ghidra-import-script.py results.json > apply_floss.py
$ python -m py_compile apply_floss.py
  File "apply_floss.py", line 160
    print(f"FLOSS: string \"{decode_string(\"REVDT0RFRF9URVNU\")}\" decoded at VA 0x401580")
                                            ^
SyntaxError: unexpected character after line continuation character
```

Both decoded-string branches now concatenate instead of nesting, so the generated line is `print("FLOSS: string \"" + decode_string("...") + "\" decoded at VA 0x401580")`, which parses. The output text is unchanged, and the stack-string and tight-string lines were already built by concatenation and are untouched.

`test_ghidra_script_parses` renders a document with one decoded string, once for each address type, and compiles the output rather than only checking the exit status, which is the gap the issue points at in the existing test. Both parameters fail on master with the SyntaxError above.

`pytest tests/test_disassembler_scripts.py -k ghidra_script_parses` passes, and black and isort are clean on both files with the line length from `.pre-commit-config.yaml`.

### DIFF
--- a/scripts/disassemblers/render-ghidra-import-script.py
+++ b/scripts/disassemblers/render-ghidra-import-script.py
@@ -223,12 +223,12 @@ def render_ghidra_script(result_document: ResultDocument) -> str:
             b64 = base64.b64encode(ds.string.encode("utf-8")).decode("ascii")
             if ds.address_type == AddressType.GLOBAL:
                 main_commands.append(
-                    f'    print(f"FLOSS: string \\"{{decode_string(\\"{b64}\\")}}\\" at global VA 0x{ds.address:x}")'
+                    f'    print("FLOSS: string \\"" + decode_string("{b64}") + "\\" at global VA 0x{ds.address:x}")'
                 )
                 main_commands.append(f'    append_comment(0x{ds.address:x}, "FLOSS: " + decode_string("{b64}"))')
             else:
                 main_commands.append(
-                    f'    print(f"FLOSS: string \\"{{decode_string(\\"{b64}\\")}}\\" decoded at VA 0x{ds.decoded_at:x}")'
+                    f'    print("FLOSS: string \\"" + decode_string("{b64}") + "\\" decoded at VA 0x{ds.decoded_at:x}")'
                 )
                 main_commands.append(f'    append_comment(0x{ds.decoded_at:x}, "FLOSS: " + decode_string("{b64}"))')
     main_commands.append('    print("Imported decoded strings from FLOSS")')
--- a/tests/test_disassembler_scripts.py
+++ b/tests/test_disassembler_scripts.py
@@ -20,6 +20,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 
 import sys
+import json
 import subprocess
 from pathlib import Path
 from functools import lru_cache
@@ -66,3 +67,35 @@ def test_disassembler_scripts(script, args):
     script_path = get_disassembler_script_path(script)
     p = run_program(script_path, args)
     assert p.returncode == 0
+
+
+def make_result_document(path: Path, address_type: str) -> Path:
+    """Write a minimal result document holding one decoded string."""
+    document = {
+        "metadata": {"file_path": "fixture.exe"},
+        "strings": {
+            "decoded_strings": [
+                {
+                    "address": 4096,
+                    "address_type": address_type,
+                    "string": "DECODED_TEST",
+                    "encoding": "ASCII",
+                    "decoded_at": 4199808,
+                    "decoding_routine": 4199654,
+                }
+            ]
+        },
+    }
+    path.write_text(json.dumps(document))
+    return path
+
+
+@pytest.mark.parametrize("address_type", ["GLOBAL", "STACK"])
+def test_ghidra_script_parses(tmp_path, address_type):
+    """The generated Ghidra script has to be valid Python, not just render cleanly."""
+    results = make_result_document(tmp_path / f"{address_type.lower()}.json", address_type)
+
+    p = run_program(get_disassembler_script_path("render-ghidra-import-script.py"), [str(results)])
+
+    assert p.returncode == 0
+    compile(p.stdout.decode("utf-8"), "apply_floss.py", "exec")


scripts/disassemblers/render-ghidra-import-script.py
```diff
@@ -223,12 +223,12 @@ def render_ghidra_script(result_document: ResultDocument) -> str:
             b64 = base64.b64encode(ds.string.encode("utf-8")).decode("ascii")
             if ds.address_type == AddressType.GLOBAL:
                 main_commands.append(
-                    f'    print(f"FLOSS: string \\"{{decode_string(\\"{b64}\\")}}\\" at global VA 0x{ds.address:x}")'
+                    f'    print("FLOSS: string \\"" + decode_string("{b64}") + "\\" at global VA 0x{ds.address:x}")'
```
vee1e: Good fix. The old fill-in string (a string with values filled in) put backslashes in the braces part, and Python rejects that. The new code joins plain parts, so it parses and prints the same text.

tests/test_disassembler_scripts.py
```diff
@@ -20,6 +20,7 @@
 # See the License for the specific language governing permissions and limitations under the License.
 
 import sys
+import json
```
vee1e: Please move this import up. Sort order wants json first, not after sys. Then run isort (the import sort tool) again.

tests/test_disassembler_scripts.py
```diff
@@ -66,3 +67,35 @@ def test_disassembler_scripts(script, args):
...
+
+@pytest.mark.parametrize("address_type", ["GLOBAL", "STACK"])
+def test_ghidra_script_parses(tmp_path, address_type):
+    """The generated Ghidra script has to be valid Python, not just render cleanly."""
+    results = make_result_document(tmp_path / f"{address_type.lower()}.json", address_type)
+
+    p = run_program(get_disassembler_script_path("render-ghidra-import-script.py"), [str(results)])
+
+    assert p.returncode == 0
+    compile(p.stdout.decode("utf-8"), "apply_floss.py", "exec")
```
vee1e: Good check. You now parse the made file with compile (the built in parse check), not just the exit code. Think about adding the same check for the other render scripts too.

PR #1419, opened 2026-09-01
Add --html standalone HTML reports

vee1e:
FLOSS can write a self-contained HTML report. Open it in a browser and the results are already there.

    floss --html sample.exe > report.html

This is the existing graphical viewer with the analysis baked in. No local server. `--html` and `--json` cannot be used together.

You can also turn a saved JSON results file into a report:

    floss --html results.json > report.html

## Screenshots

Taken in Chrome from Practical Malware Analysis Lab 01-01.exe (static strings only) after `floss --html`.

Loaded report. Results are already filled in, no upload screen:

Loaded HTML report

The viewer with no embedded results still shows the upload screen:

Empty viewer

## Notes

`--html` reads the built viewer from viewer/dist. From a source checkout, build it once:

    cd viewer && npm install && npm run build

The pip package and the standalone exe do not ship that template yet.

## Tests

Full pytest: 436 passed, 2 skipped, 6 xfailed.
Viewer lint and production build succeed.
`--html` plus `--json` errors.


floss/main.py
```diff
@@ -259,10 +269,16 @@ def main(argv=None) -> int:
                 logger.error("%s", e)
             return -1
 
-        if args.json:
-            r = floss.render.json.render(results)
-        else:
-            r = render_text(args, results, stream=sys.stdout)
+        try:
+            if args.json:
```
mr-tz: This does not need to be in the try block

floss/main.py
```diff
@@ -298,12 +314,19 @@ def main(argv=None) -> int:
...
         return 0
 
-    if args.json:
-        r = floss.render.json.render(analysis_results)
-    else:
-        # this may be slow when there's many strings, so informing users what's happening
-        logger.info("rendering results")
-        r = render_text(args, analysis_results, stream=sys.stdout)
+    try:
+        if args.json:
```
mr-tz: Also not needed in the try block

tests/test_main.py
```diff
@@ -31,12 +31,14 @@ def test_main_help(capsys):
         out = capsys.readouterr().out
         assert "usage:" in out
         assert "--json" in out
+        assert "--html" in out
```
mr-tz, line 34: Remove

tests/test_main.py
```diff
@@ -31,12 +31,14 @@ def test_main_help(capsys):
...
 
     # running without arguments prints the same help and exits with code 1
     assert floss.main.main([]) == 1
     out = capsys.readouterr().out
     assert "usage:" in out
     assert "--json" in out
+    assert "--html" in out
```
mr-tz: Remove

mr-tz (changes requested):
Let's also add tests to show this works for pyinstaller built standalone executables and from source (and if possible from pip installs). 
Good solution overall here.

floss/main.py
vee1e: removed in d638f6f

floss/main.py
vee1e: removed in d638f6f

tests/test_main.py
vee1e: removed in 41e5a93

tests/test_main.py
vee1e: removed in 41e5a93

doc/usage.md
```diff
@@ -199,6 +199,22 @@ The easiest way to explore this output is the hosted
 filter by tag, structure, or search term, and copy the strings you keep. See
 [results_document.md](results_document.md) for the schema itself.
 
+### Write a standalone HTML report (`--html`)
+
+Write a self-contained HTML file that opens in a browser with the results already
+loaded. `--html` cannot be combined with `-j/--json`.
```
mr-tz: worth explicit mention on --html and -j being incompatible?

doc/usage.md
```diff
@@ -199,6 +199,22 @@ The easiest way to explore this output is the hosted
...
+
+    floss.exe --html malware.exe > malware_strings.html
+
+The report is a copy of the [web viewer](https://mandiant.github.io/flare-floss/)
+with the result document embedded. Filtering happens in the browser, the same
```
mr-tz:
```suggestion
with the result document embedded. Interactive exploration of the data happens in the browser, the same
```

floss/render/templates/index.html
mr-tz, line 1: is this a dup of index.html - how does this get rebuild? can we do this without this file?

floss/main.py
```diff
@@ -261,6 +271,12 @@ def main(argv=None) -> int:
 
         if args.json:
             r = floss.render.json.render(results)
+        elif args.html:
+            try:
+                r = floss.render.html.render(results)
+            except floss.render.html.HtmlTemplateError as e:
```
mr-tz: should the iniital check test all requirements so we don't need an extra try/except here?

floss/main.py
```diff
@@ -300,6 +316,13 @@ def main(argv=None) -> int:
 
     if args.json:
         r = floss.render.json.render(analysis_results)
+    elif args.html:
+        logger.info("rendering results")
+        try:
+            r = floss.render.html.render(analysis_results)
+        except floss.render.html.HtmlTemplateError as e:
```
mr-tz: should the iniital check test all requirements so we don't need an extra try/except here?

doc/usage.md
vee1e: fixed in 055ccd8

doc/usage.md
vee1e: fixed in f44251d

floss/render/templates/index.html
vee1e: removed in 33d35d4

floss/main.py
vee1e: removed in 079bc64

floss/main.py
vee1e: yeah, removed the catch in 079bc64 now its all validated

doc/usage.md
mr-tz: I meant the opposite and remove this here :)

floss/render/templates/index.html
mr-tz: Do we need the different paths for exe vs pip?

.github/pyinstaller/floss.spec
```diff
@@ -55,6 +55,8 @@ datas = [
     # this gets invoked from the directory of the spec file,
     # i.e. ./.github/pyinstaller
     ('../../floss/sigs', 'sigs'),
+    # pre-built HTML viewer used by ``floss --html`` (from ``npm run build``)
+    ('../../viewer/dist/index.html', 'viewer/dist'),
```
mr-tz: Does build fail if not existent?

floss/main.py
```diff
@@ -300,6 +312,9 @@ def main(argv=None) -> int:
 
     if args.json:
         r = floss.render.json.render(analysis_results)
+    elif args.html:
+        logger.info("rendering results")
```
mr-tz: Remove the log and below and elsewhere?

doc/usage.md
vee1e: Removed the incompatibility sentence from usage.md in e96843a.

floss/render/templates/index.html
vee1e: Pip and the exe now both read floss/render/templates/index.html in 4ca788d.

.github/pyinstaller/floss.spec
vee1e: PyInstaller now exits if viewer/dist/index.html is missing in 7f59aef.

floss/main.py
vee1e: Removed the rendering results logs in 97bda1a.

.github/pyinstaller/floss.spec
```diff
@@ -50,11 +51,16 @@ with open("./floss/version.py", "wb") as f:
     )
     f.write(("__version__ = '%s'" % version).encode("utf-8"))
 
+if not Path("viewer/dist/index.html").is_file():
+    raise SystemExit("missing viewer/dist/index.html; run: cd viewer && npm install && npm run build")
+
```
mr-tz: I was just asking :)
Do you think that's a hard requirement? Probably few people build binaries on their own and if so can figure out how to disable this requirement if they dont want it.

floss/cli.py
```diff
@@ -348,6 +351,11 @@ def make_parser():
 
     output_group = parser.add_argument_group("rendering arguments")
     output_group.add_argument("-j", "--json", action="store_true", help="emit JSON instead of text")
+    output_group.add_argument(
+        "--html",
+        action="store_true",
+        help="emit a standalone HTML report instead of text; cannot be combined with -j/--json",
```
```suggestion
        help="emit a standalone HTML report instead of text",
```

viewer/package.json
```diff
@@ -7,7 +7,7 @@
   },
   "scripts": {
     "dev": "node scripts/generate-version.mjs && vite",
-    "build": "node scripts/generate-version.mjs && tsc -b && vite build",
+    "build": "node scripts/generate-version.mjs && tsc -b && vite build && node -e \"const fs=require('fs');fs.mkdirSync('../floss/render/templates',{recursive:true});fs.copyFileSync('dist/index.html','../floss/render/templates/index.html')\"",
```
mr-tz, line 10: can you explain more on the reasoning for that target path vs. staying in viewer?

mr-tz (commented):
few remaining questions, almost ready IMO

.github/pyinstaller/floss.spec
vee1e: not a hard requirement i feel. we can drop the SystemExit so a local PyInstaller build still works without npm, CI already builds the viewer first. wdyt?

mr-tz: yeah maybe just a warning then?

viewer/package.json
vee1e: its copied into templates/ so pip, source, and the exe can all use one path next to the renderer. viewer/ is not part of the Python package, so a wheel would not include viewer/dist otherwise.

floss/main.py
mr-tz: ah there was a comment:         # this may be slow when there's many strings, so informing users what's happening

maybe worth keeping until we figure out performance, WDYT?

.github/pyinstaller/floss.spec
vee1e: switched it to a warning in b388c11. If the file is missing the exe still builds, just without `--html`.

floss/main.py
vee1e: makes sense for text. i'll put the log back on the text renderer only, JSON and `--html` can stay quiet.

vee1e: done in 8b1727b

mr-tz (approved):
thanks!

PR #1439, opened 2026-09-15
fix(oss-db): stop touching databases that did not change

vee1e:
Rebuilds rewrote every database file even when its strings were identical. That is why PR #1438 shows a new hash for every `.jsonl.gz` while the listed diff covers only some libraries. Unchanged databases are now left alone.

Three causes, all in `scripts/tags/build_oss_db.py`:

1. `gzip.open()` writes the current time and the output filename into the gzip header, so identical content packs to different bytes on each run. `Converter.write` now packs canonical JSONL with `gzip.compress(..., mtime=0)`.
2. Function-name entries were appended while looping over a Python set, so their order depended on the interpreter's string hash seed. `serialize_entries` sorts the JSONL lines before writing.
3. `run_build` wrote every rebuilt library even when nothing changed. `write_library_database` compares the new entries against the existing file and skips the write when they match.

The 60,000-character PR body cap (`DIFF_PR_MAX_CHARS`) was not involved. It only truncates the displayed diff.

Tests: 3 in `tests/test_build_oss_db.py` cover byte reproducibility, the skip path, and that `run_build` forwards `existing_entries`. All 3 fail on the previous code.

Branch: +108 / -34 across 2 files.

Verification:
- `pre-commit run --all-files` (isort, black, mypy, pytest): passed
- `pytest tests/`: 442 passed, 2 skipped, 6 xfailed


mr-tz (approved):

mr-tz:
Thanks @vee1e! Let's rebuild dbs based on this now.

